PR #674, opened 2022-06-23
654 pass reactive data@main

### LINKED ISSUE #654: Pass the list of reactive data to the module.
- Blocked by https://github.com/insightsengineering/teal.slice/issues/33

Our objective is teal_module(s) to receive the data as a `list(adsl = reactive(<data.frame>), x = reactive(<data.frame>))` instead of taking data from the `FilteredData` object. In the flow-chart below what we need to do in this issue is distinguished by "bolded" frames. This means that if `teal_module` has a data argument, we need to pass a list which has `expr`, `join_keys` and `metadata` as attributes.

quosure-teal-simplify-api (2)

```r
tm_module(
  a b c <instructions for choices> # currently data_extract
  data <list of reactives, optionally with expr, metadata and join_keys attr>, 
  # datasets - FilterPanel-api if the communication with filteredData needed
  # reporter - if reporter-api needed 
)
```

Practically my first guess is that following lines might be enough before calling the module - but it depends how far refactor of the FilteredData will go (https://github.com/insightsengineering/teal.slice/issues/33)
```r
data <- lapply(
  <module datanames>, 
  function(x) {
    dataset <- datasets$get_data(dataname = x, filtered = TRUE))
    attr(dataset, "metadata") <- <metadata>
    dataset
  }
)
attr(data, "join_keys") <- <join keys>
attr(data, "expr") <- c(datasets$get_code(<module datanames>, datasets$get_filter_expr(<datanames>))
```

-----
Currently following methods from FilteredData are used in the modules, they should be passed somehow in the data attributes:
- $get_data(filtered = TRUE)
- $get_data(filtered = FALSE) ❌  [see occurrences](https://github.com/search?q=org%3Ainsightsengineering+filtered+%3D+FALSE&type=code) [and here](github.com/search?q=datasets%24get_data&type=Code). This can be fixed by including unfiltered data to the module also.
- $datanames() - this will be the names of the list
- $get_metadata(), $get_keys() - should we pass using ...? Please find the optimal solution.
- $get_varlabels - this can be extracted from the raw data.frame object

gogonzo:
closes #654 

1. Passing list of reactive data to the teal_module if has `data` argument. I'm passing the list with attributes as I think it's easier than `list(data = list(ADSL = reactive(), ...), attributes = list(expr = reactive(), ...))`. 
2. changes the validation of the teal_module which was quite strict by requiring ordered arguments, always need datasets or `...`. We can ease this requirement because some modules might not need datasets at all. in `ui_nested_tabs` and `srv_nested_tabs` we use `do.call` which pass `args` by their name (doesn't care about order)
3. Extracted some code from get_rcode which combines DDL with filter expr

-------
I found some inconsistency. For example:
- In the filter-panel we turn ADSL -> ADSL_FILTERED but everywhere we still use unchanged datanames. In data_merge [we use hardcoded _FILTERED](https://github.com/insightsengineering/teal.transform/search?q=_FILTERED&type=code) which means that "independent" data_merge still depends on the environment where ADSL_FILTERED exist (but we use dataname = "ADSL")

-----
### Sample app
```r
mod_data <- teal::module(
  ui = function(id) {
    ns <- NS(id)
    verbatimTextOutput(ns("test"))
  },
  server = function(id, data) {
      output$test <- renderText({
        paste(attr(data, "code")(), collapse = "\n")
      })
    })
  },
  label = "mod_data",
  filters = "all",
  server_args = NULL,
  ui_args = NULL
)


app <- init(
  data = teal_data(
    dataset("IRIS", iris, code = "IRIS <- iris"),
    dataset("MTCARS", mtcars, code = "MTCARS <- mtcars"),
    check = TRUE
  ),
  modules = modules(example_module(), mod_data),
  header = "My first teal application"
)

runApp(app)
```

### Sample app with lightweight reproducible object

```r
devtools::load_all("teal")

eval_code <- function(data, code) {
  # some assertions for data and code
  env <- list2env(
    sapply(data, simplify = TRUE, function(x) if (is.reactive(x)) x() else x)
  )
  eval(expr = parse(text = code),  envir = env)
  for (i in ls(env)) {
    data[[i]] <- env[[i]]
  }
  attr(data, "code") <- c(attr(data, "code"), code)
  data
}

mod_data <- teal::module(
  ui = function(id) {
    ns <- NS(id)
    div(
      verbatimTextOutput(ns("test")),
      tableOutput(ns("filtered"))
    )
  },
  server = function(id, data) {
      react1 <- reactive(eval_code(data, "new_data <- IRIS"))
      react2 <- reactive(eval_code(react1(), "new_data <- head(new_data)"))
      output$test <- renderText(attr(react2(), "code") |> paste(collapse = "\n"))
      output$filtered <- renderTable(react2()[["new_data"]])
    })
  },
  label = "mod_data",
  filters = "all",
  server_args = NULL,
  ui_args = NULL
)


app <- init(
  data = teal_data(
    dataset("IRIS", iris, code = "IRIS <- iris"),
    dataset("MTCARS", mtcars, code = "MTCARS <- mtcars"),
    check = TRUE
  ),
  modules = modules(example_module(), mod_data),
  header = "My first teal application"
)

runApp(app)

```


R/module_nested_tabs.R
```diff
@@ -188,12 +188,40 @@ srv_nested_tabs.teal_module <- function(id, datasets, modules, reporter) {
...
+      )
+      #names(data) <- paste0(names(data), "_FILTERED") # these datasets are filtered
+
+      # code from previous stages
+      attr(data, "code") <- reactive(get_datasets_code(datanames, datasets))
+
+      # join_keys
+      attr(data, "join_keys") <- reactive(datasets$get_join_keys()[datanames])
+
+      args <- c(args, data = list(data))
```
nikolas-burkoff, line 222: Something doesn't look quite right here - is `data` a list from the `sapply` which you are then adding the attributes onto but then `args` is setting `data = list(data)`

nikolas-burkoff (commented):
A couple of comments from me

R/module_nested_tabs.R
gogonzo: This is a consequence of how `c` and `append` works with lists. `c` rather appends elements of the list instead of adding a whole list. If I didn't do it, `args` would contain `adsl = <reactive>`, `adtte = <reactive>` etc. instead of `data`. 

image

R/module_nested_tabs.R
```diff
@@ -188,12 +188,40 @@ srv_nested_tabs.teal_module <- function(id, datasets, modules, reporter) {
...
     if (is_reporter_used(modules)) {
       args <- c(args, list(reporter = reporter))
     }
 
+    is_datasets_used <- isTRUE("datasets" %in% names(formals(modules$server)))
+    if (is_datasets_used) {
+      args <- c(args, datasets = datasets)
+    }
+
+    is_data_used <- isTRUE("data" %in% names(formals(modules$server)))
```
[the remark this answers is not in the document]
gogonzo: Good point

R/module_nested_tabs.R
nikolas-burkoff: fair enough - good old R...

R/module_nested_tabs.R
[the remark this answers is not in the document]
Polkas: we will need is_module_ui_arg_used and is_module_srv_arg_used to cover all scenarios. Or add an argument `is_module_arg_used(..., type = "srv" or "ui")`.

gogonzo:
@nikolas-burkoff @Polkas @mhallal1 @kpagacz 
I just want to highlight one thing. There were multiple scenarios how should we call modules to maximize simplicity.

```r
# 1. use expr argument
srv_custom <- function(data = list(...), expr = <expr>, join_keys = <join_keys>, ....)

# 2. use attributes
srv_custom <- function(data = structure(list(...), expr = <expr>, join_keys = <join_keys>, ...)

# 3. use nested list 
srv_custom <- function(data = list(data = list(....), expr = <expr>, join_keys = <join keys>, ...)
```

Together with @pawelru we think that nested list (3) is to "complicated" to put reactive data into `data` element of the list. We debated about (1) and (2) and we concluded that having `expr` and `join keys` as attribute is optimal solution. here is why:
- `expr` is a code to reproduce data and it's stricltly connected to data.
- `join_keys` is entity of the data which helps to define relationships between datasets. Also including `dm` would make this argument irrelevant as `dm` already have a join_keys
- In the future our "chunks" will be a simple object which will be a `env/list` + `expr`  which should be accessed by `[[`, `$`, `[`, `getElement` and evaluated by simply `eval_code(<chunks>, <expr>)`. Having `list` + `attr` everywhere will be easier for users to understand how to construct NEST objects.
- having `expr` and `join_keys` as arguments of the module will require us to conditionally pass them in (same as reporter, data etc.).

Please let me know what is your opinion about this.

R/module_nested_tabs.R
```diff
@@ -183,11 +188,40 @@ srv_nested_tabs.teal_module <- function(id, datasets, modules, reporter) {
...
+    # list of reactive filtered data
+    data <- sapply(
+      datanames,
+      simplify = FALSE,
+      function(x) {
+        #todo: need to include metadata, keys, datalabel
+        reactive(datasets$get_data(x, filtered = TRUE))
+      }
+    )
+    #names(data) <- paste0(names(data), "_FILTERED") # these datasets are filtered
```
gogonzo, line 214: I'd like also to highlight this. `data` contains "filtered data" which is supported by `expr` where `ADSL_FILTERED <- filter(ADSL, ...)`. So, we have little inconsistency here as `data` is named by "generic" dataname (ADSL, ADTTE, ...).

I was thinking that teal.slice should execute expr in such way that `ADSL <- filter(ADSL, <expr>)` to have only one variant.

R/get_rcode.R
```diff
@@ -59,106 +56,40 @@
...
   }
   checkmate::assert_string(title, null.ok = TRUE)
   checkmate::assert_string(description, null.ok = TRUE)
-
   rlang::push_options(width = 120)
 
+  if (!is.null(session)) {
+    lifecycle::deprecate_warn("0.11.2", "get_rcode(session)")
+  }
+
```
nikolas-burkoff, line 75: So there's a specific lifecycle argument for deprecating just an argument

image

R/module_nested_tabs.R
```diff
@@ -179,16 +208,45 @@ srv_nested_tabs.teal_module <- function(id, datasets, modules, reporter) {
...
+  args <- c(list(id = id), modules$server_args)
+  if (is_args_used(modules$server, "reporter")) {
     args <- c(args, list(reporter = reporter))
   }
 
-  if (is_module_server) {
+  is_datasets_used <- is_args_used(modules$server, "datasets")
+  if (is_datasets_used) {
+    args <- c(args, datasets = datasets)
+  }
```
nikolas-burkoff, line 222: You can probably just do `if(is_args_used(modules$server, "datasets"))` and in other places

R/module_nested_tabs.R
```diff
@@ -88,9 +88,38 @@ ui_nested_tabs.teal_modules <- function(id, modules, datasets, depth = 0L) {
...
+        # todo: need to include metadata, keys, datalabel
+        reactive(datasets$get_data(x, filtered = TRUE))
+      }
+    )
+
+    # code from previous stages
+    attr(data, "code") <- reactive(get_datasets_code(datanames, datasets))
+
+    # join_keys
+    attr(data, "join_keys") <- reactive(datasets$get_join_keys()[datanames])
```
nikolas-burkoff, line 118: should they be reactive here?

R/modules.R
```diff
@@ -108,34 +108,42 @@ append_module <- function(modules, module) {
...
 }
 
-#' @rdname is_reporter_used
+#' @rdname is_args_used
 #' @export
 #' @keywords internal
-is_reporter_used.teal_modules <- function(modules) {
-  any(unlist(lapply(modules$children, function(x) is_reporter_used(x))))
+is_args_used.teal_modules <- function(modules, args) {
+  any(unlist(lapply(modules$children, function(x) is_args_used(x, args))))
```
nikolas-burkoff, line 131: ?
```suggestion
  any(unlist(lapply(modules$children, is_args_used, args))))
```

nikolas-burkoff (commented):
A few comments from me

NEWS.md
```diff
@@ -2,8 +2,15 @@
 
 ### Enhancements
 
+* `teal_module` having `data` argument in it's arguments will receive list of reactive filter data with `"code"` and `"join_keys"` attributes.
```
mhallal1, line 5:
```suggestion
* `teal_module` having `data` argument in its arguments will receive a list of reactive filter data with `"code"` and `"join_keys"` attributes.
```

R/modules.R
```diff
@@ -108,34 +108,42 @@ append_module <- function(modules, module) {
...
-  stop("is_reporter_used function not implemented for this object")
+#' @export
+is_args_used.default <- function(modules, args) {
+  stop("is_args_used function not implemented for this object")
 }
 
-#' @rdname is_reporter_used
+#' @rdname is_args_used
 #' @export
 #' @keywords internal
```
mhallal1, line 129: do we need to export this?

gogonzo: It's not exported formally. @export tag over S3 method registers this method in the namespace (needed to dispatch by the type). If we don't put @export here the method `is_args_used.teal_module` will be treated as is instead of being `is_args_used` dispatched by type.

```
S3method(is_args_used,teal_module)
```

NEWS.md
```diff
@@ -2,8 +2,15 @@
...
 * Updated the internals of `module_teal` to reflect changes in `teal.slice`.
 
+### Breaking changes
+
+* `...` argument in the `teal_module` no longer receives `datasets` object. In order to use `datasets` in the `teal_module` please specify `datasets` explicitly.
+* Deprecated `merge_expression` argument in `get_rcode_srv` and `get_rcode` has been removed.
```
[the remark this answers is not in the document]
gogonzo: Unfortunately no, get_rcode is still here. I just removed stuff deprecated for more than year.

NEWS.md
```diff
@@ -2,8 +2,15 @@
...
 ### Enhancements
 
+* `teal_module` having `data` argument in its arguments will receive list of reactive filter data with `"code"` and `"join_keys"` attributes.
 * Updated the internals of `module_teal` to reflect changes in `teal.slice`.
 
+### Breaking changes
+
+* `...` argument in the `teal_module` no longer receives `datasets` object. In order to use `datasets` in the `teal_module` please specify `datasets` explicitly.
+* Deprecated `merge_expression` argument in `get_rcode_srv` and `get_rcode` has been removed.
+* Deprecate `session` argument in `get_rcode` function.
```
mhallal1, line 12:
```suggestion
* Deprecated `session` argument in `get_rcode` function.
```

tests/testthat/test-modules.R
```diff
@@ -24,225 +24,103 @@ testthat::test_that("Calling module() does not throw", {
...
+    server = function(id, any_argument) NULL,
   ), NA)
 
   testthat::expect_error(module(
-    label = "label",
-    server = function(id, datasets) NULL,
-    ui = function(id, datasets, ...) NULL,
-    filters = ""
+    server = function(id, any_argument) NULL,
   ), NA)
```
mhallal1, line 47: test repeated twice (see the test before).

mhallal1 (commented):
passing `data` and `datasets` to the same module seems to be possible also, is this by design?

R/module_nested_tabs.R
```diff
@@ -88,9 +88,35 @@ ui_nested_tabs.teal_modules <- function(id, modules, datasets, depth = 0L) {
 #' @export
 #' @keywords internal
 ui_nested_tabs.teal_module <- function(id, modules, datasets, depth = 0L) {
-  stopifnot(is(datasets, "FilteredData"))
+  checkmate::check_class(datasets, "FilteredData")
```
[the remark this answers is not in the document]
gogonzo: `ui_nested_tabs` needs `FilteredData` and `FilteredData` will still be delivered here for a long time in the same shape. nested_tabs is the place where convertion from FIlteredData -> data happens.   

Also, nested_tabs is not exported also which means that we don't have to simplify api for this one.

R/modules.R
```diff
@@ -277,25 +291,50 @@ module <- function(label = "module",
   checkmate::assert_function(server)
   checkmate::assert_function(ui)
   checkmate::assert_character(filters, min.len = 1, null.ok = TRUE, any.missing = FALSE)
-  checkmate::assert_list(server_args, null.ok = TRUE)
-  checkmate::assert_list(ui_args, null.ok = TRUE)
+  checkmate::assert_list(server_args, null.ok = TRUE, names = "named")
+  checkmate::assert_list(ui_args, null.ok = TRUE, names = "named")
+
+  server_formals <- names(formals(server))
+  if (!any(c("id", "input", "output", "session") %in% server_formals)) {
```
nikolas-burkoff, line 298: So something like `function(input, x)` would be allowed - I guess that's fine as if you are silly enough to do that then shiny will shout loudly at you.

R/modules.R
```diff
@@ -277,25 +291,50 @@ module <- function(label = "module",
...
-    !identical(names(formals(ui))[[2]], "datasets") && !identical(names(formals(ui))[[2]], "...")
-  ) {
+  ui_extra_args <- setdiff(names(ui_args), ui_formals)
+  if (length(ui_extra_args) > 0 && !"..." %in% ui_formals) {
     stop(
-      "module() ui argument requires a function with two ordered arguments:",
-      "\n- 'id'\n- 'datasets' or '...'"
+      "\nFollowing `ui_args` elements have no equivalent in the formals of `ui`:\n",
+      paste(paste(" -", ui_extra_args), collapse = "\n"),
+      "\n\nUopdate the `ui` arguments by including above or add `...`"
```
nikolas-burkoff, line 337:
```suggestion
      "\n\nUpdate the `ui` arguments by including above or add `...`"
```

tests/testthat/test-modules.R
```diff
@@ -541,43 +415,43 @@ get_srv_and_ui <- function() {
...
 
-testthat::test_that("is_reporter_used returns true if teal_module has reporter in server function args", {
-  testthat::expect_true(is_reporter_used(module(server = function(id, datasets, reporter) NULL)))
+testthat::test_that("is_args_used returns true if teal_module has reporter in server function args", {
+  testthat::expect_true(is_args_used(module(server = function(id, datasets, reporter) NULL), "reporter"))
 })
 
-testthat::test_that("is_reporter_used returns false if teal_module does not have reporter in server function args", {
-  testthat::expect_false(is_reporter_used(module()))
+testthat::test_that("is_args_used returns false if teal_module does not have reporter in server function args", {
```
nikolas-burkoff, line 427: These test names are a bit odd now probably need `is_args_used with args = reporter` (and I guess we can use arbitrary args not just reporter now).

We should probably also test is_args_used.function (does not need to be a ui and server function of course)

R/get_rcode.R
```diff
@@ -189,32 +118,83 @@ get_rcode <- function(datasets = NULL,
...
+      ),
+      collapse = "\n"
+    ),
+    "\n\n"
+  )
 
-  if (exists("progress")) {
-    progress$set(message = "Getting R Code", value = 1)
-    progress$close()
+  str_filter <- teal.slice::get_filter_expr(datasets, datanames)
```
nikolas-burkoff, line 184: @gogonzo @mhallal1 - this will be right, but I'm super confused about how this interacts with the merge expression which also has filtering in? No need to explain I guess as this all is being changed

nikolas-burkoff (commented):
This looks good to me, a few minor comments. I'll let @mhallal1 approve as he's been much closer to this than me

R/modules.R
gogonzo: Fine, I'm changing to
```r
!(
    "id" %in% server_formals ||
    all(c("input", "output", "session") %in% server_formals)
  )
```

I will include test also

tests/testthat/test-modules.R
gogonzo: I changed `is_args_used` (plural) to `is_arg_used` (single) because `arg` should probably be a single character instead of multiple.  
With multiple there is dilemma:
- should all together be present in any of the function or any.

Because we don't need multiple version now I have simplified a little.

mhallal1 (approved):
LGTM

PR #692, opened 2022-07-12
filter_panel_api

mhallal1:
related to https://github.com/insightsengineering/teal.slice/issues/64
Append `filter_panel_api` argument to the list of arguments in `srv_nested_tabs.teal_module`.


R/modules.R
```diff
@@ -280,7 +281,7 @@ root_modules <- function(...) {
 #' runApp(app)
 #' }
 module <- function(label = "module",
-                   server = function(id, data) {
+                   server = function(id, data, filter_panel_api) {
```
nikolas-burkoff, line 284: In the stop starting on line 305 you should also refer to filter_panel_api ?

mhallal1: Screenshot from 2022-07-12 11-17-02

PR #694, opened 2022-07-12
Fix when `module(..., filters  = NULL)`

### LINKED ISSUE #693: [Bug]: App fails when module has a data argument and `filter = NULL`
### What happened?

Following app fails as it fails on making list of data on datasets (`filter = NULL`)

```r
library(teal)
library(shiny)


app <- init(
  data = teal_data(dataset("iris", iris)),
  modules = list(
    module(
      label = "Module",
      server = function(id, data) {
        moduleServer(
          id,
          module = function(input, output, session) {
            output$data <- renderDataTable(data[["iris"]]())
          }
        )
      },
      ui = function(id) {
        ns <- NS(id)
        tagList(dataTableOutput(ns("data")))
      },
      filters = NULL
    )
  )
)
runApp(app)
```

### sessionInfo()

_No response_

### Relevant log output

_No response_

### Code of Conduct

### Contribution Guidelines

### Security Policy

gogonzo:
closes #693 

App fails after initialization but it should fail before. `filters` in fact denote active datanames and it determines which datasets land in the module in `data` list. This means that if `filters = NULL` and `data` is the argument of the module we have a problem of empty list. datasets used inside of the module shouldn't be different that `filters`.

Note: I didn't make this stop if `datasets` is in formals, because:
1. It has a potential to be a breaking change
2. `filters` together with `datasets` disable only the visibility of the filter-panel but don't impact availability of the data inside of the teal_module's server.

```r
app <- init(
  data = teal_data(dataset("iris", iris)),
  modules = list(
    module(
      label = "Module",
      server = function(id, data) {
        moduleServer(
          id,
          module = function(input, output, session) {
            output$data <- renderDataTable(data[["iris"]]())
          }
        )
      },
      ui = function(id) {
        ns <- NS(id)
        tagList(dataTableOutput(ns("data")))
      },
      filters = NULL
    )
  )
)
runApp(app)

```

### DIFF
--- a/R/modules.R
+++ b/R/modules.R
@@ -280,10 +280,10 @@ root_modules <- function(...) {
 #' runApp(app)
 #' }
 module <- function(label = "module",
-                   server = function(id, data, filter_panel_api) {
+                   server = function(id, ...) {
                      moduleServer(id, function(input, output, session) {})
                    },
-                   ui = function(id, data) {
+                   ui = function(id, ...) {
                      tags$p(paste0("This module has no UI (id: ", id, " )"))
                    },
                    filters = "all",
@@ -323,10 +323,10 @@ module <- function(label = "module",
     )
   }
 
-  if (c("data", "datasets") %in% server_args && is.null(filters)) {
+  if ("data" %in% server_formals && is.null(filters)) {
     stop(
-      "\n`filters = NULL` indicates that module doesn't need any data while it has the `data` in formals."
-      "\nPlease specify `filters` with the names of needed datasets or exclude `data` from the arguments.
+      "\n`filters = NULL` indicates that the module doesn't need any data while it has the `data` in formals.",
+      "\nPlease specify `filters` with the names of needed datasets or exclude `data` from the arguments."
     )
   }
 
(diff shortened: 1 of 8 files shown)


R/modules.R
```diff
@@ -243,9 +244,8 @@ root_modules <- function(...) {
 #' @param filters (`character`) A vector with datanames that are relevant for the item. The
 #'   filter panel will automatically update the shown filters to include only
 #'   filters in the listed datasets. `NULL` will hide the filter panel,
-#'   and the keyword `'all'` will show the filters of all datasets. The
-#'   argument can be thought of as `'active_datanames'` and may be renamed
-#'   in future versions of teal.
+#'   and the keyword `'all'` will show the filters of all datasets. `filters` can't be `NULL` when
+#'   `data` is
```
nikolas-burkoff, line 248: when data is ...

tests/testthat/test-modules.R
```diff
@@ -125,6 +125,13 @@ testthat::test_that("module requires filters argument to be a character", {
   testthat::expect_error(module(filters = NA_character_), "Contains missing values")
 })
 
+testthat::test_that("module shouldn't have filter = NULL and data or datasets in the formals", {
```
nikolas-burkoff, line 128:
```suggestion
testthat::test_that("module shouldn't have filter = NULL and data in the formals", {
```

nikolas-burkoff (approved):
looks good, one minor change to test name if you want

tests/testthat/test-modules.R
gogonzo: Sorry, I realized to late about this suggestion

PR #701, opened 2022-07-21
Example skeleton project 

nikolas-burkoff:
DRAFT PR not to be merged FyI @gogonzo @pawelru @kumamiao 

To use you may need to install teal a few times for RStudio to get it working.

This PR (unmerged) closes https://github.com/insightsengineering/NEST-roadmap/issues/59

We now have:

image

image

image

Or if you chose not to have cdisc:
image

Clearly if we are doing this for real then
1) I still suggest it goes in teal rather than having to install a new package
2) The `package_skeleton` function can be called from other IDEs
3) We'd want to structure the `package_skeleton` function in a nicer way and output a proper structure not just a single app.R file


dinakar29:
This looks great, @nikolas-burkoff!

> You could even have options to create specific apps 

Skeleton apps derived from sample apps on teal.gallery sounds like a great idea. We should consider that in the near future. Let me add that as an IDR task, as it seems pretty straightforward.

inst/rstudio/templates/project/teal_application.dcf
```diff
@@ -0,0 +1,9 @@
+Binding: project_skeleton
+Title: Teal Application
+OpenFiles: app.R
+Icon: teal.png
+
+Parameter: use_cdisc
+Widget: CheckboxInput
```
gogonzo, line 7: (in the future) Can the Widget be for example?
```
Widget: availableModulesInput
```

nikolas-burkoff: you can only have the following:

image

We may want an addin to do the add modules part rather than at this project creation time. And for creating your own custom modules you could maybe? use [snippets](https://rstudio.github.io/rstudio-extensions/rstudio_snippets.html)

pawelru: Is snippet a Rstudio-exclusive feature?

pawelru:
The addon looks good but the content is not there yet (however it's a draft so not sure if I can focus on it at all). For sure I would go away from single-file apps in favour of more sophisticated multiple files. The content is to-be-decided (researched). Feel free to to have a look what's in already archived `tealApp` project/repo somewhere. 
I also think that it would be better/easier to have scripts (in `inst/`) as files than scripts as character string.

I like addon because it promotes good practice of using "New Project" (at least from RStudio).
I dislike addon as the configuration possibilities is very limited. I can imagine that the number of config controls would grow (e.g. use testing, use logging and if yes - what log type or even custom module selector as pointed by Dawid) and we might want to group them, make it optional / conditional, etc. This is not possible. I am not yet sure if that's bad - maybe it's good to have it super simple - but I see it as a limitation for the future.

gogonzo:
@kumamiao @donyunardi we have not decided exactly what is needed here. If addin should be a single file or multiple files.
 There was proposition of having `data.R`, `modules.R` sourced from the `app.R` but wasn't formally confirmed. We suppose to promote good practices but I don't know if these are good practices. Also, we probably should initialize with empty `www/css`, `www/js`. Am I missing something?
 
Additionally, I think that it shouldn't be a separate package because users shouldn't be bothered to install a new package from github. I think we should rather keep it in teal. Alternatives are not good enough, because if we make a new independent package hosted on github then users will be requested to install this. Package will not be able to be installed on CRAN because it's a utility package, which means it can't be in teal/Suggests.

donyunardi:
@nikolas-burkoff mentioned this earlier on creating a custom_modules folder, which I thought was a good idea. Another one that comes to mind is to create a folder to store custom functions.

I think separating sourcing data.R and modules.R from app.R is a good idea. This will make the code more organized and easier to debug. And, while we're on this topic, when I looked around in Roche GitHub, I noticed people's `modules=` code was long due to the number of parameters for each module and this makes it hard to see the organization of the modules. 

What do you think if we separate the module call into its own file under a separate folder? 
Here's to illustrate what I meant:

_app.R_
```
library(teal)

source("data.R")
source("modules.R")
lapply(list.files("module", full.names = TRUE), source)

app <- init(
  data = data,
  modules = modules,
  header = "My first teal application"
)

if (interactive()) {
  runApp(app)
}
```

_data.R_
```
data <- teal_data(
  dataset("IRIS", iris),
  dataset("MTCARS", mtcars)
)
```

_modules.R_
```
modules <- modules(
  x1,
  x2,
  modules(
    label = "Nested",
    x1,
    x2
  )
)
```

_module/module1.R_
```
x1 <- example_module()
```

_module/module_2.R_
```
x2 <- example_module()
```
The structure would look like this:
```
/
-- app.R
-- data.R (TBD)
-- modules.R (TBD)
-- module/
    -- module1.R
    -- module2.R
```

nikolas-burkoff:
> Additionally, I think that it shouldn't be a separate package because users shouldn't be bothered to install a new package from github. I think we should rather keep it in teal.

Yes I agree

> on creating a custom_modules folder, which I thought was a good idea

So my hope was that existing modules would soon be simplified enough to not need separate files for all of them (i.e. `definitions.R` -> which stores all the data extract specs etc. and then one file for the app). The `custom_modules` folder was originally meant to be used if your app used user-defined modules (i.e. a module created yourself not one from tmg, tmc etc.) but I'm open to whatever structure we think makes sense.

inst/rstudio/templates/project/teal_application.dcf
donyunardi: I don't think snippet is RStudio exclusive feature (i.e. you can add code snippet on vscode too). However, the syntax will different.

donyunardi:
> So my hope was that existing modules would soon be simplified enough to not need separate files for all of them (i.e. definitions.R)

Sure, that sounds good too.

This skeleton template does seem like a possible solution for the marketplace. However, the template won't be able to show module visualization for the user to see before making a selection. The initial proposal was for the user to be able to see and select the module(s) from the marketplace. I am not sure if the requirement can be fulfilled using the skeleton template. 

It feels like we're still undecided between the marketplace will produce one file vs multiple files. I think multiple files will improve code readability and maintenance, but this could lead to major changes to our vignettes. Would this be a good topic for refinement?

kpagacz:
Oh god, it's happening

asbates:
Came across this so thought I'd share my 2 cents. It may have already been done but I'd highly suggest looking at [golem](https://github.com/ThinkR-open/golem) and [rhino](https://github.com/Appsilon/rhino/) because app templates are a big part of both. They also take very different approaches so it could be useful to compare and contrast.

pawelru:
@asbates please see linked tasks to https://github.com/insightsengineering/NEST-roadmap/issues/9

Issue #752, opened 2022-10-07
SRC showing sessionInfo() and other header information


donyunardi:
from @nikolas-burkoff 
- [x] SRC currently shows sessionInfo (or at least versions of NEST packages installed) - we are losing this. Should we be?
Edit: this might be fixed by https://github.com/insightsengineering/teal/pull/767

- [x] Also we are losing [this](https://github.com/insightsengineering/teal/blob/0f8f1e89fe49a6e33837437516ee2d53c89570c4/vignettes/teal-options.Rmd#L67) option  fixed by https://github.com/insightsengineering/teal/pull/774

- [x] We are also losing which library() calls we're using. Should we be? fixed by https://github.com/insightsengineering/teal/pull/774

donyunardi:
It doesn't have to in show R code but should be somewhere.

nikolas-burkoff:
Maybe look at putting it into teal rather than the modules would simplify things

nikolas-burkoff:
I think for the sessionInfo we could have a hyperlink in the footer which if you click on opens a pop up with the session info

Polkas:
We have three separate problems here and they are very important for the reproducibility objective (the core of the NEST).
I feel we should invest here some time and make sure we are happy with solutions.

- In my opinion already done part about session info could be improved with `sessioninfo::session_info()` or the renv lock file. Mostly nobody want to see the sessionInfo but to quickly reproduce the env so renv lock file could be the right direction.
@nikolas-burkoff what you think?
- So the "Show R Code" is not reproducible as no library calls are added by us now. This looks to be a part of teal.data and a feature of qenv.
- The teal code option is a part of qenv in teal.code for me. Not obvious have to achieve it.

Is it clear I have to go to sessionInfo to reproduce the env and then run the "Show R Code"? We should think about some Message in the "Show R Code" modal that 'please make sure the environment is properly reproduce with XYZ'.

Polkas:
As we are not adding libraries on the level of qenv we do not know what packages are really needed.
As a result we persist the old behavior of adding all packages loaded in the R session (`sessionInfo`) https://github.com/insightsengineering/teal/blob/ffb0421e1ae7d5251e13dfb1bd2dd9dadd587924/R/get_rcode_utils.R#L41-L52 I think we should create a separate issue for investigating it as it was not implemented or discussed here

Another thing is that I proposed improvements to https://github.com/insightsengineering/teal/pull/767 which was not commented. @nikolas-burkoff 

[...]

PR #774, opened 2022-11-09
add hashing

### LINKED ISSUE #751: qenv - Show R Code check = FALSE notification; moving hash out of teal.slice and teal.data into teal
From @nikolas-burkoff 
Do we care about when check = FALSE having some notification in the SRC like we currently do?

Also we are losing the check on the hash of the datasets

mhallal1:
closes #751 

The dataset are being hashed using `rlang::hash` in:
1) `data_to_datasets` when using `qenv`
2) `get_rcode` when using chunks `chunks`

-`calculate_hashes` function has been created to be used in both places.
-the filtering is not returned in `get_datasets_code` now but rather in `get_rcode` and `datasets_to_data`.
-messages and warnings calls are now returned in `get_datasets_code` instead of character strings.

Please test with `chunks` and `qenv` examples from tmg, tmc ...

Also closes https://github.com/insightsengineering/teal/issues/752


nikolas-burkoff (commented):
In the NEWS we should mention not only have the hashes been moved here but that we now use rlang instead of digest (this is important if the hash algorithms produce different answers - I'm not sure they are)

NEWS.md
```diff
@@ -13,6 +13,7 @@
 
 * Updated examples to use `scda.2022`.
 * Added R session information into a link in the footer of `teal` applications.
+* Added data hashing step using `rlang` instead of `digest` package and hashing check in the reproducible code.
```
nikolas-burkoff, line 16:
```suggestion
* Added data hashing step using `rlang` instead of `digest` package to calculate the hash (which has been moved from `teal.data` and `teal.slice`). There is now an explicit hashing check in the reproducible code output 
```

R/module_nested_tabs.R
```diff
@@ -255,13 +256,32 @@ srv_nested_tabs.teal_module <- function(id, datasets, modules, reporter) {
     }
   )
 
+  hashes <- calculate_hashes(datanames, datasets)
```
nikolas-burkoff, line 259: I think the hashes should be calculated earlier than this as here we are recalculating hashes for each module separately as they will all have different datanames - I think `.datasets_to_data` should have an extra argument hashes which is a named list of hashes for all datasets and then get_datasets_code only outputs the hashes of those needed anyway?

nikolas-burkoff (commented):
Let me know what you think about these:

R/module_nested_tabs.R
nikolas-burkoff: though on the other hand if hashing is quick then the extra complexity of moving it up the chain is probably not worth it...

mhallal1: The good place to calculate the hashes once would be in `modules_tabs_with_filters` but we will then have to calculate the hashes twice in `ui_modules_tabs_with_filters` and `srv_modules_tabs_with_filters` and propagate them as arguments in both `ui_srv_nested_tabs` and `srv_nested_tabs` as `.datasets_to_data` is called in both `ui_srv_nested_tabs.teal_module` and `srv_nested_tabs.teal_module`.
We decided with Nik that we will keep the implementation as it is now as the calculation cost is negligible.

nikolas-burkoff:
Looking at the reporter vignette in teal adding the report adds the src:
image

image

R/get_rcode.R
```diff
@@ -189,11 +183,7 @@ get_datasets_code <- function(datanames, datasets) {
     "\n\n"
   )
 
-  str_filter <- teal.slice::get_filter_expr(datasets, datanames)
-  if (str_filter != "") {
-    str_filter <- paste0(str_filter, "\n\n")
```
nikolas-burkoff, line 194: so we are losing this - does that matter?

nikolas-burkoff: tbh probably not

mhallal1: I tested with empty filters, no problem.

nikolas-burkoff (approved):

nikolas-burkoff:
just double check the internal roxygen mentions hashes for the funcitons which now have them

Issue #756, opened 2022-10-12
Remove soft deprecated function `bookmarkableShinyApp`


donyunardi:

donyunardi:
### Acceptance Criteria:
* Remove the code
* Update NEWS
* Also need to remove https://github.com/insightsengineering/teal/blob/e96c8dac314033d0f43917a8f4ab605417b99ae9/R/module_teal.R#L160-L189

PR #782, opened 2022-11-22
Remove soft deprecated function `bookmarkableShinyApp`

chlebowa:
Closes [this](https://github.com/insightsengineering/teal/issues/756).


NEWS.md
```diff
@@ -14,8 +14,10 @@
 * Updated examples to use `scda.2022`.
 * Added R session information into a link in the footer of `teal` applications.
 * Added data hashing step using `rlang` instead of `digest` package to calculate the hash (which has been moved from `teal.data` and `teal.slice`). There is now an explicit hashing check in the reproducible code output.
+* Removed deprecated function `root_modules`.
```
mhallal1, line 17: The NEWS entry has been added here but was not removed from line 41, please remove it.

mhallal1: and deprecated functions can be grouped in a single entry.

donyunardi (changes requested):
I got this error message when trying to run a simple teal app:

```r
example("init", package="teal")
```

image

Which was caused by this line:
https://github.com/insightsengineering/teal/blob/07d17d133c253dbcad46dc01198eb613aec11327/R/module_teal.R#L210

Maybe we can just remove this `if` statement?

chlebowa:
> (...) caused by this line:
> 
> https://github.com/insightsengineering/teal/blob/07d17d133c253dbcad46dc01198eb613aec11327/R/module_teal.R#L210
> 
> Maybe we can just remove this `if` statement?

I believe was part of the deprecated bookmarking mechanism. Removed altogether. Tests pass, example app works.

donyunardi (approved):
I saw this warning when adding teal.reporter card in exploratory app:
```r
[WARNING] Deprecated: --self-contained. use --embed-resources --standalone
```

I think this is pandoc related.

Otherwise, LGTM!

PR #786, opened 2022-12-01
`shinyvalidate` improvements

chlebowa:
Related to [this issue](https://github.com/insightsengineering/teal.osprey/issues/185)

Adds the `validate_inputs` functions to `teal` so they can be called in modules.


tests/testthat/test-gather_fails.R
```diff
@@ -0,0 +1,469 @@
+library(shiny)
+library(shinyvalidate)
```
nikolas-burkoff, line 2: Don't add library here - if the packages are in the depends/suggests/imports then it's not needed

nikolas-burkoff (commented):
- As we're using `shinyvalidate` in the examples we need to add it to SUGGESTS in the DESC file

- I'm not sure about the name of the function - as we are not just gathering any validation failures but also calling shiny::validate as well - not sure what to call it instead though but making it clear there's "validation" going on in the function is probably a good idea? 

- I wonder if there's a way to use S3 dispatching here so that users of these functions only need to know one rather than having to look up which of the three they need to use in their case?

- I don't think there's any case where we want to call these functions but not call shiny::validate (i.e. just get the messages and then do something with them) - the reason I ask is by coupling the gathering to the validation we can't get the messages out if we need them. We could decouple completely and only have the gathering messages here and let the module developer add the shiny::validate themselves?

chlebowa:
> I wonder if there's a way to use S3 dispatching here so that users of these functions only need to know one rather than having to look up which of the three they need to use in their case?

I thought about it. 
One solution I see would be to get rid of `gather_fails_com` and have one method for `InputValidator` and another for a `list`, the latter with an extra argument for one of multiple blocks of messages.
Or we could go with just `gather_fails_grp` and always pass a list, with an argument to control grouping. But then there is no good way to have a default header.
I don't particularly like any of these but that's probably a matter of opinion.

<hr>

> I'm not sure about the name of the function - as we are not just gathering any validation failures but also calling shiny::validate as well - not sure what to call it instead though but making it clear there's "validation" going on in the function is probably a good idea?

I'm open to suggestions. The name is derived from the original design, where it would only collate messages and put into a `validate` call.

> I don't think there's any case where we want to call these functions but not call shiny::validate (i.e. just get the messages and then do something with them) - the reason I ask is by coupling the gathering to the validation we can't get the messages out if we need them. We could decouple completely and only have the gathering messages here and let the module developer add the shiny::validate themselves?

That was the original design, the function would only put messages together and then it would be called with `validate(need(is.null(gather_fails(iv)), gather_fails(iv))`. I wanted to wrap `is.null` inside the function.

tests/testthat/test-gather_fails.R
chlebowa: Unless it's in Depends, should it not be attached for the example to work?

nikolas-burkoff: So if its in suggests or imports then you can use `shinyvalidate::` inside the tests 

image

we don't need to guard against using suggests functions in the tests, in the same way we don't need to for the examples  for R cmd check to pass

chlebowa: Pardon me, I phrased my question poorly, I meant "for the example to work in its current form".
Yes, the namespace prefix precludes the need for attaching but that is what I explicitly wanted to avoid. It seems using `library` in examples is common practice, no?

nikolas-burkoff: in examples yes, but in test files no?

tests/testthat/test-gather_fails.R
```diff
@@ -0,0 +1,469 @@
...
+      )
+    })
+  }
+
+  shiny::testServer(server, {
+    session$setInputs(
+      "letter" = "a",
+      "number" = 2L
+    )
+    expect_error(values())
```
nikolas-burkoff, line 114: I wonder if we should explicitly expect a shiny validate error class

tests/testthat/test-gather_fails.R
```diff
@@ -0,0 +1,469 @@
...
+    )
+    expect_identical(values(), list(
+      "letter" = input[["letter"]],
+      "number" = input[["number"]]
+    ))
+  })
+})
+
+
+test_that("gather_fails: invalid inputs raise errors in output", {
```
nikolas-burkoff, line 94: I think we should have `testthat::` where needed throughout the file

tests/testthat/test-gather_fails.R
nikolas-burkoff: ah I see we have it in later tests

tests/testthat/test-gather_fails.R
```diff
@@ -0,0 +1,469 @@
...
+        "choose a value between 0.5 and 3",
+        "\n"
+      ),
+      collapse = "\n"
+    ))
+  })
+})
+
+
+test_that("hierarchical validation", {
```
nikolas-burkoff, line 374: better name here?

tests/testthat/test-gather_fails.R
chlebowa: done

tests/testthat/test-gather_fails.R
chlebowa: sequential?

tests/testthat/test-gather_fails.R
chlebowa: 😁

chlebowa:
> > I'm not sure about the name of the function - as we are not just gathering any validation failures but also calling shiny::validate as well - not sure what to call it instead though but making it clear there's "validation" going on in the function is probably a good idea?
> 
> I'm open to suggestions. The name is derived from the original design, where it would only collate messages and put into a `validate` call.

Changed to `validate_inputs` and `validate_inputs_segregated`.

R/validate_inputs.R
```diff
@@ -0,0 +1,164 @@
...
+#' and prints messages grouped by validator. If elements of \code{validators} are named,
+#' the names are used as headers for their respective message groups.
+#'
+#'
+#' @name validate_inputs
+#'
+#' @param ... for \code{validate_inputs} any number of `InputValidator` objects \cr
+#'            for \code{validate_inputs_segregated} arguments passed to `shiny::validate`
+#' @param header `character(1)` optional generic validation message
+#' @param validators optionally named `list` of `InputValidator` objects, see \code{Details}
```
nikolas-burkoff, line 29: I'd replace "optionally" with "either" - as optionally sort of suggests you don't have to use this argument

R/validate_inputs.R
```diff
@@ -0,0 +1,164 @@
...
+#' send input validation messages to output
+#'
+#' Captures messages from `InputValidator` objects and collates them
+#' into one message passed to `validate`.
+#'
+#' `shiny::validate` is used to withhold rendering of an output element until
+#' certain conditions are met and a print a validation message in place
+#' of the output element.
+#' `shinyvalidate::InputValidator` allows to validate input elements
+#' and display specific messages in their respective input widgets.
```
nikolas-burkoff, line 11:
```suggestion
#' `shinyvalidate::InputValidator` allows the validation of input elements
#' and to display specific messages in their respective input widgets.
```

R/validate_inputs.R
```diff
@@ -0,0 +1,164 @@
...
+#'
+#' @param ... for \code{validate_inputs} any number of `InputValidator` objects \cr
+#'            for \code{validate_inputs_segregated} arguments passed to `shiny::validate`
+#' @param header `character(1)` optional generic validation message
+#' @param validators optionally named `list` of `InputValidator` objects, see \code{Details}
+#'
+#' @return
+#' Returns NULL if the final validation call passes and a `shiny.silent.error` if it fails.
+#'
+#' @seealso \code{[shinyvalidate::InputValidator]} \code{[shiny::validate]}
```
nikolas-burkoff, line 34:
```suggestion
#' @seealso [`shinyvalidate::InputValidator`],  [`shiny::validate`]
```

I think we can do this?

R/validate_inputs.R
```diff
@@ -0,0 +1,164 @@
...
+#'
+#' @examples
+#' library(shiny)
```
nikolas-burkoff, line 37: teal depends on shiny so we don't actually need this line?

nikolas-burkoff (commented):
Example unhappy (also I may change "hierarchical" to "sequential")

image

Otherwise lots of minor tweaks to the roxygen etc. but all looking good - once example works I'll approve and we can merge

NEWS.md
```diff
@@ -1,5 +1,9 @@
 # teal 0.12.0.9011
 
+### New features
+
+* Added the `validate_inputs` function that produces informative error messages in app output.
```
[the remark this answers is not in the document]
chlebowa: > I may also move this below the "Major breaking changes" section but since we're going to tidy up the NEWS before the next release i think it's fine

Is it not a "new feature"?

_pkgdown.yml
```diff
@@ -52,6 +52,7 @@ reference:
   - title: Validation functions
     contents:
       - starts_with("validate_")
+      - starts_with("gather_fails")
```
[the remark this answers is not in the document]
chlebowa: Only at the bottom of the file use `@keywords internal`, pkgdown breaks if the main ones are not added here.

R/validate_inputs.R
chlebowa: It's not an "either", it must be a list with optional names.

R/validate_inputs.R
```diff
@@ -0,0 +1,164 @@
+
+#' send input validation messages to output
+#'
+#' Captures messages from `InputValidator` objects and collates them
+#' into one message passed to `validate`.
+#'
+#' `shiny::validate` is used to withhold rendering of an output element until
+#' certain conditions are met and a print a validation message in place
```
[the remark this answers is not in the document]
chlebowa: We are both wrong here. It should read "is used to withhold (...) and print".

R/validate_inputs.R
nikolas-burkoff: ah - yes I misread the "of" for "or"

NEWS.md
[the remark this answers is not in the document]
nikolas-burkoff: it is yes - I would move the whole New feature section below the "we have changed how teal modules work after 5 years" breaking change mention. But this can all be done later

_pkgdown.yml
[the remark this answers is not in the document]
chlebowa: Yes, but I hear they are to deprecated ("are we still using those?") so maybe it makes sense to add an extra line for these two?

R/validate_inputs.R
```diff
@@ -0,0 +1,164 @@
+
+#' send input validation messages to output
+#'
+#' Captures messages from `InputValidator` objects and collates them
```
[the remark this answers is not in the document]
nikolas-burkoff: nah it's ok

_pkgdown.yml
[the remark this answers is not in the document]
nikolas-burkoff: > Yes, but I hear they are to deprecated ("are we still using those?")

They are used quite a bit downstream so I doubt they will be going.. but happy for you to add these extra two explicitly - as long as pkgdown doesn't then output them twice :)

R/validate_inputs.R
chlebowa: Yes! Thank you! I could NOT get the links to work 😠

R/validate_inputs.R
chlebowa: Just so the example is self-contained.

nikolas-burkoff (approved):
👍

asbates:
Copied over from [tmc](https://github.com/insightsengineering/teal.modules.clinical/pull/699#issuecomment-1339812277): 

I'm wondering if the messages in the main panel should be highlighted more. I know the inputs themselves have red. I'm not entirely sure it would be obvious to all users that all the errors are in the main panel. Except maybe if they are already familiar with the modules.

NEWS.md
```diff
@@ -8,6 +8,9 @@
 
 * Due to deprecation of `chunks` in `teal.code`, the `teal` framework now uses their replacement (`qenv`) instead. The documentation in `teal` has been updated to reflect this and custom modules written with `chunks` should be updated to use `qenv`.
 
+### New features
+
+* Added the `validate_inputs` and `validate_inputs_segregated` functions that produce informative error messages in app output.
```
mhallal1, line 13:
```suggestion
* Added the `validate_inputs` and `validate_inputs_segregated` functions that produce informative `shinyvalidate` error messages in the app output.
```

R/validate_inputs.R
```diff
@@ -0,0 +1,164 @@
+
+#' send input validation messages to output
```
mhallal1, line 2:
```suggestion
#' Send `shinyvalidate` input validation messages to output
```

R/validate_inputs.R
```diff
@@ -0,0 +1,164 @@
...
+#'
+#' Captures messages from \code{InputValidator} objects and collates them
```
mhallal1, line 4:
```suggestion
#' Captures messages from \code{shinyvalidate::InputValidator} objects and collates them
```

R/validate_inputs.R
```diff
@@ -0,0 +1,164 @@
...
+#' into one message passed to \code{validate}.
```
mhallal1, line 5:
```suggestion
#' into one message passed to \code{shiny::validate}.
```

R/validate_inputs.R
mhallal1: I think it would be a good idea to link to `shinyvalidate` functions as they are external to `teal`, what do you think @chlebowa ? Maybe not all of them but at least the description. For this we can use the format `[shinyvalidate::InputValidator()]`

mhallal1: and is there a reason to use `\code{}` here? I do see any instance of this usage in `teal`

R/validate_inputs.R
```diff
@@ -0,0 +1,164 @@
...
+  failings <- unlist(fail_messages)
+
+  shiny::validate(shiny::need(is.null(failings), failings), ...)
+}
+
+
+### internal functions
+
+#' @keywords internal
+# internal used by all methods
```
mhallal1, line 137:
```suggestion
```

chlebowa:
> Copied over from [tmc](https://github.com/insightsengineering/teal.modules.clinical/pull/699#issuecomment-1339812277): I'm wondering if the messages in the main panel should be highlighted more. I know the inputs themselves have red. I'm not entirely sure it would be obvious to all users that all the errors are in the main panel. Except maybe if they are already familiar with the modules.
> 
> I can work on this tomorrow if no one else has gotten to it.

Highlighted how?

Validation messages come with a CSS class and `validate_inputs_segregated` can further modify it.

R/validate_inputs.R
chlebowa: done

R/validate_inputs.R
mhallal1: @chlebowa `\code{}` is the old way of assigning code, we now use ``, hence there is no occurence of `\code{}` in `teal`.
Could you please re-open the PR and correct them to conserve the good practice?

asbates:
> > Copied over from [tmc](https://github.com/insightsengineering/teal.modules.clinical/pull/699#issuecomment-1339812277): I'm wondering if the messages in the main panel should be highlighted more. I know the inputs themselves have red. I'm not entirely sure it would be obvious to all users that all the errors are in the main panel. Except maybe if they are already familiar with the modules.
> > I can work on this tomorrow if no one else has gotten to it.
> 
> Highlighted how?
> 
> Validation messages come with a CSS class and `validate_inputs_segregated` can further modify it.

Maybe something similar to highlighting of inputs: red text and red box. And/or larger font size. The fact that validation messages have a class would make this fairly easy. I'll open another issue and we can address this later.

PR #792, opened 2022-12-30
add function for conditional validation

chlebowa:
Introducing `crule` function to modify rule funcitons for input validation by `shinyvalidate`.


R/utils.R
```diff
@@ -33,3 +35,65 @@ get_teal_bs_theme <- function() {
...
+  checkmate::assert_multi_class(rule, c("function", "formula"))
+  checkmate::assert_class(substitute(condition), "call")
+
+  if (inherits(rule, "formula")) {
+    rule <- rlang::as_function(rule)
+  }
+
+  dep <- paste(deparse(substitute(condition)), collapse = "")
+  ex <- list(
+    str2lang(sprintf("if (isFALSE(%s)) return(NULL)", dep)),
```
gogonzo, line 94: you can work with language directly like this:
```r
cond <- quote(a == "b")
substitute(if (xxx) "true" else "false", list(xxx = cond))
# if (a == "b") "true" else "false"

```

```suggestion
  cond <- substitute(condition) # language
  ex <- list(
    substitute(if (isFALSE(cond)) return(NULL), list(cond = cond)),
```

R/utils.R
```diff
@@ -33,3 +35,65 @@ get_teal_bs_theme <- function() {
...
+    body(rule)
+  )
+  body(rule) <- as.call(c(as.name("{"), ex))
+  rule
```
gogonzo, line 98: You don't have to convert to character - use substitute instead

```suggestion
  body(rule) <- call(
    "{",
    substitute(if (isFALSE(condition)) return(NULL)),
    body(rule)
  )
  rule
```

chlebowa: Thanks, I tried and failed to do just that 👍

chlebowa:
After deliberating with @gogonzo we decided that this is syntactic sugar and that we can do without it.

PR #793, opened 2023-01-02
fix bug in `validate_inputs`

chlebowa:
`validate_inputs` captures messages even if validator is not enabled
this change rectifies it


R/validate_inputs.R
```diff
@@ -108,6 +108,11 @@ validate_inputs <- function(..., header = "Some inputs require attention") {
   lapply(vals, checkmate::assert_class, "InputValidator")
   checkmate::assert_string(header, null.ok = TRUE)
 
+  if (any(vapply(vals, function(iv) isFALSE(iv$.__enclos_env__$private$enabled), logical(1L)))) {
```
nikolas-burkoff, line 111: Can we have a helper function to abstract out this

```
is_enabled <- function(iv) {
  isTRUE(iv$.__enclos_env__$private$enabled
} 
```
Then we can have

`if (!all(vapply(vals, is_enabled, logical(1))`  etc. and can simplify `Filter(is_enabled, vals)`

and if shinyvalidate change their implementation we only need to change in one place.

nikolas-burkoff (commented):
Can we add a test or two please :)

R/validate_inputs.R
nikolas-burkoff: wonder if it's worth adding an issue into  shinyvalidate to get `iv$is_enabled()` added for us...

chlebowa: Done and done.

nikolas-burkoff (approved):
I'd adapt the test so that as well as there being warnings, it checks that you don't get a shiny validate error in the case that everything is good in all non-disabled validators 

But yup looks good

R/validate_inputs.R
```diff
@@ -108,6 +108,11 @@ validate_inputs <- function(..., header = "Some inputs require attention") {
   lapply(vals, checkmate::assert_class, "InputValidator")
   checkmate::assert_string(header, null.ok = TRUE)
 
+  if (!all(vapply(vals, validator_enabled, logical(1L)))) {
+    warning("Some validators are disabled and will be omitted.", call. = TRUE)
+  }
+  vals <- Filter(validator_enabled, vals)
```
nikolas-burkoff, line 114: I guess you could move this inside the if statement (and in same case below)

R/validate_inputs.R
```diff
@@ -108,6 +108,11 @@ validate_inputs <- function(..., header = "Some inputs require attention") {
   lapply(vals, checkmate::assert_class, "InputValidator")
   checkmate::assert_string(header, null.ok = TRUE)
 
+  if (!all(vapply(vals, validator_enabled, logical(1L)))) {
+    warning("Some validators are disabled and will be omitted.", call. = TRUE)
```
nikolas-burkoff, line 112: Should we use a  WARN from teal.logger here? an R warning may be a bit too much

chlebowa:
> I'd adapt the test so that as well as there being warnings, it checks that you don't get a shiny validate error in the case that everything is good in all non-disabled validators
> 
> But yup looks good

So the test for `validate_inputs` has only one validator but the one for `validate_inputs_segregated` has one validator enabled and another one disabled, both passing. Is this what you mean?

I should probably add this to the former, too.

R/validate_inputs.R
kpagacz: You probably should stick to just one logging channel. Either use the built-in R messages and warnings or use the logger solution.

kpagacz: The reasoning is that logger outputs can be piped to a file without R warnings. If someone was to do it, would they miss this warning?

R/validate_inputs.R
```diff
@@ -161,3 +170,10 @@ gather_and_add <- function(iv, header) {
   failings <- add_header(fail_messages, header)
   failings
 }
+
+#' @keywords internal
+#' test if an InputValidator object is enabled
+#' returns logical of length 1
```
kpagacz, line 176: that's a sloppy doc string :P

chlebowa: My mistake.

R/validate_inputs.R
chlebowa: `teal.logger` it is then. Any advice on how to write unit tests for it?

gogonzo: I've checked this issue here and situation is not that simple for me, so let me know if you have some solution or a strong opinion about how to handle this situation:

It's not straightforward how to test modules which return warnings through the `logger`. `options(TEAL.LOG_LEVEL)` + `capture_output` in tests doesn't seem to be optimal solution. I've checked there is no function in the `logger` to "convert" logger-conditions into the exceptions. 

With @chlebowa we tried to find the way to temporary (for tests) turn logger entries into R errors, warnings, messages and it's also not so easy neither. Currently, it's possible to do access `level` and `msg` in [layout_teal_glue_generator](https://github.com/insightsengineering/teal.logger/blob/b2108770072f8def9d5e372cd306ba9e51a062ae/R/register_logger.R#L106) via `register_logger` which would have to be called once again especially for tests.

Alternative solution (not for today) is to use `logger::log_errors`, `logger::log_warnings` and change all `logger::log_warn` in teal packages to generic `warning`. How would it work then? - `log_errors` uses `withGlobalCallingHandlers` and intercept errors (their messages) and append them into logger. Our modules then can work normally with `warning`, `stop` and they could be appended into logger if one adds `logger::log_errors` on top of the `app.R` file.

chlebowa: As this is not a trivial issue, I have modified the tests to test console output as a temporary measure.

_pkgdown.yml
```diff
@@ -52,7 +52,7 @@ reference:
   - title: Validation functions
     contents:
       - starts_with("validate_")
-      - starts_with("validate_inputs")
+      - validate_inputs
```
nikolas-burkoff, line 55: Probably don't need this line as validate_inputs starts with "validate"

tests/testthat/test-validate_inputs.R
```diff
@@ -1,142 +1,303 @@
...
-    iv$add_rule("letter", shinyvalidate::sv_in_set(LETTERS, "choose a capital letter"))
-    iv$add_rule("number", ~ if (as.integer(.) %% 2L == 1L) "choose an even number")
-    iv$enable()
-    values <- shiny::reactive({
-      validate_inputs_segregated(list(iv))
-      list(
-        "letter" = input[["letter"]],
-        "number" = input[["number"]]
+testthat::test_that(
+  "validate_inputs_segregated: disabled validators raise warnings (validator list)",
```
nikolas-burkoff, line 38: still got tests with name `validate_inputs_segregated`

R/validate_inputs.R
```diff
@@ -101,63 +99,81 @@
 #'   shinyApp(ui, server)
 #' }
```
nikolas-burkoff, line 101:
```suggestion
```

R/validate_inputs.R
```diff
@@ -101,63 +99,81 @@
...
+# recursive object type test
+# returns logical of length 1
+is_validators <- function(x) {
+  all(if (is.list(x)) unlist(lapply(x, is_validators)) else inherits(x, "InputValidator"))
 }
 
+#' @keywords internal
+# test if an InputValidator object is enabled
+# returns logical of length 1
+validator_enabled <- function(x) {
```
nikolas-burkoff, line 130: I would add comment with the shinyvalidate issue here

nikolas-burkoff (commented):
A few small comments and a bit more work needed on the tests I think

tests/testthat/test-validate_inputs.R
```diff
@@ -81,33 +186,71 @@ testthat::test_that("validate_inputs: invalid inputs raise errors in output", {
...
+
+testthat::test_that("invalid inputs raise errors in output (nested validator list)", {
   server <- function(input, output, session) {
     iv <- shinyvalidate::InputValidator$new()
     iv$add_rule("letter", shinyvalidate::sv_in_set(LETTERS, "choose a capital letter"))
     iv$add_rule("number", ~ if (as.integer(.) %% 2L == 1L) "choose an even number")
     iv$enable()
     values <- shiny::reactive({
-      validate_inputs_segregated(list(iv))
+      validate_inputs(list(iv))
```
nikolas-burkoff, line 253: Here you have list(iv) on line 148 you have list(list(iv)) ?

nikolas-burkoff (approved):

tests/testthat/test-validate_inputs.R
chlebowa: good catch!

PR #795, opened 2023-01-10
convert check to assert

pawelru:
refers to https://github.com/insightsengineering/nestdevs-tasks/issues/3
fix checks so that we fail instead of just return false that is not used at all

### DIFF
--- a/R/init.R
+++ b/R/init.R
@@ -161,13 +161,16 @@ init <- function(data,
 
   checkmate::assert_string(title, null.ok = TRUE)
   checkmate::assert_class(data, "TealData")
-  checkmate::assert_multi_class(modules, c("list", "teal_modules"))
+  checkmate::assert_multi_class(modules, c("teal_module", "list", "teal_modules"))
   checkmate::assert_list(filter, min.len = 0, names = "unique")
   checkmate::assert_subset(names(filter), choices = teal.data::get_dataname(data))
   checkmate::assert_character(id, max.len = 1, any.missing = FALSE)
 
   teal.logger::log_system_info()
 
+  if (is(modules, "teal_module")) {
+    modules <- teal::modules(modules)
+  }
   if (is(modules, "list")) {
     modules <- do.call(teal::modules, modules)
   }
--- a/R/init.R
+++ b/R/init.R
@@ -161,8 +161,7 @@ init <- function(data,
 
   checkmate::assert_string(title, null.ok = TRUE)
   checkmate::assert_class(data, "TealData")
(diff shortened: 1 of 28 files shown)


R/init.R
```diff
@@ -161,8 +161,7 @@ init <- function(data,
 
   checkmate::assert_string(title, null.ok = TRUE)
   checkmate::assert_class(data, "TealData")
-  checkmate::check_list(modules)
-  checkmate::check_class(modules, "teal_modules")
+  checkmate::assert_multi_class(modules, c("list", "teal_modules"))
```
nikolas-burkoff, line 164: So I think we should do this rather than allow a list:

```
checkmate::assert_multi_class(modules, c("teal_modules", "teal_modules"))
if (inherits(modules, "teal_modules") modules <- modules(modules)
```

We should probably update an example or two (e.g. the one for example_modules() ) to not use `modules()` and also update NEWS

pawelru: This would be a new feature and I tried to fix the bug only but in general you are right - that would be a nice enhancement. Let me do that.

nikolas-burkoff (approved):
I updated the docs so that init says `modules` can be of the appropriate type which then meant I needed to update the docs of other modules as for those they cannot be list of teal_module.

R/init.R
```diff
@@ -161,16 +163,14 @@ init <- function(data,
...
+  checkmate::check_class(modules, "teal_modules")
   checkmate::assert_list(filter, min.len = 0, names = "unique")
   checkmate::assert_subset(names(filter), choices = teal.data::get_dataname(data))
   checkmate::assert_character(id, max.len = 1, any.missing = FALSE)
 
   teal.logger::log_system_info()
 
-  if (is(modules, "teal_module")) {
-    modules <- teal::modules(modules)
-  }
```
pawelru: hmmm you reverted my changes - was that intentional?

nikolas-burkoff: Nope - whoops

nikolas-burkoff: ok so this is weird if you look here: https://github.com/insightsengineering/teal/blob/checkmate_check_to_assert/R/init.R#L174 the changes are on this branch ??

R/init.R
```diff
@@ -169,7 +171,7 @@ init <- function(data,
 
   teal.logger::log_system_info()
 
-  if (is(modules, "list")) {
+  if (is(modules, "teal_module")) {
```
nikolas-burkoff, line 174:
```suggestion
  if (is(modules, "teal_module") || is(modules, "list")) {
```

PR #798, opened 2023-01-11
Update bootstrap vignette

### LINKED ISSUE #738: Enhance vignette for Bootstrap when using `run_with_themer` and `bs_theme_update`
As a user of the themes, I would hope to also have another example run with `bslib::bs_theme_update()` - once I explored with `bslib::run_with_themer()` and choose my own colors/themes, an example of how to update the app with the previously chosen theme would be helpful.

_Originally posted by @kumamiao in https://github.com/insightsengineering/teal/pull/727#discussion_r981686873_

asbates:
Updates the bootstrap theming vignette to add more details on using `bslib::run_with_themer` to customize a `teal` application theme. Includes screenshots for each step as well as explanation of how to apply the new theme.

Fixes #738


donyunardi (changes requested):
The contents are good.
One small request, can we do a better circle annotation on the images/screenshots?
In Mac, I could do this using Preview. 
I think in Windows you can do something like MS Paint?

asbates:
> The contents are good. One small request, can we do a better circle annotation on the images/screenshots? In Mac, I could do this using Preview. I think in Windows you can do something like MS Paint?

You don't like my masterful hand drawing? :). Do these new images look better?

donyunardi (approved):

PR #688, opened 2022-07-08
fix test

mhallal1:
fixes #687 

Same approach as https://github.com/insightsengineering/teal/pull/684/files

PR #800, opened 2023-01-19
fix failing tests on R 4.3

gogonzo:
closes https://github.com/insightsengineering/coredev-tasks/issues/427


nikolas-burkoff:
I think we need to remove the time-date from the comparison though as that has caused failed tests when the calls to Sys.time() end up returning different seconds https://github.com/insightsengineering/teal/pull/688

nikolas-burkoff (approved):

PR #801, opened 2023-01-25
Add diagram, update NEWS

donyunardi:
fixes insightsengineering/coredev-tasks#427

My initial plan was to remove the documentation from `srv_nested_tabs` that pertains to the `uml_nested_tabs.jpg`.
Latex didn't like the underscore in the file name and this figure doesn't seem to exist in teal repo.

However, I found this ticket #391.

The documentation in `srv_nested_tabs` feels like a copy from the original post so I thought the image there is the missing diagram. 

I decided to just add the image and make sure not using underscore in the jpg file name.


gogonzo (approved):
I'm not sure about the diagram. @donyunardi do you think it helps to understand the nested tabs more? If you think it is I approve

donyunardi:
Per conversation, I'm removing the diagram information since it's outdated and it's an internal function.

donyunardi:
dry-run in R/dev again and it passed.
Merging.

PR #804, opened 2023-02-07
803 data reactivity@main

### LINKED ISSUE #803: [BUG]: teal apps are slow and vulnerable to the change in the data
If you run any app containing n-modules (eg. `exploratory`) please do following:
1. Turn on logs `options(teal.log_level = "TRACE", teal.show_js_log = TRUE)`
2. Run the app
3. Add and change any filter value and see what happens in the logs - you can see logs being called from many modules.

We have following problem:
- All modules which have observe or observeEvent they trigger calculation of the reactives to get needed values. Normally it's not a problem with the outputs, which trigger re-calculation of the reactives when they are visible. For example renderPlot in any module triggers only when we change the tab.
- We were vulnerable to this problem for years and solution might be very simple - to guard reactives passed to the modules by `teal` with simple `eventReactive` based on the change of the tabs.

nikolas-burkoff:
Closes #803 

Lots of TODOs in the code - especially the make id's unique of modules in teal::init

```
library(teal)

delay_module <- function(label = "delay", sleep_time = 0.5) {
  module(label = label, server = function(id, data) {

      observeEvent(data[[1]](), {
        print(sprintf("trigger for module %s of %f", label, sleep_time))
        Sys.sleep(sleep_time)
        print(sprintf("trigger for module %s of %f completed", label, sleep_time))
      })

    })
  })
}


my_iris <- iris
my_iris$Species[1] <- NA

app <- init(
  data = teal_data(dataset("IRIS", my_iris )),
  modules = modules(
    delay_module("A"),
    modules(label = "Tab2", delay_module("B")),
    modules(label = "Tab3", delay_module("C"), delay_module("D")),
    delay_module("E")    
  )
)

runApp(app)
```


R/init.R
```diff
@@ -177,6 +177,8 @@ init <- function(data,
     modules <- do.call(teal::modules, modules)
   }
 
+  # TODO make module names unique here
```
nikolas-burkoff, line 180: best to do this after the reporter module has been added

gogonzo:
> I tested this branch with the exploratory app, it seems that the reactivity is not triggered there when changing the tabs of modules so none of the modules is calculated at the beginning or when changing.

I had the same but then I've tried individual modules and they are working. Problem found in the sample apps is that they have first tab which doesn't need data. When I removed front-page and dataviewer modules everything was fine. So, there is some problem with init.

mhallal1:
> > I tested this branch with the exploratory app, it seems that the reactivity is not triggered there when changing the tabs of modules so none of the modules is calculated at the beginning or when changing.
> 
> I had the same but then I've tried individual modules and they are working. Problem found in the sample apps is that they have first tab which doesn't need data. When I removed front-page and dataviewer modules everything was fine.

Agreed, plus some `shinyvalidate` errors in regression and bivariate modules.

nikolas-burkoff:
Also I think a few modules (PCA, outliers,...) aren't happy

donyunardi:
I also experienced the same thing that were mentioned.
I guess we can close this without merging in favor of #806?

mhallal1:
> I also experienced the same thing that were mentioned. I guess we can close this without merging in favor of #806?

If we get agreement on #806 then we will close this one.

nikolas-burkoff:
Closed in favor of #806

PR #806, opened 2023-02-13
teal reactivity solution 2

### LINKED ISSUE #803: [BUG]: teal apps are slow and vulnerable to the change in the data

mhallal1:
closes #803 

Solution 2 to solve the reactivity of teal apps:
1) Transformed `srv_nested_tabs.teal_module` into `moduleServer`
2) Added a NULL `renderUI` in `srv_nested_tabs.teal_module` to trigger the filtered data when the tab is changed
3) Updated `.datasets_to_data` to accept a `reactiveVal` trigger function to refresh the filtered data
4) Updated and added a couple of tests

Please test with exploratory app or similar apps.

### DIFF
--- a/R/module_nested_tabs.R
+++ b/R/module_nested_tabs.R
@@ -88,14 +88,14 @@ ui_nested_tabs.teal_modules <- function(id, modules, datasets, depth = 0L) {
 #' @export
 #' @keywords internal
 ui_nested_tabs.teal_module <- function(id, modules, datasets, depth = 0L) {
+  ns <- NS(id)
   checkmate::assert_class(datasets, "FilteredData")
   args <- isolate(teal.transform::resolve_delayed(modules$ui_args, datasets))
-  args <- c(list(id = id), args)
+  args <- c(list(id = ns("module")), args)
 
   if (is_arg_used(modules$ui, "datasets")) {
     args <- c(args, datasets = datasets)
   }
-
 
   if (is_arg_used(modules$ui, "data")) {
     data <- .datasets_to_data(modules, datasets)
@@ -105,6 +105,7 @@ ui_nested_tabs.teal_module <- function(id, modules, datasets, depth = 0L) {
   tags$div(
     id = id,
     class = "teal_module",
+    uiOutput(ns("data_reactive"), inline = TRUE),
     tagList(
       if (depth >= 2L) div(style = "mt-6"),
       do.call(modules$ui, args)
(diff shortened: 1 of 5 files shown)


donyunardi:
Looking at the log, it looks like module will only be processed when clicked.
However, adding filter(s) felt significantly slower than usual.
There are lots of trace generated during filtering activity.

Can you try it on your end?

mhallal1:
> Looking at the log, it looks like module will only be processed when clicked. However, adding filter(s) felt significantly slower than usual. There are lots of trace generated during filtering activity.
> 
> Can you try it on your end?

Could you be more specific with an example or a screenshot of the logs to compare?
I did not notice this behaviour.

donyunardi:
I tried it again and I don't see the behavior anymore, maybe my machine was too cluttered last night.
I tried with multiple example apps and it feels good.
Thanks for working on this!

Just for sanity check can someone else also test this out?

R/module_nested_tabs.R
```diff
@@ -179,42 +181,61 @@ srv_nested_tabs.teal_module <- function(id, datasets, modules, reporter) {
...
-  if (is_arg_used(modules$server, "filter_panel_api")) {
-    filter_panel_api <- teal.slice::FilterPanelAPI$new(datasets)
-    args <- c(args, filter_panel_api = filter_panel_api)
-  }
+    # trigger the data when the tab is selected
+    trigger_data <- reactiveVal(1L)
+    output$data_reactive <- renderUI({
+      lapply(datanames, function(x) {
+        datasets$get_data(x, filtered = TRUE)
+      })
```
gogonzo, line 207: I will measure timing of this and in the line `268: eventReactive(trigger_data(), datasets$get_data(x, filtered = TRUE))`

I hope double evaluation of `get_data()` doesn't influence anything (filter operation suppose to be cached)

gogonzo: I confirm that reactive in `FilteredDataset$dataset_filtered` is evaluated once and cached. This means that each `datasets$get_data()` takes just fractions of the seconds. So having multiple `datasets$get_data()` here doesn't have any influence on the performance.

gogonzo (commented):
I prefer this solution over https://github.com/insightsengineering/teal/pull/804 as it doesn't need any additional changes. Also module_nested_tabs are less complicated as they don't need passing `module_id` to determine what is active. Apps works as expected. I give a pre-approval

R/module_nested_tabs.R
```diff
@@ -88,9 +88,10 @@ ui_nested_tabs.teal_modules <- function(id, modules, datasets, depth = 0L) {
 #' @export
 #' @keywords internal
 ui_nested_tabs.teal_module <- function(id, modules, datasets, depth = 0L) {
+  ns <- NS(id)
   checkmate::assert_class(datasets, "FilteredData")
   args <- isolate(teal.transform::resolve_delayed(modules$ui_args, datasets))
-  args <- c(list(id = id), args)
+  args <- c(list(id = ns("module")), args)
```
nikolas-burkoff, line 94: line 102 - just checking that that's ok (for example missing_data and variable_browser have data in the UI)

nikolas-burkoff (commented):
Code looks good, I'd definitely recommend using this PR over the other one which can be closed. Will run some tests and then approve

nikolas-burkoff:
It would be great if this doesn't trigger on initialization of the app only when the tab is created

tests/testthat/test-module_nested_tabs.R
```diff
@@ -223,7 +285,7 @@ testthat::test_that("srv_nested_tabs.teal_module doesn't pass filter_panel_api i
     moduleServer(id, function(input, output, session) checkmate::assert_class(filter_panel_api, "FilterPanelAPI"))
   })
 
-  testthat::expect_error(
+  testthat::expect_warning(
```
mhallal1, line 288: I could not find a better way to improve this test. It passes but throws an error internally. 
The test does not really test what it claims it is testing.

nikolas-burkoff: So I wonder if using `rlang::env_has(environment(), "data", inherit = FALSE)` (or datasets, or filterpanelapi) would help here? - I'll test

nikolas-burkoff: for the not passed case we want this:
```
      checkmate::assert_false(
        tryCatch(
          checkmate::test_class(data, "tdata"),
          error = function(cond) FALSE
        )
      )
    })
 ```
and you replace data with datasets/filter_panel_api

for the it is passed case we can have assert_class() directly
and then expect_no_error (or expect_error(..., NA)) for the testserver

nikolas-burkoff (approved):
Once my comment inline is addressed (and if using expect_no_error then the version of testthat needs to be bumped up in the DESC file) this can be merged in

tests/testthat/test-module_nested_tabs.R
mhallal1: done.

PR #808, opened 2023-02-24
Tdata in tealdata@main

mhallal1:
related to https://github.com/insightsengineering/teal.data/issues/121

-Updated `module_Teal` and `module_nested_Tabs` to accept `tdata`
-removed `tdata` into `teal.data`
-updated tests and docs


Melkiades:
I think this needs to get in before becoming stale. I may work on this if it is still a thing to have @donyunardi

mhallal1:
> I think this needs to get in before becoming stale. I may work on this if it is still a thing to have @donyunardi

Plans might have changed meanwhile regarding `tdata`, @gogonzo can advice more on this.

Issue #818, opened 2023-04-03
Documentation fixes


asbates:
There are a couple of issues (some potential) noted in the review of documentation in #815. I think they need more discussion, hence this issue.

- The vignette "Adding support for Reporting to custom modules" uses `qenv` in an example. I think we should add a short description of qenv, point to `teal.code`, or remove the `qenv` usage possible.
- ~~The examples in `module_tabs_with_filters` uses `:::`. These shouldn't be used.~~
- I don't understand the purpose of the second example in `module_tabs_with_filters`. If it doesn't add anything helpful it should be removed. Similar examples are used in other teal functions so they may need updating as well
- ~~Similarly for `module_teal`. It uses `:::` and I'm not sure the example is helpful.~~

m7pr:
blocked by #870

PR #819, opened 2023-04-03
Documentation fixes for release review

### LINKED ISSUE #815: Review `teal` documentation before release
### Summary
Related with insightsengineering/nestdevs-tasks#13
In preparation to release `teal`, we need to review/assess all documentation in the package. 

### Acceptance Criteria
* Review roxygen examples on all functions
  * Test examples
  * If it's using chunk, update to `qenv`.
* * Review vignettes on these topics only:
  * Review any content around `qenv`
  * Review any content around `chunks` (i.e. make sure the right language around deprecation)
  * Review any content around shinyvalidate
* Update NEWS
* Update README (if applicable)
* Review DESCRIPTION file (i.e. dependency versions, etc)

asbates:
Fixes a few documentation issues that need to be addressed for release. Note that not all items from #815 are in this PR. Some of them need discussion, so I opened #818.

Closes #815


R/module_nested_tabs.R
```diff
@@ -133,14 +133,14 @@ srv_nested_tabs <- function(id, datasets, modules, reporter = teal.reporter::Rep
 #' @rdname srv_nested_tabs
 #' @export
 #' @keywords internal
-srv_nested_tabs.default <- function(id, datasets, modules, reporter) {
+srv_nested_tabs.default <- function(id, datasets, modules, reporter = teal.reporter::Reporter$new()) {
```
donyunardi, line 136: I don't think we need to define the default value during the method dispatch because it's being specified during the creation of the generic function here:
https://github.com/insightsengineering/teal/blob/e8a67d1f623c16b101b70ddd39caae2426d6d2e3/R/module_nested_tabs.R#L127-L130

However, I did look at `ui_nested_tabs` where it defines the default value during dispatch for `depth`.
For consistency, I'm fine with adding the default value during method dispatch.

R/validations.R
```diff
@@ -262,7 +262,7 @@ validate_no_intersection <- function(x, y, msg) {
 #'
 #' ui <- fluidPage(
 #'   selectInput("arm", "Select treatment",
-#'     choices = c("ARM", "ARMCD", "ACTARM", "ACTARMCD", "TRT01P", "TRT01A"),
+#'     choices = c("ARM", "ARMCD", "ACTARM", "TRT"),
```
donyunardi, line 265: Good findings.

Also, we should improve the validation message given out here:

image

The current message implies that all columns do not exist where it supposed to be just for `TRT` column.

donyunardi (approved):
Looks good to me.

R/validations.R
asbates: #821

R/module_nested_tabs.R
asbates: It's actually needed here otherwise we get an error. I don't think this is because S3 though. I think it's with how `srv_nested_tabs.teal_modules` works.  It uses the following code ([here](https://github.com/insightsengineering/teal/blob/e8a67d1f623c16b101b70ddd39caae2426d6d2e3/R/module_nested_tabs.R#L153-L155))

```r
    modules_reactive <- sapply(names(modules$children), USE.NAMES = TRUE, function(id) {
      srv_nested_tabs(id = id, datasets = datasets, modules = modules$children[[id]], reporter = reporter)
    })
```

In the call to `srv_nested_tabs`, `id` is a string. There is no method for this so it reverts back to the generic. Then it hits `inherits(reporter, "Reporter")` and it fails. There's a recursiveness here generic -> method -> generic that I think is throwing things off.

asbates: #822

PR #820, opened 2023-04-04
Update workflows

walkowif:
Resolves https://github.com/insightsengineering/idr-tasks/issues/513, https://github.com/insightsengineering/idr-tasks/issues/522 and https://github.com/insightsengineering/idr-tasks/issues/533

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -15,6 +15,13 @@ on:
       - main
   workflow_dispatch:
 
+  # Rerun workflows once pkgdown has finished running, to add coverage
+  # and unit test report to gh-pages branch.
+  workflow_run:
+    workflows: ["Docs 📚"]
+    types:
+      - completed
+
 jobs:
   audit:
     name: Audit Dependencies 🕵️‍♂️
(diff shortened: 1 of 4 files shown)


walkowif:
@cicdguy This is more or less how the propagation PR will look like.

donyunardi:
All checks are skipped. 
Will we only see the impact to teal once this is merged to main?

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,20 @@
+---
+name:  Dependency Test 🔢
+
+on:
+  schedule:
+    - cron: '45 3 * * *'
```
walkowif, line 6: @pawelru The idea is to run the workflow nightly. Is this ok?

pawelru: Up to you guys. Those are long workflows so probably doesn't make much sense to execute on every push.

.github/workflows/check.yaml
```diff
@@ -28,6 +35,10 @@ jobs:
       additional-env-vars: |
         _R_CHECK_CRAN_INCOMING_REMOTE_=false
       additional-r-cmd-check-params: --as-cran
+      enforce-note-blocklist: true
+      note-blocklist: '*'
```
pawelru, line 39: does that mean that we want to fail on any note? We don't want that. We have identified already a list of notes that we want to fail on -> https://github.com/insightsengineering/idr-tasks/issues/505

PR #824, opened 2023-04-12
723 default header@main

chlebowa:
Closes [this issue](https://github.com/insightsengineering/teal.modules.clinical/issues/723).

Changed default values of `header` and `footer` in `init` from `tags$p("Add title here")` to `tags$p()`.

### DIFF
--- a/R/init.R
+++ b/R/init.R
@@ -92,11 +92,11 @@
 #'  By adding the `filterable` attribute it is possible to control which variables can be filtered for each
 #'  dataset. See the example below where `ADSL` can only be filtered by `AGE`, `SEX` or `RACE`.
 #'
-#' @param header (`character` or `shiny.tag`) \cr
+#' @param header (`shiny.tag`) \cr
 #'   the header of the app. Note shiny code placed here (and in the footer
 #'   argument) will be placed in the app's `ui` function so code which needs to be placed in the `ui` function
 #'   (such as loading css via [htmltools::htmlDependency()]) should be included here.
-#' @param footer (`character` or `shiny.tag`)\cr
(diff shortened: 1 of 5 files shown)


R/init.R
```diff
@@ -155,8 +155,8 @@ init <- function(data,
                  modules,
                  title = NULL,
                  filter = list(),
-                 header = tags$p("Add Title Here"),
-                 footer = tags$p("Add Footer Here"),
```
Melkiades, line 159: It makes sense! Is there a way to add these interactively afterwards?

chlebowa: You mean after the app starts? I don't think so. They can be specified by the app developer.

Melkiades (approved):
Lgtm! Thanks for taking care of this. It was quite ugly. Is it not breaking downstream?

R/init.R
Melkiades: Ok, if someone wants to do that they can ask for it as a new feature probably. For the moment, it was not used imo

chlebowa:
> Lgtm! Thanks for taking care of this. It was quite ugly. Is it not breaking downstream?

I checked a few `tmc` apps and they work when `header = tags$p()`. I will double check with this branch installed.

R/init.R
Melkiades: it is not checked though (no assertion). Is it only because it is a shiny.tag?

chlebowa: There were no assertions previously. I will add them, good idea :+1:

chlebowa:
> This requires some investigation.

Fixed by reinstalling NEST.

PR #825, opened 2023-04-13
187 new api@filter panel refactor@main

chlebowa:
Introduces changes necessary to handle the new Filter Panel API coming from [this PR](https://github.com/insightsengineering/teal.slice/pull/222).


asbates:
I get an error when running the cdsic example app in insightsengineering/teal.slice/pull/165. I am on branch 187_... for both teal and teal.slice

insightsengineering/teal.slice/pull/165

donyunardi:
I think the example code fails because the `default_filters` object is not using the new API structure with `filter_settings` and `filter_vars`.

So, when I changed the `default_filters` to this:
```r
default_filters <- filter_settings(
    filter_var(dataname = "ADSL", varname = "categorical2", selected = c("a", "b")),
    filter_var(dataname = "ADSL", varname = "numeric", selected = c(0, 140), keep_na = TRUE, keep_inf = TRUE),
    filter_var(dataname = "ADTTE", varname = "AVAL", selected = c(10, 1000))
)
```

It works.

However, I ran into couple issues:

### 1. Got a new error message, but the app still work:
    

    
This is what I see in the console:
```r
Listening on http://127.0.0.1:6104
[WARN] 2023-04-13 14:18:16.3260 pid:72894 token:[cafcf5b4] teal.slice Programmatic range specification on numeric was adjusted to existing slider ticks. It is now broader in order to contain the specified values.
[WARN] 2023-04-13 14:18:16.3598 pid:72894 token:[cafcf5b4] teal.slice Programmatic range specification on AVAL was adjusted to existing slider ticks. It is now broader in order to contain the specified values.
Warning: Error in ggplot2::geom_histogram: Problem while computing aesthetics.
ℹ Error occurred in the 1st layer.
Caused by error in `check_aesthetics()`:
! Aesthetics must be either length 1 or the same as the data (381)
✖ Fix the following mappings: `x`
  238: <Anonymous>
  237: signalCondition
  236: signal_abort
  235: rlang::abort
  234: cli::cli_abort
  233: handlers[[1L]]
  232: <Anonymous>
  231: signalCondition
  230: signal_abort
  229: rlang::abort
  228: cli::cli_abort
  227: check_aesthetics
  226: compute_aesthetics
  225: l$compute_aesthetics
  224: f
  217: by_layer
  216: ggplot_build.ggplot
  214: print.ggplot
  209: func
  207: f
  206: Reduce
  197: do
  196: hybrid_chain
  168: drawPlot
  154: <reactive:plotObj>
  138: drawReactive
  125: renderFunc
  124: valueFunc
  111: valueFunc
   96: func
   94: f
   93: Reduce
   84: do
   83: hybrid_chain
   82: output$teal-main_ui-filter_panel-active-ADSL-filter-_var_numeric-card-inputs-plot
    1: runApp
Input to asJSON(keep_vec_names=TRUE) is a named vector. In a future version of jsonlite, this option will not be supported, and named vectors will be translated into arrays instead of objects. If you want JSON object output, please use a named list instead. See ?toJSON.
```

### 2. I can't seem to get the default filter working for column with logical value.
If I add the logical field to the `default_filters` with the setup below, the app crashed:
```r
default_filters <- filter_settings(
    filter_var(dataname = "ADSL", varname = "logical"),
    filter_var(dataname = "ADSL", varname = "logical1", selected = FALSE)
)
```

Maybe we should make a new issue on this but want to see if someone else can reproduce my error.

donyunardi:
I guess another question is, do we need to retain the old way to setup filter state while we introduce the new API?

asbates:
@donyunardi I definitely should have updated it to the new API. Can't believe I forgot!

Whether we need to support the old way, I think that support was added. Is that right @chlebowa ?

chlebowa:
1) Filter specification should work with both the old and the new way and it does for me. If it fails on your machine, I suggest a face to face to figure it out.
2) The second error, with the `logical` column, was caused by the `format` method and is now fixed.
3) I will look into the slider thing.

chlebowa:
slider bug is fixed

Issue #835, opened 2023-05-30
Improper indentation in get_rcode and get_rcode_utils


m7pr:
### What happened?

This is a followup after https://github.com/insightsengineering/teal.slice/issues/308

Imagine running below application:

```{R}
library(shiny)
library(teal.slice)
library(teal)
library(scda)
options(teal.log_level = "TRACE", teal.show_js_log = TRUE)

funny_module <- function (label = "Filter states", datanames = "all") {
  checkmate::assert_string(label)
  module(
    label = label,
    filters = datanames,
    ui = function(id, ...) {
      ns <- NS(id)
      div(
        h2("The following filter calls are generated:"),
        verbatimTextOutput(ns("filter_states")),
        verbatimTextOutput(ns("filter_calls")),
        actionButton(ns("reset"), "reset_to_default")
      )
    },
    server = function(input, output, session, data, filter_panel_api) {
      checkmate::assert_class(data, "tdata")
      observeEvent(input$reset, set_filter_state(filter_panel_api, default_filters))
      output$filter_states <-  renderText({
        logger::log_trace("rendering text1")
        filter_panel_api |> get_filter_state() |> yaml::as.yaml()
      })
      output$filter_calls <- renderText({
        logger::log_trace("rendering text2")
        attr(data, "code")()
      })
    }
  )
}

ADSL <- synthetic_cdisc_data("latest")$adsl

#############
# this is the variable that's causing the error when selected as filter
ADSL$empty <- NA 
#############

ADSL$logical1 <- FALSE
ADSL$logical <- sample(c(TRUE, FALSE), size = nrow(ADSL), replace = TRUE)
ADSL$numeric <- rnorm(nrow(ADSL))
ADSL$categorical <- sample(letters[1:10], size = nrow(ADSL), replace = TRUE)
ADSL$categorical2 <- sample(letters[1:3], size = nrow(ADSL), replace = TRUE, prob = c(.1, .3, .6))
ADSL$date <- Sys.Date() + seq_len(nrow(ADSL))
ADSL$date2 <- rep(Sys.Date() + 1:3, length.out = nrow(ADSL))
ADSL$datetime <- Sys.time() + seq_len(nrow(ADSL)) * 3600 * 12
ADSL$datetime2 <- rep(Sys.time() + 1:3 * 43200, length.out = nrow(ADSL))

ADSL$numeric[sample(1:nrow(ADSL), size = 10)] <- NA
[...]

PR #836, opened 2023-05-30
835 - get_rcode_*() starts new code chunks in newlines

m7pr:
This is an answer to #835 

Before the change

Capture1

After the change

capture2


m7pr:
When merging remove commented code as lintr do not pass checks for the commented code

chlebowa (approved):

Issue #834, opened 2023-05-25
Decouple `scda` from teal


donyunardi:
Related to https://github.com/insightsengineering/nestdevs-tasks/issues/1

### Summary
After careful observation, it appears that the scda package is not essential for teal. Consequently, it is advantageous for us to separate scda from teal.

### Files affected:
* dummy_functions.R
* init.R (examples)
* modules_debugging.R (examples)
* validations.R (examples)
* including-adam-data-in-teal.Rmd
* preprocessing-data.Rmd

### Acceptance Criteria
* Use iris, mtcars, or create a minimal example data that resides only within the package for unit tests and roxygen examples.
  This minimal example data will be stored as .rda files in the /data folder.
* Revise the roxygen examples to use the minimal example data.
* If applicable, Update the unit tests to utilize the new data.
* Update affected vignettes.
* Update DESCRIPTION file to remove `scda` and related packages.
* Update staged_dependencies.yml

PR #858, opened 2023-06-29
Decouple scda

gogonzo:
closes https://github.com/insightsengineering/teal/issues/834

see also:
- https://github.com/insightsengineering/scda/pull/125
- https://github.com/insightsengineering/scda.2022/pull/123


R/dummy_functions.R
```diff
@@ -53,34 +53,49 @@ get_dummy_filter <- function(data) { # nolint
...
+    stats::rnorm(10, mean = 700, sd = 200), # dummy OS level
+    stats::rnorm(10, mean = 400, sd = 100), # dummy EFS level
+    stats::rnorm(10, mean = 450, sd = 200) # dummy PFS level
+  )
 
   ADSL$logical_test <- sample(c(TRUE, FALSE, NA), size = nrow(ADSL), replace = TRUE) # nolint
-  ADSL$SEX[1:150] <- NA # nolint
+  ADSL$SEX[c(2, 5)] <- NA # nolint
+
+  cdisc_data_obj <- cdisc_data(
```
pawelru, line 73:
```suggestion
  cdisc_data_obj <- teal.data::cdisc_data(
```

R/init.R
```diff
@@ -111,41 +111,41 @@
 #' @include modules.R
 #'
 #' @examples
```
pawelru, line 113: Hmmm still thinking about the example for this `init()` function...
To me, most of the people looking at this example would came from clinical trials domain so it would be good to speak in the same language - i.e. use ADAM model. WDYT @lcd2yyz ?

R/dummy_functions.R
```diff
@@ -53,34 +53,49 @@ get_dummy_filter <- function(data) { # nolint
...
-    ADSL <- scda::synthetic_cdisc_data("latest")$adsl # nolint
-    ADAE <- scda::synthetic_cdisc_data("latest")$adae # nolint
-    ADLB <- scda::synthetic_cdisc_data("latest")$adlb # nolint
-  })
+  ADSL <- data.frame( # nolint
+    STUDYID = "study",
+    USUBJID = 1:10,
+    SEX = sample(c("F", "M"), 10, replace = TRUE),
+    AGE = stats::rpois(10, 40)
+  )
```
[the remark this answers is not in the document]
gogonzo: @pawelru it's a private function, used only in tests and documentation of other internal functions. Do you want this to still change it to `ex_cdisc_data` or something else?

R/init.R
gogonzo: I can copy paste data created in `get_dummy_data` so we can have an ADaM alike data here.

R/dummy_functions.R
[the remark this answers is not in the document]
pawelru: only if it happens that this is being defined multiple times

R/init.R
gogonzo: There is also vignette `including-adam-data-in-teal.Rmd` for people who want to use teal in clinical trials domain. I'd prefer to keep `teal::init` documentation generic.

R/dummy_functions.R
chlebowa: Even though `teal.data` is in `Depends`?

R/dummy_functions.R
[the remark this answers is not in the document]
gogonzo: @pawelru I don't want to export example data function just for a sake of one publicly available examples. I think it make sense to show also `cdisc_data()` call when calling `teal::init`.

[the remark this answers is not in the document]
gogonzo: I've changed `dummy_*` to `example_*` and kept them as unexported

kartikeyakirar (approved):
Looks good!

PR #859, opened 2023-07-02
298 reset button

chlebowa:
Relates to [this issue](https://github.com/insightsengineering/teal.slice/issues/298).

Adds module for management of snapshots of application state.
A snapshot contains in formation on all filter states currently existing within the app, as well as their application to particular modules.
A snapshot can be added at any time by the user.
A snapshot can be restored at any time by the user.
A snapshot can be saved to file.
The initial state of the application is the first snapshot (not displayed in the list).

In addition, `teal_slices` handles global-module-specific specification differently. The deciding argument is `module_specific`. `mapping$global_filters` will decide which - if any - filters are active on start-up in global mode.


chlebowa:
### TESTING

Set `teal` and `teal.slice` to `298_reset_button@filter_panel_refactor@main`
Use the following app.


```
options(teal.log_level = "WARN", teal.show_js_log = TRUE)
# options("teal.bs_theme" = bslib::bs_theme(version = 5))
# options(shiny.trace = TRUE)
# todo: change groupCheckbox to include locked (not able to interact with)
#       change groupCheckbox to have some colors (instead of grey)
# todo: available filter should present information about selected values (for example tooltip)

library(shiny)
library(scda)
library(scda.2022)
library(teal.data)
library(teal.transform)
library(teal.modules.general)
pkgload::load_all("../teal")

funny_module <- function (label = "Filter states", datanames = "all") {
  module(
    label = label,
    filters = datanames,
    ui = function(id, ...) {
      ns <- NS(id)
      div(
        h2("The following filter calls are generated:"),
        verbatimTextOutput(ns("filter_states")),
        verbatimTextOutput(ns("filter_calls")),
        actionButton(ns("reset"), "reset_to_default")
      )
    },
    server = function(input, output, session, data, filter_panel_api) {
      checkmate::assert_class(data, "tdata")
      observeEvent(input$reset, set_filter_state(filter_panel_api, default_filters))
      output$filter_states <-  renderPrint({
        logger::log_trace("rendering text1")
        filter_panel_api %>% get_filter_state()
      })
      output$filter_calls <- renderText({
        logger::log_trace("rendering text2")
        attr(data, "code")()
      })
    }
  )
}

default_filters <- teal::teal_filters(
  filter_var("iris", "Sepal.Length"),
  filter_var("iris", "Sepal.Width"),
  filter_var("iris", "Species"),
  filter_var("mtcars", "mpg"),
  exclude_varnames = list(
    iris = c("Petal.Length", "Petal.Length")
  ),
  mapping = list(
    funny = c("iris Sepal.Length"),
    funny2 = c("iris Sepal.Width"),
    global_filters = "mtcars mpg"
  )
)
app <- init(
  data = teal_data(
    dataset("iris", iris),
    dataset("mtcars", mtcars)
  ),
  modules = modules(
    tm_data_table(
      "table",
      variables_selected = list(),
      dt_args = list()
    ),
    modules(
      label = "tab1",
      funny_module("funny", datanames = NULL),
      funny_module("funny2", datanames = "iris")
    )
  ),
  filter = default_filters
)


runApp(app, launch.browser = TRUE)

```

R/module_snapshot_manager.R
```diff
@@ -0,0 +1,257 @@
...
+        )
+      })
+    })
+
+  })
+}
+
+
+
+
```
kartikeyakirar, line 218:
```suggestion
```
Removing whitelines

inst/css/sidebar.css
```diff
@@ -22,3 +22,29 @@ a.disabled {
...
+  flex: 1 1 auto;
+  padding: 0em 1em;
+  width: min-content;
+}
+
+.snapshot_table_row {
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+}
```
kartikeyakirar, line 43:
```suggestion
}

```
adding Whiteline

inst/css/sidebar.css
```diff
@@ -22,3 +22,29 @@ a.disabled {
...
+.snapshot_table_row *:first-child {
+  flex: 1 1 80%;
+}
```
kartikeyakirar, line 46:
```suggestion
}

```
adding Whiteline

R/module_filter_manager.R
```diff
@@ -107,109 +117,85 @@ filter_manager_srv <- function(id, filtered_data_list, filter) {
...
-        )
-      }
-    )
+    # Create mapping fo filters to modules in matrix form (presented as data.frame).
+    mapping_matrix <- reactive({
+      mapping_ragged <- lapply(filtered_data_list, function(x) slices_field(x$get_filter_state(), "id"))
+      all_names <- slices_field(slices_global(), "id")
+      mapping_smooth <- lapply(mapping_ragged, is.element, el = all_names)
+      as.data.frame(mapping_smooth, row.names = all_names)
+    })
```
gogonzo, line 130: mapping_matrix doesn't handle module names properly. Spaces are turned into dots and thus restore process doesn't work

R/zzz.R
```diff
@@ -23,6 +23,11 @@
   )
 }
 
-# Use non-exported function from teal.slice.
+# Use non-exported function(s) from teal.slice.
 # This is a temporary measure and will be removed two release cycles from now (now meaning 0.13.0).
 as.teal_slices <- getFromNamespace("as.teal_slices", "teal.slice") # nolint
+# This one is here because there are numerous instances of getting all ids from teal_slices.
+# The alternative is to always call vapply(<X>, `[[`, character(1), "id").
+slices_field <- getFromNamespace("slices_field", "teal.slice") # nolint
```
gogonzo, line 31: I'm against creating and using syntactiv sugar for one liners based on base function. I would import this here
```suggestion
```

gogonzo (commented):
There is a gap in a design regarding restoring states for non-module-specific filter panel. Is there an idea how to enable this?

R/module_filter_manager.R
chlebowa: This should be solved by adding `check.names = FALSE`.

R/zzz.R
chlebowa: removed

chlebowa:
> There is a gap in a design regarding restoring states for non-module-specific filter panel. Is there an idea how to enable this?

Sorry for the oversight.

I introduced snapshot management to the `FilteredData` class as well. I am not happy about it as there is a substantial amount of code duplication but the conditions are somewhat different. I will try to refine the solution so that one module can be used in both cases but I do not see a way at the moment.

R/module_snapshot_manager.R
```diff
@@ -0,0 +1,241 @@
...
+        shinyjs::click(id = "teal-main_ui-filter_manager-show", asis = TRUE)
+      }
+    })
+
+    # Restore initial state.
+    observeEvent(input$snapshot_reset, {
+      s <- "Initial application state"
+      ### Begin restore procedure. ###
+      snapshot <- snapshot_history()[[s]]
+      snapshot_state <- reassemble_slices(snapshot)
```
gogonzo, line 146: reassemble_slices imported from `teal.slice` namespace fails as `teal_slices` has no `mapping` and `module_specific` argument. We need to create a copy in `teal` to use `teal::teal_slices`.

Now we have `store`, `restore`, `as.list`, `slices_to_list`, `as.list`, `reassemble`, `disassemble_slices`. We need to figure out consistent way of converting slices to list and oposite.

chlebowa: Indeed. I returned the functions to `teal`, there are now copies in both packages.

[naming issue here](https://github.com/insightsengineering/teal.slice/issues/382)

chlebowa:
Now that `filter_panel_refactor@main` has been merged, I invite all of you to review again.

R/module_snapshot_manager.R
```diff
@@ -0,0 +1,276 @@
...
+        output[[id_saveme]] <- downloadHandler(
+          filename = function() {
+            sprintf("teal_snapshot_%s_%s.json", s, Sys.Date())
+          },
+          content = function(file) {
+            snapshot <- snapshot_history()[[s]]
+            snapshot_state <- reassemble_slices(snapshot)
+            teal.slice::slices_store(tss = snapshot_state, file = file)
+          }
+        )
```
gogonzo, line 196: observers are reinstantiated for all snapshots each time when `snapshot_history()` changes. `observeEvent` and `downloadHandler` should be outside of `renderUI` and instantiated only once per snapshot.

tests/testthat/test-init.R
```diff
@@ -192,7 +200,11 @@ testthat::test_that("init filter fails when mapping don't refer to available fil
   testthat::expect_error(
     init(
       data = list(iris = iris),
+<<<<<<< HEAD
+      modules = teal:::get_dummy_modules(),
+=======
```
gogonzo, line 205: FYI there are still some conflicts

chlebowa: Oh my :scream:

R/module_snapshot_manager.R
chlebowa: They do not seem to be duplicated though. No matter how many times snapshot history is updated, a click on a save button only triggers one handler, a click on a select button restores the state once (no visible multiple updates to the filter panel). Are you sure this is a problem?

I have rewritten it (not pushed yet) so that observers, handlers and divs are each stored in separate `reactiveValues` and one observer on `snapshot_history` only appends elements that are not already there but does not recreate the whole list. This way observers and handlers are instantiated only once. Would that be better?

gogonzo: Yes, click will trigger one handler. And yes, they are not duplicated, duplicated is operation of setting these observers.

> Are you sure this is a problem?

Doesn't seem right to create new observer in a session and overwrite the previous one. I'm pretty sure there some side effects in a `session` object because of this (relinking perhaps).

> Would that be better?

Hard to imagine final look. 

-----

What if div stays in `renderUI` and observers are moved to `observeEvent(added_snapshots(), ...`, similarly to this, see lines 363-394:

https://github.com/insightsengineering/teal.slice/blob/3f4c6ff49eb9f7816fed7849af53bfd431e85484/R/FilterStates.R#L375

chlebowa: I thought about smth similar but I did not want to juggle multiple `reactiveVal`s as I have 2 or 3 lists to maintain here. Let me push what I described above (the current version will be commented out) so you can have a look. It is not much different.

gogonzo: I had my shiny lesson now. Invalidation between elements happens from one-to-another and session is not involved in the process. I was wrong in the first comment as adding new observer doesn't update whole reactivity graph - it is smart enough to pass invalidation locally between two elements (reactive -> endpoint)

gogonzo: @chlebowa you can back to the state before my first comment. This scenario differs from `srv_active` as we only add snapshots. so we don't need to remove observers at any point.

P.S. Update "do not reinstantiate observers" hasn't changed anything from this perspective as there is `observeEvent(snapshot_history(), ... observeEvent() ...)`

chlebowa: After some talking we decided to create UI elements and server logic for individual snapshots outside of `renderUI`. They will be stored in `reactiveValues` and new ones will be appended only if they do not yet exist.

R/module_filter_manager.R
```diff
@@ -111,109 +127,91 @@ filter_manager_srv <- function(id, filtered_data_list, filter) {
     }
     filtered_data_list <- flatten_nested(filtered_data_list)
 
-    # global list of slices (all available teal_slice)
-    slices_global <- reactiveVal(filter)
+    # # flatten nested alternative
+    # filtered_data_list <- unlist(filtered_data_list)
+    # names(filtered_data_list) <- sub("(.+)\\.(.+$)", "\\2", names(filtered_data_list))
```
chlebowa, line 132: @gogonzo 
May I bother you for an opinion on this?

gogonzo: Comment to remove. I prefer your solution with the recursive call

chlebowa:
Some unit tests could no longer be run so I had to remove them. I will gladly accept suggestions on using `testServer` for multi-modular setups.

R/module_snapshot_manager.R
```diff
@@ -0,0 +1,301 @@
...
+        updateTextInput(inputId = "snapshot_name", value = , placeholder = "Meaningful, unique name")
+      } else {
+        snapshot <- disassemble_slices(slices_global())
+        attr(snapshot, "mapping") <- matrix_to_mapping(mapping_matrix())
+        snapshot_update <- c(snapshot_history(), list(snapshot))
+        names(snapshot_update)[length(snapshot_update)] <- snapshot_name
+        snapshot_history(snapshot_update)
+        removeModal()
+        # Reopen filter manager modal by clicking button in the main application.
+        shinyjs::click(id = "teal-main_ui-filter_manager-show", asis = TRUE)
```
gogonzo, line 143: not really fan of this line, but I think it's okey at least here.

gogonzo (approved):
I consider technicals on the teal side robust. 

image

m7pr:
I love the elegancy of the graph/workflow

chlebowa:
Great chart indeed @gogonzo and you beat me to it :slightly_smiling_face: 

I don't quite like the click either but since we can't think of anything better at the moment, let's put a pin in it, maybe we will come up with something later. Unless @lcd2yyz withdraws the requirement.

m7pr:
:fire:

PR #868, opened 2023-07-21
382 rename internal functions

chlebowa:
Related to [this issue](https://github.com/insightsengineering/teal.slice/issues/382)

`disassemble_slices` function is removed, will use `as.list.teal_slices` from `teal.slice`
`reassemble_slices` function is remove, will use `as.teal_slices` from this package
`as.teal_slices` function lifted from `teal.slice` is renamed to `list_to_teal_slices`


gogonzo (approved):
wundabar

PR #877, opened 2023-07-26
769 replace filters argument

### LINKED ISSUE #769: Replace `filters` argument in module and replace with `datanames`
Currently each module (teal_module class) has a field  [`filters`](https://github.com/insightsengineering/teal/blob/79678e78267098d1a0377f0b6b9cd8bba1cea988/R/modules.R#L289) which was used for a long time to indicate what datasets are used by the module so that filter-panel displays only module-relevant filter states.

Acceptance criteria:
- deprecate `filters` argument in `module` and recommend to use `datanames` instead.
- ~resolve `module$filter` in `teal::init` to include parent `datanames` also~ [done](https://github.com/insightsengineering/teal/blob/cf43277be1154ff0f3ef5e397118ab6051986e17/R/init.R#L148) 
- ~`resolve `module$filter` from `"all"` to all possible dataname`~ [done](https://github.com/insightsengineering/teal/blob/cf43277be1154ff0f3ef5e397118ab6051986e17/R/init.R#L148)
- `module$datanames` determines which datasets should be presented in filter-panel [see this](https://github.com/insightsengineering/teal/blob/cf43277be1154ff0f3ef5e397118ab6051986e17/R/module_tabs_with_filters.R#LL174C69-L174C85)  [see this](https://github.com/insightsengineering/teal.slice/blob/28ff2b274ba23ebb933ac4c98eea6df294d89c6d/R/FilteredData.R#L638)
- `module$datanames` determines which datasets are passed to the module [see this](https://github.com/insightsengineering/teal/blob/cf43277be1154ff0f3ef5e397118ab6051986e17/R/module_nested_tabs.R#L244)
- `module$datanames` should be `NULL` when `module$server` has no `data` argument. [see docs of the filters argument](https://github.com/insightsengineering/teal/blob/cf43277be1154ff0f3ef5e397118ab6051986e17/R/modules.R#L167)
- `module$datanames` when `NULL` filter-panel should be hidden. [logic here](https://github.com/insightsengineering/teal/blob/cf43277be1154ff0f3ef5e397118ab6051986e17/R/module_tabs_with_filters.R#L179)
- All apps in [teal.gallery](https://github.com/insightsengineering/teal.gallery) shouldn't have any failures

### LINKED ISSUE #821: Fix validation message
> Good findings.
> 
> Also, we should improve the validation message given out here:
> 
> image
> 
> The current message implies that all columns do not exist where it supposed to be just for `TRT` column.

_Originally posted by @donyunardi in https://github.com/insightsengineering/teal/pull/819#discussion_r1157792052_

### Acceptance Criteria
* Validation message should only show the missing columns, not all columns selected in the shiny input.            
* Using the example above, the validation should say `ADSL does not have the required variables: TRT`

chlebowa:
Fixes #769 

Argument is renamed as per request.
Downstream effects are resolved.

I also identified instances of partial name matching when indexing lists and removed them. There might be more. This is an argument against using `list[["name"]]` over `list$name`. The occurrence of `modules$filter` was an unwelcome surprise.

Also fixes #821


R/module_nested_tabs.R
```diff
@@ -330,11 +327,8 @@ srv_nested_tabs.teal_module <- function(id, datasets, modules, is_module_specifi
...
-    datanames,
-    simplify = FALSE,
-    function(x) {
-      rlang::hash(datasets$get_data(x, filtered = FALSE))
-    }
-  )
+  if (is.null(names(datanames))) {
+    datanames <- structure(datanames, names = datanames)
+  }
+  lapply(datanames, function(x) rlang::hash(datasets$get_data(x, filtered = FALSE)))
```
gogonzo, line 333:
```suggestion
  lapply(datanames, function(x) rlang::hash(datasets$get_data(x, filtered = FALSE)))
```

Dropping names from the list will cause dropping hashes from the code. See here
https://github.com/insightsengineering/teal/blob/720a374771b296460f0eee9dabf00d4891f7c831/R/get_rcode_utils.R#L73

R/module_nested_tabs.R
```diff
@@ -172,18 +172,16 @@ srv_nested_tabs.teal_modules <- function(id, datasets, modules, is_module_specif
...
+    modules_reactive <- lapply(names(modules$children), function(module_id) {
+      srv_nested_tabs(
+        id = module_id,
+        datasets = datasets[[labels[module_id]]],
+        modules = modules$children[[module_id]],
+        is_module_specific = is_module_specific,
+        reporter = reporter
+      )
+    })
+    names(modules_reactive) <- names(modules$children)
```
gogonzo, line 184: What is wrong with `sapply` in this case? `sapply` adds names for free. In my opinion the only controversy with `sapply` could be `simplify = TRUE` (default), but we can just use `simplify = FALSE`.

R/module_nested_tabs.R
```diff
@@ -292,14 +291,12 @@ srv_nested_tabs.teal_module <- function(id, datasets, modules, is_module_specifi
...
+  datanames <- if (is.null(module$datanames)) datasets$datanames() else module$datanames
+  datanames <- structure(datanames, names = datanames)
 
   # list of reactive filtered data
-  data <- sapply(
-    USE.NAMES = TRUE,
-    X = datanames,
-    FUN = function(x) eventReactive(trigger_data(), datasets$get_data(x, filtered = TRUE))
-  )
+  data <- lapply(datanames, function(x) eventReactive(trigger_data(), datasets$get_data(x, filtered = TRUE)))
```
gogonzo, line 299: `data` needs to be a named list.
```suggestion
  data <- sapply(datanames, simplify = FALSE, function(x) eventReactive(trigger_data(), datasets$get_data(x, filtered = TRUE)))
```

R/init.R
```diff
@@ -141,13 +141,13 @@ init <- function(data,
   join_keys <- data$get_join_keys()
   resolve_modules_datanames <- function(modules) {
     if (inherits(modules, "teal_modules")) {
-      modules$children <- sapply(modules$children, resolve_modules_datanames, simplify = FALSE)
+      modules$children <- lapply(modules$children, resolve_modules_datanames)
```
gogonzo, line 144: Names are dropped here, is this okey? Elements of `teal_modules` should be named

chlebowa:
Common answer to `sapply` comments:

> What is wrong with `sapply` in this case? `sapply` adds names for free. In my opinion the only controversy with `sapply` could be `simplify = TRUE` (default), but we can just use `simplify = FALSE`.

> Names are dropped here, is this okey? Elements of `teal_modules` should be named

I change `sapply` to `lapply` as a force of habit, mostly because of simplification results being not entirely secure. It is also frowned upon in some circles, e.g. on Bioconductor.

As you mentioned, `lapply` can reuse the names of `X`, even though it cannot **build** names from the values of `X`. That is why I added names to the character vector that goes into `lapply` whenever possible and where not possible, I added names afterwards. Indeed, the names are even verified by unit tests.

Perhaps I did go a little overboard. If we agree that in all these cases the result should be a list (which is apparently certain but one can never bee too careful), I will restore `sapply` with `simplify = FALSE`.

We could use `Map` -  it calls `mapply` under the hood so names are created like with `sapply` but it always goes `simplify = FALSE`. I say could but I do not necessarily think we should as it is not a commonly used function. We do throw `Filter` around a lot, though.

gogonzo:
@chlebowa I think  `sapply(X, simplify = FALSE)` is entirely predictable and useful - we can use single call instead of two when we want to `USE.NAMES`

chlebowa:
blocked pending [fixing python app](https://github.com/insightsengineering/teal.gallery/issues/39)

chlebowa:
unblocked

gogonzo (approved):
👍

PR #879, opened 2023-07-30
867 modify mapping matrix

### LINKED ISSUE #867: modify `mapping_matrix` in filter manager module
Blocked by insightsengineering/teal.slice#298 

The `mapping_matrix` displays the mapping of filters to modules as a logical matrix (actually a `data.frame`).
image
It only says whether a filter is applied to a module. However, some filters are not available for all modules. 

In case a filter CANNOT be applied to a module, the mapping matrix should display `NA`.

Bear in mind the mapping matrix is passed to the snapshot manager module, where it is converted to the `mapping` attribute of `teal_slices` by the `mapping_to_matrix` function (or something similar) and that process does not foresee NAs.

### LINKED ISSUE #878: render mapping matrix
The filter manager module displays the mapping matrix in `output$slices_table` with `renderTable`. By default all (character) columns are adjusted to the left. This is not ideal - column contents should be centered but row names (state ids) should be left-adjusted.

bad:
image
good:
image

The solution would be to do `align = c("l", rep("c", ncol(mapping_matrix)))` but that would require accessing the mapping matrix outside a reactive context and it is a reactive expression. The mapping matrix changes dimensions during the app session and thus cannot be `isolate`d here.

An alternative would be to put the whole `output$slices_table <- renderTalble({})` call in `observe` but that solution is best avoided.

Build the HTML for the table from scratch and set the desired alignment using CSS only, if possible.

chlebowa:
Fixes #867 
Fixes #878 

The mapping matrix is modified to display filters that cannot be applied to a particular module as NA.

In `matrix_to_mapping` filters that are NA are treated as TRUE because global filters may not apply to some modules.

Requires a new public method in `FilteredData`, introduced [here](https://github.com/insightsengineering/teal.slice/pull/408).

### DIFF
--- a/R/module_filter_manager.R
+++ b/R/module_filter_manager.R
@@ -142,7 +142,7 @@ filter_manager_srv <- function(id, filtered_data_list, filter) {
       state_ids_global <- vapply(slices_global(), `[[`, character(1L), "id")
       mapping_smooth <- lapply(filtered_data_list, function(x) {
         state_ids_local <- vapply(x$get_filter_state(), `[[`, character(1L), "id")
-        state_ids_allowed <- vapply(x$get_available_teal_slices(), `[[`, character(1L), "id")
+        state_ids_allowed <- vapply(x$get_available_teal_slices()(), `[[`, character(1L), "id")
         states_active <- state_ids_global %in% state_ids_local
         ifelse(state_ids_global %in% state_ids_allowed, states_active, NA)
       })
--- a/R/module_filter_manager.R
+++ b/R/module_filter_manager.R
@@ -204,10 +204,7 @@ filter_manager_srv <- function(id, filtered_data_list, filter) {
 filter_manager_module_srv <- function(id, module_fd, slices_global) {
   moduleServer(id, function(input, output, session) {
     # Only operate on slices that refer to data sets present in this module.
-    available_slices <- reactive({
-      Filter(function(slice) slice$dataname %in% module_fd$datanames(), slices_global())
-    })
-    module_fd$set_available_teal_slices(available_slices)
+    module_fd$set_available_teal_slices(reactive(slices_global()))
 
     # Track filter state of this module.
     slices_module <- reactive(module_fd$get_filter_state())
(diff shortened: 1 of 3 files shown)


R/module_filter_manager.R
```diff
@@ -137,22 +137,29 @@ filter_manager_srv <- function(id, filtered_data_list, filter) {
...
     mapping_matrix <- reactive({
-      module_states <- lapply(filtered_data_list, function(x) x$get_filter_state())
-      mapping_ragged <- lapply(module_states, function(x) vapply(x, `[[`, character(1L), "id"))
-      all_names <- vapply(slices_global(), `[[`, character(1L), "id")
-      mapping_smooth <- lapply(mapping_ragged, is.element, el = all_names)
-      as.data.frame(mapping_smooth, row.names = all_names, check.names = FALSE)
+      state_ids_global <- vapply(slices_global(), `[[`, character(1L), "id")
+      mapping_smooth <- lapply(filtered_data_list, function(x) {
+        state_ids_local <- vapply(x$get_filter_state(), `[[`, character(1L), "id")
+        state_ids_allowed <- vapply(x$get_available_teal_slices()(), `[[`, character(1L), "id")
```
gogonzo, line 145: I think we can do slight improvement in the related code. Line 208 in this file is not needed as `FilteredData$set_available_teal_slices` subsets by available datasets already.

https://github.com/insightsengineering/teal/blob/2cb0011d7f47dce7abc648e1840a4b561b6c07da/R/module_filter_manager.R#L208

https://github.com/insightsengineering/teal.slice/blob/57d873c0b06893e556f221a9829820395d09ca6e/R/FilteredData.R#L153

gogonzo, line 145: I'm just curious if this should be isolated (red one one a graph)? `FilteredData$available_teal_slices` is triggered by `slices_global`, so it seams to be redundant. I'm also worried that this reactivity graph might get out of control if we introduce new reactives. I think that making `eventReactive` might be a better solution to limit people in the future to trigger this by other reactives. Same here https://github.com/insightsengineering/teal.slice/pull/408#discussion_r1279336853

What do you think?


```suggestion
        state_ids_allowed <- vapply(isolate(x$get_available_teal_slices()()), `[[`, character(1L), "id")
```

local-filter-panel-Page-13

gogonzo: Sorry, above is not true. Mapping matrix is displayed only when the filter manager is seen, so it doesn\t really matter.

gogonzo (approved):

PR #901, opened 2023-08-17
don't throw validation messages unnecessarily

pawelru:
During rendering TLG-Catalog I noticed that each shiny app throws a message about `data` argument in the reporter previewer module. This fix aim to silent unnecessary messages

https://github.com/insightsengineering/tlg-catalog/actions/runs/5888144351/job/15968763355

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # teal 0.14.0.9001
 
+### Miscellaneous
+
+* Enhance a `module` validation checks so that it won't throw messages about `data` argument unnecessarily.
+
 # teal 0.14.0
 
 ### New features
--- a/R/modules.R
+++ b/R/modules.R
@@ -232,7 +232,7 @@ module <- function(label = "module",
     )
   }
 
-  if (!is.element("data", server_formals)) {
+  if (!is.element("data", server_formals) && !is.null(datanames)) {
     message(sprintf("module \"%s\" server function takes no data so \"datanames\" will be ignored", label))
     datanames <- NULL
   }


chlebowa:
I appreciate removing the message for the reporter module but will other modules not be affected as well?

pawelru:
Other modules will be affected and that's good. Reporter is just one great example.
I have made this analysis and it seems that `datasets` argument [is required (i.e. non-missing) but nullable](https://github.com/insightsengineering/teal/blob/bbf13f5e435cc945ca0680344ebf71233ba90a05/R/modules.R#L201). If your module does not consume `data` (e.g. reporter module), setting up `datasets` as NULL to should be reasonable but this leads to the aforementioned warning. We have to have a path to enable warning-free processing of non-data modules.

pawelru:
I have added one more commit that changes a default value for datanames with that respect.
For a very very simple modules that does not define server part of the module we got the warning.
```
app <- init(
  data = list(test = data.frame()),
  modules = list(
    module(
      ui = function(id) {
        tags$p("Hello World!")
      }
    )
  )
)
```

chlebowa:
Having given it some thought, I agree with extending the condition on displaying the warning but I do not particularly like the `if` in the argument default. The help for the argument is quite explicit, plus with the recent changes (new filter specification, module-specific filters) filter panel options are more hands-on and requiring the app dev to change `"all"` to `NULL` by themselves is in line with that.

pawelru:
> and requiring the app dev to change "all" to NULL by themselves is in line with that.

I would (not so strongly) disagree with that. This is a little bit of code smell for me of interdependent arguments. Obviously, we won't completely remove the interdependency here but my intention is to enhance the default as I believe that (even though this is an exposed parameter) in most of the cases users / module developers don't specify this argument and just stick to the default value which should be either `"all"` or `NULL`.

chlebowa:
I understand. Is an enhanced default necessary, though? Seeing as in most cases developers set `datanames = "all"`, why not have that as default and if one is bothered by the warning, one can set `datanames = NULL` themselves.

pawelru:
> I understand. Is an enhanced default necessary, though?

The background is that I am trying to sell `teal::module` as a standard shiny module packed into a single object. I wanted to state that module consists of ui and server elements (at the minimum) with few optional extras as well as few extra methods. Reverting it back would make this statement _less true_ but this is not really necessary. I think I am also ok with reverting this back and process this as a standard issue with a proper discussion among developers.

pawelru:
ok I reverted back my recent change about default arg value

chlebowa:
I'm in two minds, honestly. I was hoping someone else would chime in. If you're ok with the app dev having to type a little more, I'll approve this as it is now.

chlebowa (approved):

PR #907, opened 2023-08-24
898 save app state

### LINKED ISSUE #898: [Research] Ability to save and restore the state of teal app
Related to https://github.com/insightsengineering/NEST-roadmap/issues/82

### Summary
This research assignment aims to explore the feasibility of incorporating a feature enabling users to save and restore the complete state of a teal app, encompassing both filters and encoding information.

### Acceptance Criteria
* Evaluation or potential Proof of Concept (PoC) for the research.

chlebowa:
Closes #898 

Prototype for saving and restoring app state, i.e. selection state of all inputs in app. When complete, can be easily combined with saving filter state.

#### The problem 

In essence, everything comes down to two actions: 1) find state of all inputs in the app, and 2) set state of all inputs in the app. Once we have the state in hand, saving and retrieving it is trivial.

Ad 1) One simply has to go through the `input` object. Modules slightly complicate things as things are namespaced and separated but that is solved by finding the master session object (the one for the app, not for individual modules) and working from there.
Ad 2) This is not hard in principle as one can set the state of any input with `session$sendInputMessage`, using the master session object and namespaced input ids. However...

#### The issue

Inputs in `data_extract_spec`s are hierarchical, meaning given a dataset, one first selects a column and then a value from that column. The value input is created in response to selecting a column value. Now, when one wants to restore input states, one will trigger an observer and within that observer `session$sendInputMessage` will be called. This will queue messages that will update all inputs as required but those messages will only be applied after the observer closes. Then, when the higher order input is updated, the lower order input will be reset to default.

See the app for a simplified example.

So, the nature of the reactive cycle kills the reset process. I tried forcing flushes, I tried calling the update function recursively, at the moment I cannot solve this.

#### Note

Saving and loading may still have some bugs (probably does), I don't want to get into minutiae before the major issues are solved.

#### Simple testing app
```
library(shiny)
rm(list = ls())
devtools::load_all("../teal")

nameUI <- function(id) {
  ns <- NS(id)
  tagList(
    actionButton(ns("inside"), "inside"), # for browser in module
    uiOutput(ns("inputs1")),
    uiOutput(ns("inputs2")),
    uiOutput(ns("inputs3")),
    plotOutput(ns("plot")),
    verbatimTextOutput(ns("value")),      # past and current values of value
    NULL
  )
}

nameServer <- function(id) {

    ns <- session$ns

    observeEvent(input$inside, {
      browser()
    })

    value_history <- reactiveVal()
    observeEvent(input[["value"]], {
      add <- if (is.null(input[["value"]])) "NULL" else input[["value"]]
      update <- c(value_history(), add)
      value_history(update)
    }, ignoreNULL = FALSE)
    output[["value"]] <- renderPrint({
      value_history()
    })


    output[["inputs1"]] <- renderUI({
      choices <- names(Filter(is.factor, warpbreaks))
      selectInput(ns("column"), "column", choices = choices, selected = choices[1])
    })
    output[["inputs2"]] <- renderUI({
      col <- req(input[["column"]])
      choices <- levels(warpbreaks[[col]])
      selectInput(ns("value"), "value", choices = c("please select" = "", choices), selected = NULL)
    })
    output[["inputs3"]] <- renderUI({
      choices <- letters[1:4]
      radioButtons(ns("letter"), "letter", choices = choices, selected = choices[1], inline = TRUE)
    })


    output[["plot"]] <- renderPlot({
      col <- req(input[["column"]])
      val <- req(input[["value"]])
      data <- warpbreaks[warpbreaks[[col]] == val, "breaks"]
      tryCatch(
        hist(data, breaks = 10, las = 1),
        warning = function(w) req(FALSE)
      )
    })

  })
}

ui <- fluidPage(
  actionButton("store", "store"),        # capture inputs
  actionButton("restore", "restore"),    # set inputs
  nameUI("module1"),                     # module
  verbatimTextOutput("grab"),            # current grab
  NULL
)

  nameServer("module1")

  grab_history <- reactiveVal({
    list(
      "Initial input state" = isolate(app_state_grab())
    )
  })

  observeEvent(input$store, {
    grab_update <- c(grab_history(), list(app_state_grab()))
    grab_history(grab_update)
  })
  observeEvent(input$restore, {
    grab <- rev(grab_history())[[1L]]
    app_state_restore(grab)
  })

  output[["grab"]] <- renderPrint({
    rev(grab_history())[[1L]]
  })

}

shinyApp(ui, server)
```


m7pr:
Hey, well done with `shinyjs::click()` idea. I was able to run the app and the described behavior is provided. I skimmed through the code and it's clean and readable. Really solid, as you always keep up to high standards. I think you are good to go exploring how would you store filters, even though we know limitations of the `click` bottlenecks.

chlebowa:
Thanks @m7pr :slightly_smiling_face: 

I have done some more research into how one can interfere with the reactive cycle. I wanted to manually manipulate invalidation flags but I didn't find a way. I tried dynamically injecting JS that would ignore some events on some elements, the assumption was that I could kill part of the reactivity for the duration of the update, but in the end it has to be brought back up in another flush cycle. I give up. The cyclical clicks are the bes I can do.

To mitigate the side effect I tried to add a loader that will obscure the plot until all updates are complete but it keeps popping up anyway.

I will spend a little more time to make sure saving and loading works properly and then open the PR for review.

gogonzo (commented):
Looks like we are trying redo what shiny offers with [bookmarking](https://shiny.posit.co/r/articles/share/bookmarking-state/). There is a nice option to save states on the server and reuse them with url like `https://gallery.shinyapps.io/bookmark-saved/?_state_id_=<STATE ID>`. Maybe we can store there selected inputs plus `reactiveValues` (eg. slices_global).

In the "ancient" teal we had bookmarking which handled filter-states

https://github.com/insightsengineering/teal/blob/8240395fdbb568f11645a231a7fd34bcda31b0da/R/module_teal.R#L190

Maybe we can address it in a similar way?

PR #916, opened 2023-09-08
898 save app state version 2

### LINKED ISSUE #898: [Research] Ability to save and restore the state of teal app

chlebowa:
Closes #898.
Alternative to #907.

### OVERVIEW

Inputs are stored with `shiny`'s built-in bookmarking mechanism using the "server" way.

[literature](https://shiny.posit.co/r/articles/share/bookmarking-state/)

Using bookmarks requires that the `shiny` option `"bookmarkStore"` be set to `"server"`. Normally this is achieved by calling `enableBookmarking("server")` before running the app or by setting `enableBookmarking = "server"` in a `shinyApp` call. 
Here the option is set with a loading hook in `teal`.

Bookmarks are added to a list much like snapshots but elements of the list are URLs pointing to bookmarked applications that open in new windows.

The bookmark URL is obtained by the `grab_state` function in an admittedly shady way, using internal classes and function from `shiny`.

#### ISSUES

Restoring inputs occurs (as far as we know) as the `shiny` session starts, before the app is actually ready.

##### data-dependent inputs

`teal` modules often utilize a pattern for creating data-dependent selections:
```
# in ui funciton
  selectInput(ns("myInput"), "label", choices = NULL)
# and then in server
  updateSelectInput(inputId = "myInput", choices = data_derived_choices)
```
When the app starts and the module is not running yet, a value cannot be restored because it must be a subset of `choices`.

##### hierarchical inputs
When the choices or values of a junior input depend on the value of the senior one, restoring the senior can cause the restored value of the junior to be overwritten by invalidation following the change of value of the senior. Note that the senior can be defined ui-side or server-side and restoration results will vary.

##### SOLUTION: use `renderUI` rather than `update*Input`.

Replacing the aforementioned pattern for data-dependent inputs with `renderUI` does allow to restore input state in many cases. It certainly works in `example_module`.

Here is an app that highlights different restoration success with `update*` and `renderUI`. Note that neither is perfect.
```
devtools::load_all("../teal")

basic_module_ui <- function(id) {
  ns <- NS(id)
  tagList(
    # by update
    selectInput(ns("parent_srv"), "parent_srv", choices = NULL),
    selectInput(ns("child_choices_srv"), "child_choices_srv", choices = NULL),
    numericInput(ns("child_value_srv"), "child_value_srv", value = NULL),
    selectInput(ns("parent_ui"), "parent_ui", choices = LETTERS[26:23]),
    numericInput(ns("child_value_ui"), "child_value_ui", NULL),
    
    # # by render
    # uiOutput(ns("select_parent_srv")),
    # uiOutput(ns("select_child_choices_srv")),
    # uiOutput(ns("select_child_value_srv")),
    # selectInput(ns("parent_ui"), "parent_ui", choices = LETTERS[26:23]),
    # uiOutput(ns("select_child_value_ui")),
    
    verbatimTextOutput(ns("final")),
    state_manager_ui(ns("state_manager")),
    NULL
  )
}

basic_module_server <- function(id) {
    
    ns <- session$ns
    
    # by update
    updateSelectInput(inputId = "parent_srv", choices = c("", "lower", "upper"))
    observeEvent(input$parent_srv, {
      ch <- switch(input$parent_srv, "lower" = letters, "upper" = LETTERS, "")
      updateSelectInput(inputId = "child_choices_srv", choices = na.omit(ch[1:4]))
      updateNumericInput(inputId = "child_value_srv", value = match(input$parent_srv, c("lower", "upper"), nomatch = 0))
    })
    observeEvent(input$parent_ui, {
      updateNumericInput(inputId = "child_value_ui", value = match(input$parent_ui, LETTERS[26:23], nomatch = 0))
    })
    
    # # by render
    # output$select_parent_srv <- renderUI({
    #   selectInput(ns("parent_srv"), "parent_srv", choices = c("", "lower", "upper"))
    # })
    # output$select_child_choices_srv <- renderUI({
    #   ch <- switch(req(input$parent_srv), "lower" = letters, "upper" = LETTERS, "")
    #   selectInput(ns("child_choices_srv"), "child_choices_srv", choices = na.omit(ch[1:4]))
    # })
    # output$select_child_value_srv <- renderUI({
    #   numericInput(ns("child_value_srv"), "child_value_srv", value = match(input$parent_srv, c("lower", "upper"), nomatch = 0))
    # })
    # output$select_child_value_ui <- renderUI({
    #   numericInput(ns("child_value_ui"), "child_value_ui", value = match(input$parent_ui, LETTERS[26:23], nomatch = 0))
    # })
    
    output$final <- renderPrint({
      list(
        parent_srv = input$parent_srv, child_choices_srv = input$child_choices_srv, child_value_srv = input$child_value_srv, 
        parent_ui = input$parent_ui, child_value_ui = input$child_value_ui
      )
    })
    
    state_manager_srv("state_manager")
  })
}

ui <- function(request) fluidPage(basic_module_ui("basic_module"))
server <- function(input, output, session) basic_module_server("basic_module") 
shinyApp(ui, server)
```

##### data extracts

Inputs created by data extract specs - which can be hierarchical - are not restored. This is probably because setting values by `teal.transform` mechanisms occurs _after_ the bookmark has been restored.

#### SUMMARY

The last issue is one that I and @gogonzo have not been able to overcome. We have delved into the (internal) classes that `shiny` uses to restore state in the hopes that if we triggered the process manually from `teal`, we would be able to bypass overwriting values by data extracts but with no success.

Perhaps removing the infamous `insertUI` from `teal` would remove some of the obstacles but we cannot predict that with any degree of certainty.

It seems with the current state of `teal` the only way to properly restore full app state is the solution presented in #907, as flawed as it is.


chlebowa:
As a last-ditch effort I tried to modify the `session$onBookmarked` to include `session$sendInputMessage` that would "manually" set input state. This was not only redundant but also ran into the same problem as (solution 1) because updates are applied once only when the app flushes and then junior inputs are invalidated and recalculated based on the (now) current state of their seniors.

gogonzo (commented):
I don't see anything in this solution that isn't in [solution 3](https://github.com/insightsengineering/teal/pull/1011). If you agree @chlebowa we can drop this PR from the discussion

chlebowa:

PR #918, opened 2023-09-12
readme updates

m7pr:
Related to https://github.com/insightsengineering/coredev-tasks/issues/462


.Rbuildignore
```diff
@@ -1,3 +1,4 @@
+LICENSE
```
averissimo, line 1: Shouldn't the LICENSE file be removed?

averissimo (approved):
LGTM!

PR #919, opened 2023-09-15
910 Remove Report Previewer column

### LINKED ISSUE #910: Remove Report Previewer column from the Snapshot Manager
I don't think we need to include the Report Previewer module in the snapshot manager.

Screenshot 2023-08-29 at 3 30 57 PM

chlebowa:
Fixes #910 

The column named `"Report previewer"` is removed from the mapping matrix display.
Since the data.frame sent to `renderTable` potentially has one less column, the `align` argument is modified accordingly.

Additionally, `"Report previewer"` has been reserved as a module label during module creation, much like `"global_filters"`.
Since the actual previewer is supposed to have that label, it is set after the module is created.

This will only work as long as the name `"Report previewer"` does not change.

### DIFF
--- a/R/module_filter_manager.R
+++ b/R/module_filter_manager.R
@@ -165,9 +165,9 @@ filter_manager_srv <- function(id, filtered_data_list, filter) {
         }
 
         # Report Previewer will not be displayed.
-        mm[!grepl("Report previewer", names(mm))]
+        mm[names(mm) != "Report previewer"]
       },
-      align = paste(c("l", rep("c", sum(!grepl("Report previewer", names(filtered_data_list))))), collapse = ""),
+      align = paste(c("l", rep("c", sum(names(filtered_data_list) != "Report previewer"))), collapse = ""),
       rownames = TRUE
     )
 
--- a/R/module_filter_manager.R
+++ b/R/module_filter_manager.R
@@ -147,7 +147,6 @@ filter_manager_srv <- function(id, filtered_data_list, filter) {
         ifelse(state_ids_global %in% state_ids_allowed, states_active, NA)
       })
 
-      # mapping_smooth <- mapping_smooth[grep("Report previewer", names(mapping_smooth), invert = TRUE)]
       as.data.frame(mapping_smooth, row.names = state_ids_global, check.names = FALSE)
     })
 
--- a/R/module_filter_manager.R
+++ b/R/module_filter_manager.R
@@ -147,6 +147,7 @@ filter_manager_srv <- function(id, filtered_data_list, filter) {
         ifelse(state_ids_global %in% state_ids_allowed, states_active, NA)
       })
 
+      # mapping_smooth <- mapping_smooth[grep("Report previewer", names(mapping_smooth), invert = TRUE)]
       as.data.frame(mapping_smooth, row.names = state_ids_global, check.names = FALSE)
     })
 
@@ -164,9 +165,10 @@ filter_manager_srv <- function(id, filtered_data_list, filter) {
           rownames(mm) <- ""
         }
 
-        mm
+        # Report Previewer will not be displayed.
+        mm[!grepl("Report previewer", names(mm))]
       },
-      align = paste(c("l", rep("c", length(filtered_data_list))), collapse = ""),
+      align = paste(c("l", rep("c", sum(!grepl("Report previewer", names(filtered_data_list))))), collapse = ""),
       rownames = TRUE
     )
 
--- a/R/modules.R
+++ b/R/modules.R
@@ -212,7 +212,18 @@ module <- function(label = "module",
   }
 
   if (label == "global_filters") {
-    stop("Label 'global_filters' is reserved in teal. Please change to something else.")
+    stop(
+      sprintf("module(label = \"%s\", ...\n  ", label),
+      "Label 'global_filters' is reserved in teal. Please change to something else.",
+      call. = FALSE
+    )
+  }
+  if (label == "Report previewer") {
+    stop(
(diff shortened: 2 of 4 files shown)


chlebowa:
Tricky question. If an app dev decides to name a module `"Report previewer"`, the actual previewer will be the last one, so a check like that would remove the previewer but not the module. However, if that happens in an app that doesn't use the reporter and that module is placed at the end, it will still be removed.

I guess the question is how likely is it that an app dev names a module `"Report previewer"`? Also, didn't that use to be forbidden?

averissimo:
Tricky indeed. Then it would have to check if there is there is any `modulePreviewer` beforehand.

That looks like a dark and long rabbit hole.

As far as I can see, it's not easy to detect if the `reporter_previewer_module` was appened from the `filter_manager_modal_srv` context _(without passing that information along)_

> Also, didn't that use to be forbidden?

I tried to name it and it seemed to work (and hide on filter manager)

image

image

averissimo:
However, this seems so farfetched, that it may be better to **accept this scenario** or:

* prevent that name from being used as module title
* output a warning when it's used

chlebowa:
> > Also, didn't that use to be forbidden?
> 
> I tried to name it and it seemed to work (and hide on filter manager)

It does but I have a feeling it shouldn't and it didn't use to.

@donyunardi @lcd2yyz do you think we can forbid that?

donyunardi:
This line will always force "Report previewer" to be the last module, after all other modules.
https://github.com/insightsengineering/teal/blob/139971a42add5084c69928d7ca1271773f3c4e53/R/module_teal.R#L238-L240

An alternative way is to combine the ideas that you both presented: check if the last element of the `filtered_data_list` is equal to `Report previewer`. 

`reporter_previewer_module` is exported, and may be an issue if the label argument for this module is updated to anything other than `Report previewer` if a user chooses to customize the teal framework. 

https://github.com/insightsengineering/teal/blob/139971a42add5084c69928d7ca1271773f3c4e53/R/reporter_previewer_module.R#L15-L17

At the moment, I'd say it's very unlikely.

@chlebowa WDYT?

> It does but I have a feeling it shouldn't and it didn't use to.

I am not aware of any restrictions on module names. I'd suggest we pursue this option if we see it's happening to avoid scope creep of this PR.

donyunardi (changes requested):
Maybe a record in NEWS?

chlebowa:
> This line will always force "Report previewer" to be the last module, after all other modules.

Not always because one can run an app with modules that do not use reporters and name the first module `"Report previewer"`, however unlikely that may be.

> `reporter_previewer_module` is exported, and may be an issue if the label argument for this module is updated to anything other than `Report previewer` if a user chooses to customize the teal framework.

Then again, the module is not meant to be called explicitly and is added automatically by `srv_teal`:
```
    if (is_arg_used(modules, "reporter") && !is_any_previewer(modules)) {
      modules <- append_module(modules, reporter_previewer_module())
    }
```

R/module_filter_manager.R
```diff
@@ -164,9 +164,10 @@ filter_manager_srv <- function(id, filtered_data_list, filter) {
           rownames(mm) <- ""
         }
 
-        mm
+        # Report Previewer will not be displayed.
+        mm[!grepl("Report previewer", names(mm))]
```
donyunardi, line 168: > Not always because one can run an app with modules that do not use reporters and name the first module "Report previewer", however unlikely that may be.

I am aware of that. To make it clear, I was thinking something like this:

```suggestion
        # Report Previewer will not be displayed.
        if ("Report previewer" %in% names(mm)[length(mm)]) {
          mm <- mm[, -length(mm)]
        }
        
        mm
```

This way, even though we have a user module with "Report previewer" as the label name, it will not be removed. 
We only check the last element of mm due to current teal behavior.

WDYT?

chlebowa: But if you label a module `"Report previewer"` in an app without reporters, it will be removed, while it should not.

R/modules.R
```diff
@@ -212,7 +212,18 @@ module <- function(label = "module",
...
+      "Label 'global_filters' is reserved in teal. Please change to something else.",
+      call. = FALSE
+    )
+  }
+  if (label == "Report previewer") {
+    stop(
+      sprintf("module(label = \"%s\", ...\n  ", label),
+      "Label 'Report previewer' is reserved in teal.",
+      call. = FALSE
+    )
```
donyunardi, line 226:
```suggestion
  if (label %in% c("global_filters", "Report previewer")) {
    stop(
      sprintf(
        "module(label = %1$s, ...
        Label %1$s is reserved in teal. Please change to something else.
        ",
        label
      ),
      call. = FALSE
    )
  }
```

Perhaps we can combine the code for optimization?

donyunardi (dismissed):
Made suggestion on optimizing some code.
Everything else looks good.

Thanks @chlebowa !

R/modules.R
chlebowa: The line break within the `fmt` string adds loads of white space to the message, you should do 
```
    sprintf(
      "module(label = \"%s\", ...\n  %s",
      label,
      "Label 'global_filters' is reserved in teal. Please change to something else."
    )
```

The benchmark is exactly the same, down to a microsecond.
```
Unit: microseconds
    expr    min      lq     mean  median     uq     max neval cld
     new 36.793 41.2085 51.34256 44.1210 52.019 275.206   500   a
 current 36.873 41.2010 51.51580 44.2335 50.337 289.881   500   a
```

The current is 3 lines less.

PR #921, opened 2023-09-20
update `teal.slice` dependency

chlebowa:
quick fix

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -32,7 +32,7 @@ Depends:
     R (>= 4.0),
     shiny (>= 1.7.0),
     teal.data (>= 0.3.0),
-    teal.slice (>= 0.4.0),
+    teal.slice (>= 0.4.0.0011),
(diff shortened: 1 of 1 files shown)


DESCRIPTION
```diff
@@ -32,7 +32,7 @@ Depends:
     R (>= 4.0),
     shiny (>= 1.7.0),
     teal.data (>= 0.3.0),
-    teal.slice (>= 0.4.0),
+    teal.slice (>= 0.4.0.0011),
```
averissimo, line 35:
```suggestion
    teal.slice (>= 0.4.0.9011),
```

averissimo (approved):
Thanks @chlebowa !!

This will require another step just before the release so it's bumped to whatever `teal.slice` version is next

> - [ ] Make sure that the minimum dependency versions are updated in the DESCRIPTION file for the package.

_(from the release template)_

PR #924, opened 2023-09-25
904 teal_data in teal_module

### LINKED ISSUE #904: [Feature request] Rework `<teal_module>$server` to get `tdata` object
After https://github.com/insightsengineering/teal.data/pull/161 modules should be refactored (all teal.modules.*) to receive `tdata` object from `teal.data`.

Dilemmas to solve:
- what to do with `tdata` class in `teal` which is sent to `data` argument in `teal_module$server`? Should we soft-deprecate?
- Should we enable another object to work with `data_merge` and `data_extract` module which is a candidate to refactor?

chlebowa:
Closes #904 

Added `as_tdata` function to convert the data object received by modules from `teal_data` to `tdata` class.
Modified `srv_nested_tabs.teal_module` now returns a `teal_data`.

To make a module handle a `teal_data` object without (much) modification, the incoming `data` argument should be downgraded to the old class.
```
      data <- as_tdata(data)
```

Also modified `example_module` to handle `teal_data` objects.


R/module_nested_tabs.R
```diff
@@ -285,45 +283,33 @@ srv_nested_tabs.teal_module <- function(id, datasets, modules, is_module_specifi
...
   checkmate::assert_class(datasets, "FilteredData")
-  checkmate::assert_class(trigger_data, "reactiveVal")
 
   datanames <- if (is.null(module$datanames)) datasets$datanames() else module$datanames
 
   # list of reactive filtered data
   data <- sapply(
     datanames,
-    function(x) eventReactive(trigger_data(), datasets$get_data(x, filtered = TRUE)),
+    function(x) datasets$get_data(x, filtered = TRUE),
```
gogonzo, line 298: This `eventReactive(trigger_data(), ...)` is very important for teal. This blocks invisible inactive tabs to invalidate data reactives. Removing this causes slowdown of teal application, because every change in the filter panel causes recalculation in all modules.

chlebowa: Don't we lose this separation anyway by putting all datasets into one `qenv`?

gogonzo: No, because each teal_module listens to some data related reactives. Request goes up to teal and if we remove this  `eventReactive(trigger_data(), ...)` then all module will be recalculated (because filtered data has changed).
`trigger_data()` is based on module visibility, so if module is not-visible, `trigger_data()` doesn't change.

R/module_nested_tabs.R
```diff
@@ -243,7 +243,7 @@ srv_nested_tabs.teal_module <- function(id, datasets, modules, is_module_specifi
     }
 
     if (is_arg_used(modules$server, "data")) {
-      data <- .datasets_to_data(modules, datasets, trigger_data)
+      data <- reactive(.datasets_to_data(modules, datasets, trigger_data))
```
gogonzo, line 246: `trigger_data` isn't used in `.datasets_to_data`
```suggestion
      data <- eventReactive(trigger_data(), .datasets_to_data(modules, datasets))
```

chlebowa: What about 
```
  data <- sapply(
    datanames,
    function(x) eventReactive(trigger_data(), datasets$get_data(x, filtered = TRUE)),
    simplify = FALSE
  )
```

R/module_nested_tabs.R
```diff
@@ -307,23 +304,16 @@ srv_nested_tabs.teal_module <- function(id, datasets, modules, is_module_specifi
   )
 
   hashes <- calculate_hashes(datanames, datasets)
-  metadata <- lapply(datanames, datasets$get_metadata)
-  names(metadata) <- datanames
 
-  new_tdata(
+  teal.data::new_teal_data(
```
gogonzo, line 308: if data is a list of eventReactives then this constructor doesn't work

```
  data <- sapply(
    datanames,
    function(x) eventReactive(trigger_data(), datasets$get_data(x, filtered = TRUE)),
    simplify = FALSE
  )
```

chlebowa: This constructor works on `FilteredData`, does it not?

gogonzo: `teal.data::new_teal_data` has no dependency on `FIlteredData` - it is generic constructor of teal_data class. In your code here `teal.data::new_teal_data` uses `data` which is a list of reactives (see like 300). This is destined to fail. This is why I suggested to remove `trigger_data` from this function and use it in a parent frame like below

```
data <- eventReactive(trigger_data(), .datasets_to_data(modules, datasets))
```

chlebowa: I clearly misunderstand. We are talking about `datasets_to_data` which converts `FilteredData` to `teal_data`, yes?

gogonzo: Yes, we are discussing `.datasets_to_data`.  Below lines 300-309, let's go through these calls:

- data is a list of reactives which return filtered data
- hashed - let's ignore this
- teal.data::new_teal_data(data, ...) - call to create teal_data object from list of reactives. Call itself probably wouldn't throw because we don't care what developer pushes to qenv (there is no assert on dataset). Instead of passing data.frames you are passing reactives ;]


```r
data <- sapply(
    datanames,
    function(x) eventReactive(trigger_data(), datasets$get_data(x, filtered = TRUE)),
    simplify = FALSE
  )

  hashes <- calculate_hashes(datanames, datasets)

  teal.data::new_teal_data(
    data,
    ...
  )


```

chlebowa: So what you're saying is `trigger_data` **shouldn't** be used in `.datasets_to_data`? And `.datasets_to_data` should be called in response to `trigger_data`?

gogonzo: Yes, I think [this comment](https://github.com/insightsengineering/teal/pull/924#discussion_r1342628538) is exactly what we need. `.datasets_to_data` shouldn't use `trigger_data`

chlebowa: Got it :+1:

R/module_nested_tabs.R
```diff
@@ -243,7 +243,7 @@ srv_nested_tabs.teal_module <- function(id, datasets, modules, is_module_specifi
     }
 
     if (is_arg_used(modules$server, "data")) {
-      data <- .datasets_to_data(modules, datasets, trigger_data)
+      data <- reactive(.datasets_to_data(modules, datasets))
```
gogonzo, line 246: `trigger_data()` is one of the most important lines in teal. It triggers data only when module is visible. Please read comments (219-222) and see the code (223-233).
 
```suggestion
      data <- eventReactive(trigger_data(), .datasets_to_data(modules, datasets))
```

R/dummy_functions.R
```diff
@@ -92,23 +92,36 @@ example_module <- function(label = "example teal module", datanames = "all") {
...
         teal.widgets::verbatim_popup_srv(
           id = "rcode",
-          verbatim_content = attr(data, "code")(),
+          verbatim_content = attr(data_downgraded(), "code")(),
           title = "Association Plot"
         )
       })
     },
     ui = function(id, data) {
+      data_downgraded <- reactive(.tdata_downgrade(data))
```
gogonzo, line 118: There are two `ui`s. 

Another important thing:
- please don't (ever) send `data` to `ui`, this is one of the reasons we have a refactor.

R/module_nested_tabs.R
```diff
@@ -306,29 +301,22 @@ srv_nested_tabs.teal_module <- function(id, datasets, modules, is_module_specifi
   # list of reactive filtered data
   data <- sapply(
     datanames,
-    function(x) eventReactive(trigger_data(), datasets$get_data(x, filtered = TRUE)),
+    function(x) isolate(datasets$get_data(x, filtered = TRUE)),
```
gogonzo, line 304: Function is run in a reactive context (line 246) so you can remove isolate (you'll have just two tests to fix)

```suggestion
    function(x) isolate(datasets$get_data(x, filtered = TRUE)),
```

gogonzo (commented):
`example_module` doesn't need downgraded data. `.tdata_downgrade` isn't needed at all - why would someone downgrade module which works with `teal_data`?

If you apply my comments and fix example_module to be as follows - everything should work (tested already on my side)

<details>
<summary> example_module </summary>

```r
example_module <- function(label = "example teal module", datanames = "all") {
  module(
    label,
    ui = function(id) {
      ns <- NS(id)
        output = dataTableOutput(ns("table")),
        encoding = div(
        )
      )
    },
    server = function(id, data) {
      checkmate::assert_class(data, "reactive")
        observeEvent(data(), {
          updateSelectInput(session, "dataname", choices = teal.data::datanames(data()))
        })

        output$table <- renderDataTable(data()[[input$dataname]])

          id = "rcode",
          verbatim_content = reactive(teal.code::get_code(data())),
          title = "Association Plot"
        )
      })
    },
    datanames = datanames
  )
}

```
</details>

R/dummy_functions.R
chlebowa: You will notice that it was `ui = function(id, data)` before this PR.

The two UI functions were a merge artefact.

chlebowa:
> `example_module` doesn't need downgraded data. `.tdata_downgrade` isn't needed at all - why would someone downgrade module which works with `teal_data`?

As previously discussed, `.tdata_downgrade` may be necessary for external `teal` modules that will NOT wirk with `teal_data`.
Putting it in `example_module` was a proof of concept and an example to be discussed. See opening post.

gogonzo (approved):
Clean!

PR #925, opened 2023-10-04
Introduce teal_data class

gogonzo:
Followup to:
- https://github.com/insightsengineering/teal.data/pull/178


R/init.R
```diff
@@ -116,7 +116,10 @@ init <- function(data,
   logger::log_trace("init initializing teal app with: data ({ class(data)[1] }).")
   data <- teal.data::to_relational_data(data = data)
```
vedhav, line 117: I think this is a residual code that needs to be removed

R/init.R
```diff
@@ -138,7 +140,7 @@ init <- function(data,
 
   # resolve modules datanames
   datanames <- teal.data::get_dataname(data)
-  join_keys <- data$get_join_keys()
+  join_keys <- teal.data::get_join_keys(data)
```
chlebowa, line 143: I believe this line makes the `datanames` argument to `new_teal_data` mandatory (docs say "required when...").

R/module_nested_tabs.R
```diff
@@ -314,12 +314,14 @@ srv_nested_tabs.teal_module <- function(id, datasets, modules, is_module_specifi
...
-        teal.slice::get_filter_expr(datasets, datanames)
-      )
+      {
+        c(
+          get_rcode_str_install(),
+          get_rcode_libraries(),
+          get_datasets_code(datanames, datasets, hashes),
+          teal.slice::get_filter_expr(datasets, datanames)
+        )
+      }
```
chlebowa, line 324: Since this is a single call the braces are not required. If this is a style choice, ok.

R/module_teal.R
```diff
@@ -162,10 +162,7 @@ srv_teal <- function(id, modules, raw_data, filter = teal_slices()) {
 
     # loading the data -----
     env <- environment()
-    datasets_reactive <- reactive({
-      if (is.null(raw_data())) {
-        return(NULL)
-      }
+    datasets_reactive <- eventReactive(raw_data(), ignoreNULL = TRUE, {
```
chlebowa, line 165: Any reason for explicitly setting `ignoreNULL` to default?

gogonzo: No, I should remove it because eventReactive has `ignoreNULL = TRUE` by default, which is exactly what we need here.

R/init.R
```diff
@@ -114,9 +114,11 @@ init <- function(data,
                  footer = tags$p(),
                  id = character(0)) {
   logger::log_trace("init initializing teal app with: data ({ class(data)[1] }).")
-  data <- teal.data::to_relational_data(data = data)
 
-  checkmate::assert_class(data, "TealData")
+  if (!inherits(data, c("TealData", "teal_data"))) {
+    data <- teal.data::to_relational_data(data = data)
+  }
```
pawelru, line 120: you can delegate this if into the method -> https://github.com/insightsengineering/teal.data/pull/178#discussion_r1365312526

I expect that you need to do this in many places so this could be more DRY

R/utils.R
```diff
@@ -47,6 +47,29 @@ include_parent_datanames <- function(dataname, join_keys) {
...
+  teal.slice::init_filtered_data(
+    x = as.list(x@env)[datanames],
+    join_keys = x@join_keys,
+    code = teal.data:::CodeClass$new(
+      code = paste(teal.code::get_code(x), collapse = "\n"),
+      dataname = teal.data::get_dataname(x)
+    ),
+    check = FALSE
+  )
+}
```
vedhav, line 71: I would still prefer to have this implementation in the `teal.slice` package along with the other `FilterData-utils` methods. This way we can keep on using the `teal.slice::init_filtered_data` without deprecating it, the only deprecation will be that it will stop supporting the `TealData` class.

vedhav: Maybe I overlooked something

chlebowa: I imagine this will be outright replaced by `init_filtered_data`, which by then will only accept `teal_data` and have no other methods. However, as long the `TealData` class exists instantiating `FilteredData` on those objects has to be supported and then all methods of `init_filtered_data` must be somewhat consistent.

gogonzo: I have an opinion that `teal.slice` should limit dependency on other `teal*` classes. `init_filtered_data` should be a generic function which needs only `data`, `join_keys`. Other stuff like `code`, `metadata`, `datalabel` etc. are not needed. Constructor could look as simple as this:

```r
init_filtered_data(
  data = list(adsl = adsl, adtte = adtte),
  join_keys = join_keys() # could be empty
)
```

`join_keys` are now `JoinKeys` (R6) class. But it could be easily represented as list. So JoinKeys could be simplified and `teal.slice` would not depend on any `teal` package.

vedhav: Thank you both for the perspectives. I agree.

R/module_teal_with_splash.R
```diff
@@ -65,17 +66,37 @@ srv_teal_with_splash <- function(id, data, modules, filter = teal_slices()) {
       shinyjs::showLog()
     }
 
-    is_pulled_data <- teal.data::is_pulled(data)
     # raw_data contains TealDataAbstract, i.e. R6 object and container for data
     # reactive to get data through delayed loading
     # we must leave it inside the server because of callModule which needs to pick up the right session
```
pawelru, line 71: is this still true?

R/module_teal_with_splash.R
```diff
@@ -65,17 +66,37 @@ srv_teal_with_splash <- function(id, data, modules, filter = teal_slices()) {
...
-    if (is_pulled_data) {
-      raw_data <- reactiveVal(data) # will trigger by setting it
+    raw_data <- if (inherits(data, "teal_data")) {
+      reactiveVal(data)
+    } else if (teal.data::is_pulled(data)) {
```
pawelru, line 74: just for clarity for future work - can you please add class condition so that it's clear that it's for `TealDataAbstract`

R/utils.R
```diff
@@ -47,6 +47,29 @@ include_parent_datanames <- function(dataname, join_keys) {
...
 
+
+
+#' Create a `FilteredData`
+#'
+#' Create a `FilteredData` object from a `teal_data` object
+#' @param x (`teal_data`) object
+#' @return (`FilteredData`) object
+#' @keywords internal
+teal_data_to_filtered_data <- function(x) { #     nolint
```
pawelru, line 58: why `# nolint`?

R/utils.R
```diff
@@ -47,6 +47,29 @@ include_parent_datanames <- function(dataname, join_keys) {
...
+  checkmate::assert_class(x, "teal_data")
+  datanames <- x@datanames
+
+  teal.slice::init_filtered_data(
+    x = as.list(x@env)[datanames],
+    join_keys = x@join_keys,
+    code = teal.data:::CodeClass$new(
```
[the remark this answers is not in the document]
gogonzo: `CodeClass` will be removed soon after internal release and we won't have this problem anymore. Better, FilteredData won't contain `CodeClass` anymore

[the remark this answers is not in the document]
pawelru: this is great to hear!
How about CRAN release then? Wouldn't it complain about this?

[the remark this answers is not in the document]
gogonzo: When releasing to CRAN all everything be removed and will depend on `teal_data` classes and methods. There will be no `TealData` and friends in the whole code base (except teal.connector.xyz where I plan to make a switch - `if (packageVersion(teal.data) < 0.4.0) THEN do old way  ELSE do new way`)

R/init.R
pawelru: let's continue discussing in the aforementioned thread - closing this one

R/init.R
```diff
@@ -114,9 +114,11 @@ init <- function(data,
...
+  checkmate::assert_multi_class(data, c("TealData", "teal_data"))
```
chlebowa, line 121: I don't understand this assertion in the context of the preceding conversion of these classes to `TealData`.

Also, am I right to assume this function will be removed during the final changes?

R/init.R
```diff
@@ -151,6 +153,17 @@ init <- function(data,
...
+            sprintf(
+              "Module %s has datanames that are not available in a 'data':\n %s not in %s",
+              modules$label,
+              toString(extra_datanames),
+              toString(datanames)
+            )
+          )
+        }
         datanames_adjusted <- intersect(modules$datanames, datanames)
         include_parent_datanames(dataname = datanames_adjusted, join_keys = join_keys)
```
chlebowa, line 168: This seems to be neither assigned nor returned.

R/init.R
```diff
@@ -142,7 +144,7 @@ init <- function(data,
 
   # resolve modules datanames
   datanames <- teal.data::get_dataname(data)
-  join_keys <- data$get_join_keys()
+  join_keys <- teal.data::get_join_keys(data)
   resolve_modules_datanames <- function(modules) {
```
chlebowa, line 148: Can this function be moved to `utils.R`?

R/module_teal_with_splash.R
```diff
@@ -52,36 +52,59 @@ ui_teal_with_splash <- function(id,
...
+          # raw_data is a reactive which returns data only when submit button clicked
+          # otherwise it returns NULL
+          do.call(
+            teal.data::teal_data,
+            c(
+              lapply(data$get_datasets(), function(x) x$get_raw_data()),
+              code = data$get_code(),
+              join_keys = data$get_join_keys()
+            )
+          )
```
chlebowa, line 96: This is duplicated. Can it be avoided?

R/init.R
gogonzo: Assertion will be replaced by `checkmate::assert_multi_class(data, c("teal_data", "ddl"))`

R/init.R
gogonzo: yes it is. This is the code: 

```r
      modules$datanames <- if (identical(modules$datanames, "all")) {
        datanames
      } else if (is.character(modules$datanames)) {
        extra_datanames <- setdiff(modules$datanames, datanames)
        if (length(extra_datanames)) {
          stop(
            sprintf(
              "Module %s has datanames that are not available in a 'data':\n %s not in %s",
              modules$label,
              toString(extra_datanames),
              toString(datanames)
            )
          )
        }
        datanames_adjusted <- intersect(modules$datanames, datanames)
        include_parent_datanames(dataname = datanames_adjusted, join_keys = join_keys)
      }

``

R/module_teal_with_splash.R
gogonzo: will be removed soon also. Conversion from TealData to teal_data is temporary

R/init.R
chlebowa: Right!

R/init.R
m7pr: was thinking about the same

R/module_teal_with_splash.R
gogonzo: Don't worry, I know every line in `teal`. It will be removed as soon as ddl touches main 👍

tests/testthat/test-init.R
chlebowa, line 1: should `init` accept empty `teal_data`?
Currently does but then error is raised by `init_filtered_data` downstream.

chlebowa (approved):

Issue #895, opened 2023-08-11
[summary] `data` refactor


gogonzo:
WIP....

1. Simplifying `cdisc_data` and `teal_data`

Specifying data in teal can be changed to simply specifying a list of datasets and code (automatically parsed) 

```r
library(scda)
ADSL <- synthetic_cdisc_data("latest")$adsl
ADTTE <- synthetic_cdisc_data("latest")$adtte
ADRS <- synthetic_cdisc_data("latest")$adrs


data <- teal.data::cdisc_data(
  ADSL = ADSL, ADTTE = ADTTE, ADRS = ADRS,
  code = quote({
    ADSL <- scda::synthetic_cdisc_data("latest")$adsl
    ADTTE <- scda::synthetic_cdisc_data("latest")$adtte
    ADRS <- scda::synthetic_cdisc_data("latest")$adrs
  })
)
```

In the same time, it is still possible to use legacy `cdisc_dataset` which is not soft deprecated
```
cdisc_data(
    cdisc_dataset("ADSL", ADSL),
    cdisc_dataset("ADTTE", ADTTE),
    cdisc_dataset("ADRS", ADRS),
    code = '
      ADSL <- scda::synthetic_cdisc_data("latest")$adsl
      ADTTE <- scda::synthetic_cdisc_data("latest")$adtte
      ADRS <- scda::synthetic_cdisc_data("latest")$adrs
    '
  )
```

2.

Issue #860, opened 2023-07-05
[Feature Request]: Make a `data` module


gogonzo:
### Feature description

After [establishing reactive link between data and modules](https://github.com/insightsengineering/teal/issues/669). We should decide how we want data module to be displayed. Should it be:
- separate tab
- sidebar panel

Restrictions:
- It can't be a modal as some connectors using modal, and it's not possible to open modal from modal (there can be only one modal at the time)

PR #926, opened 2023-10-04
Introduce the new DDL (ddl class)

gogonzo:

## install
<details>
<summary>Installation code </summary>

```r
staged.dependencies::dependency_table(
  project = "insightsengineering/teal@https://github.com",
  project_type = "repo@host",
  ref = "ddl@teal_data@main",
  verbose = 1
) |> staged.dependencies::install_deps()
```
</details>

## ddl class

<details><summary>unfold</summary>

Small class which contains `ui` and `server`. `server` calls `run_ddl` which evaluates a `code` using input and create `teal_data` object. Function `ddl` accepts several arguments which are used to generate proper `ui` and `server`.

tdata-DDL process (3)

Above diagram depict how class works. Please follow steps below and look at the diagram. 

1. One needs to specify `ddl` arguments to fit someone's needs. Take for example `custom_connector`. Let's assume that we need to call `scda::synthetic_cdisc_data('latest')` for two datasets.

```r
data_code = paste(
    "adsl <- scda::synthetic_cdisc_data('latest')$adsl",
    "adtte<- scda::synthetic_cdisc_data('version')$adtte",
    collapse = "\n"
  )
)
```

Imagine that above datasets are on the remote server so we need to call `open_conn` function and close after pull. Imagine that `open_conn` and `close_conn` are in some external package. 

```r
code = paste(c(
  "open_conn(username = input$username, password = input$password)",
  "adsl <- scda::synthetic_cdisc_data('latest')$adsl",
  "adtte<- scda::synthetic_cdisc_data('version')$adtte",
  "close_conn()"
  ),
  collapse = "\n"
)
```

`username` and `password` should be specified by the app user. To do so we need to create `ui` which will create `username` and `password` inputs which are used during evaluation of above code.

```r
ui = function(id) {
  ns <- NS(id)
  tagList(
  )
}
```

If we have `ui` we need also `server` to utilize input. When submit button is clicked all `input` values are passed to `ddl_run` where all magic happens. `ddl_run` executes the code and substitutes all `input$<name>` with equivalent `input_list` element. `server` function needs to have `...` in the formals where all necessary objects are passed through. 

```r
submit_button_server <- function(id, ...) {
    tdata <- eventReactive(input$submit, {
      req(input$pass)
      ddl_run(input_list = reactiveValuesToList(input), ...)
    })

    # would need to make sure we handle reactivity correctly here as teal::init expects not reactive tdata...
    return(tdata)
  })
}
```

Because user `username` and `password` is not something we would like to include in the show R code. We need to replace them with some other code which can be run by anyone who receives the code. To replace args with some desired value or call we use `offline_args` to substitute code with these.

```
offline_args = list(
  username = quote(askpass::askpass("Please enter username")),
  password = quote(askpass::askpass("Please enter password"))
)  
```

At this point `ddl_run` evaluated "online" code, and has environment with the results of the evaluation. Also "offline" code is available after substitution. At the end `teal_data` is constructed from data and offline code

```r
teal_data(data = env_list, code = masked_code, join_keys)
```

~~⚠️ As one can see `ddl_run` require only `input_list`. It is a simplification for potential app developer who doesn't need to handle additional arguments (`code`, `input_mask`) in the `server`. Consequence of this simplification is that `server` specified in `ddl()` have an enclosing environment changed to the environment where `ddl_run`, `code`, `input_mask` is available. Another consequence is that `ddl_run` theoretically doesn't exist anywhere - it is created on the fly when `ddl` object is created. It means that developing own `ddl` connector in external "pkg" can't use private functions from that "pkg".~~

</details>

## Example teal app with ddl

<details>
<summary>teal app code</summary>

use `teal.data`, `teal` and `teal.slice` from `refactor` branch

```r
library(scda)
library(teal)

data <- ddl(
  code = '
    open_conn(username = input$username, password = input$password)
  ',
  input_mask = list(
  ),
  ui = function(id) {
    ns <- NS(id)
    tagList(
    )
  },
  server = function(id, ...) {
      eventReactive(input$submit, {
        ddl_run(input = input, ...)
      })
    })
  },
  datanames = c("ADSL", "ADTTE"),
  join_keys = cdisc_join_keys("ADSL", "ADTTE")
)

app <- teal::init(
  data = data,
  modules = list(
    teal.modules.general::tm_data_table("Data Table")
  )
)

runApp(app)

```

</details>

## Internal Connectors

https://github.com/insightsengineering/nestdevs-tasks/issues/38


R/utils.R
```diff
@@ -46,3 +46,26 @@ include_parent_datanames <- function(dataname, join_keys) {
...
+
+
+
+#' Create a `FilteredData`
+#'
+#' Create a `FilteredData` object from a `teal_data` object
+#' @param x (`teal_data`) object
+#' @return (`FilteredData`) object
+#' @keywords internal
+teal_data_to_filtered_data <- function(x) { #     nolint
```
chlebowa, line 58: `# nolint` may be redundant

R/utils.R
```diff
@@ -46,3 +46,26 @@ include_parent_datanames <- function(dataname, join_keys) {
...
+  checkmate::assert_class(x, "teal_data")
+  datanames <- x@datanames
```
chlebowa, line 60: `datanames` is only used once and in a simple call, consider using `x@datanames` directly.

R/ddl.R
```diff
@@ -0,0 +1,193 @@
...
+#' @param expr (`expression`)\cr
+#'  Syntatically valid R code to be executed in the shiny session.
+#'  shouldn't be specified when `code` is specified.
+#'
+#' @param code (`character`, `language`)\cr
+#'  Object containing code to be evaluated to load data. Shouldn't be specified when `expr`
+#'  is specified.
+#'
+#'
+#' @param ui (`shiny.tag`)\cr
```
chlebowa, line 14:
```suggestion
#' @param ui (`function`)\cr
```

R/ddl.R
```diff
@@ -0,0 +1,193 @@
...
+  if (missing(datanames)) {
+    stop("`dataname` argument is required")
+  }
+
+
+  # function creates  object from the code, input and input_mask
+  # function defined here to have access to the arguments
+  ddl_run <- function(input = list()) {
+    checkmate::assert_list(input)
+    if (inherits(input, "reactivevalues")) {
```
chlebowa, line 70:
```suggestion
    if (inherits(input, "reactiveValues")) {
```

R/ddl.R
```diff
@@ -0,0 +1,193 @@
...
+      input <- shiny::reactiveValuesToList(input)
+    }
+    env <- list2env(list(input = input))
+    # substitute by online args and evaluate
+    eval(code, envir = env)
+
+    if (identical(ls(env), character(0))) {
+      warning("DDL code returned NULL. Returning empty  object")
```
chlebowa, line 78: Inaccurate message, code may have returned but made no assignments.

R/ddl.R
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    # don't keep input further we don't want to keep input in the @env of teal_data
+    # but we want to keep other non-dataset objects created in the code
+    env_list <- as.list(env)
+    env_list <- env_list[names(env_list) != "input"]
```
chlebowa, line 84: I'm not sure about this. Shouldn't all created variables be available for later use? Is it a valid assumption that they will never be used again?

R/ddl.R
```diff
@@ -0,0 +1,193 @@
...
+
+    # substitute by offline args
+    for (i in names(input_mask)) {
+      input[[i]] <- input_mask[[i]]
+    }
```
chlebowa, line 89:
```suggestion
input[names(input_mask)] <- input_mask[names(input_mask)]
```

R/ddl.R
```diff
@@ -0,0 +1,193 @@
...
+#'   `shiny` ui module containing inputs which `id` correspond to the
+#'   args in the `code`.
+#'
+#' @param server (`function`)\cr
+#'   `shiny` server module returning data. This server should execute
+#'   `code` and return a reactive data containing necessary data. To handle
+#'   evaluation and code masking process it is recommended to use `ddl_run`.
+#'   Package provides universal `username_password_server` which
+#'   runs [ddl_run] function, which returns `teal_data` object.
+#'   Details in the the example
```
pawelru, line 24:
```suggestion
#'   Details in the the example.
```

R/ddl.R
```diff
@@ -0,0 +1,193 @@
...
+ddl <- function(expr,
+                code,
+                ui = submit_button_ui,
+                input_mask = list(),
+                server = submit_button_server,
+                join_keys = teal.data::join_keys(),
+                datanames) {
+  if (!missing(expr) && !missing(code)) {
+    stop("Only one of `expr` or `code` should be specified")
+  }
```
pawelru, line 53: [`rlang::check_exclusive`](https://rlang.r-lib.org/reference/check_exclusive.html)

(please implement in other places if necessary)

R/ddl.R
```diff
@@ -0,0 +1,193 @@
...
+  if (!missing(expr)) {
+    code <- substitute(expr)
+  }
+  if (is.character(code)) {
+    code <- parse(text = code)
+  }
+
+  if (missing(datanames)) {
+    stop("`dataname` argument is required")
+  }
```
pawelru, line 63: [`rlang::check_required`](https://rlang.r-lib.org/reference/check_required.html)

The advantage of this is that we will be having consistent error message that end-users might be already familiar with.

(please implement in other places if necessary)

DESCRIPTION
pawelru, line 1: Please add NEWS + minimal version required

R/utils.R
pawelru: yes - please use getters from teal.data API instead of S4 accessors to loosen coupling to its implementation

R/ddl.R
chlebowa: I'm not particularly familiar with `rlang` error messages myself.

R/ddl.R
chlebowa: `argument "datanames" is missing, with no default`
This is a very frequent error message in R and should be familiar enough.

R/ddl.R
```diff
@@ -0,0 +1,210 @@
+#' DDL object
+#'
+#' Object to execute custom DDL code in the shiny session
+#'
+#' @param expr (`expression`)\cr
+#'  Syntatically valid R code to be executed in the shiny session.
```
kpagacz, line 6:
```suggestion
#' Syntactically valid R code to be executed in the shiny session.
```

R/ddl.R
```diff
@@ -0,0 +1,210 @@
...
+#'
+#' @param server (`function`)\cr
+#'   `shiny` server module returning data. This server should execute
+#'   `code` and return a reactive data containing necessary data. To handle
+#'   evaluation and code masking process it is recommended to use `ddl_run`.
+#'   Package provides universal `username_password_server` which
+#'   runs `ddl_run` function, which returns `teal_data` object.
+#'   Details in the the example
+#'
+#' @param input_mask (`list` named)\cr
```
kpagacz, line 26: This name could be more precise. Something like: `obfuscated_inputs` or `hidden_inputs` or `replaced_inputs` better conveys the intentions, I think. Masking has a very specific meaning in either statistical programming or programming in general, and it has little to do with making sure that something is not visible.

R/ddl.R
```diff
@@ -0,0 +1,210 @@
...
+  #  `ddl_run(input = input)`
+  # instead of
+  #  `ddl_run(x = ddl_object, input = input))`
+  ddl_run <- function(input = list()) {
+    ddl_run2(x = ddl_object, input = input)
+  }
+
+  # changing enclosing environment of the server to have access to ddl_fun function
+  # Thanks to this ddl object contains only ui and server functions
+  #  and server function can be run just by calling ddl$server("<id>")!
```
kpagacz, line 86: And pray people don't bind ddl_run in their server functions, huh? ;)

kpagacz: I don't think this change is what you really want because people might in their server functions reference variables present in the enclosing environment of `server`. If you change the environment, you sever that connection. A smarter idea is to either:
* add `ddl_run` to the present enclosing environment of `server`
* insert another environment in the tree that contains only the `ddl_run` function. The environment tree would look like: ddl_run_env <- the old enclosing environment. I haven't tried, but it should work.

kpagacz: I also guess that the current proposition has some chances of working in an interactive session, due to how environment() is going to have global env as an ancestor, so the values defined outside of server can be found anyway, but it really is going to break if people want to use a server function defined in a separate package, because the enclosing environment of a package function is usually "older" than the global env.

R/ddl.R
kpagacz: rlang error system ftw!

R/ddl.R
gogonzo: Thanks @kpagacz . Agree and not. "Masking" in R has a specific meaning which is a synonym of replacement or overriding. For example overriding function when loading namespaces is called "masking" ("function x has been masked"). [data-masking](https://search.r-project.org/CRAN/refmans/rlang/html/topic-data-mask.html) is apparently a thing.

R/ddl.R
gogonzo: Thanks for highlighting this! This is a very good remark and we will definitely rediscuss this.

R/ddl.R
pawelru: One of the main features of `rlang@1.0.0` is better display of the error messages - [link](https://www.tidyverse.org/blog/2021/12/rlang-1-0-0-errors/). Since the release of this version, most (if not all?) of tidyverse packages followed this. That's why I do believe that this might be more coherent across all the packages one might use hence: easier to understand and take appropriate action.

R/ddl.R
chlebowa: To be fair, I was thinking in colloquial terms when I first proposed `mask` but it does express the meaning quite well.

R/ddl.R
m7pr: I had issues in the past year with `rlang` handling `errors`. Needed to use `rlang::cnd_message_lines(error)` on Windows on every `error` message in `tryCatch` to get an accurate `error` message. Otherwise errors were always empty. If people have Windows and old `rlang` versions they may not see an error at all.

m7pr: I would also vote up for striving away from `tidyverse` as they have a great history of breaking changes and tend to have multiple dependencies in it. I don't think general error messages in `rlang` are better than customized, personalized human-crafted ones that we thought through to inform the user in the most useful manner as possible.

pawelru: I do echo similar mixed feeling about tidyverse but this is actually about r-lib which I consider much more solid.

> I don't think general error messages in rlang are better than customized, personalized human-crafted ones that we thought through to inform the user in the most useful manner as possible.

Yes and no. Yes because the role of the error message is to clearly indicate what's wrong and why to allow a quick fix by the end users. No, because if we will be too verbose then it won't be clear anymore. The other thing that I am actually trying to address here is consistency. Just a few examples of our codebase that I managed to find within few minutes:
https://github.com/insightsengineering/teal.slice/blob/686f1f76ac6130325df9314fea419800c139f35a/R/teal_slice.R#L134-L135
https://github.com/insightsengineering/teal.modules.general/blob/eef36c1dda918fcbebe948d8bef189bd7e545974/R/tm_g_bivariate.R#L762-L764
https://github.com/insightsengineering/teal.data/blob/c04b7edef44d80fd4cdf0c6196d5c9c3de7c84b1/R/utils.R#L21-L24
https://github.com/insightsengineering/teal.slice/blob/686f1f76ac6130325df9314fea419800c139f35a/R/teal_slice.R#L134-L135
If we can delegate the check as well as error message for a separate functionality called from all of those places then we would achieve some consistency. It's mostly done via `checkmate` for individual objects checks. I hope that `rlang` is a good candidate to fill in the gaps of what cannot be done via `checkmate`.
One might raise that we don't want to introduce a new dependency but `rlang` is very likely to be already there (directly or indirectly). What's great about it is that this is a zero-dependency package. That changed my view on this package quite significantly.

m7pr: Those 4 examples are actually fine in my opinion. Very clear :)
If rlang has zero dependencies than I'm less against, but still it's better to have a package that has 8 dependencies, than a packages with 15 dependencies where 7 of them has zero dependencies. The overview of such package become repellent. Each package extends the time of the installation. And each package require us and some more advanced users who would like to read the code to understand another workflow. How many workflows can we use in a teal designed to be simple and seamless for R users?

R/ddl.R
```diff
@@ -0,0 +1,223 @@
+#' DDL object
+#'
+#' Object to execute custom DDL code in the shiny session
+#'
+#' @param expr (`expression`)\cr
+#'  Syntatically valid R code to be executed in the shiny session.
+#'  shouldn't be specified when `code` is specified.
```
m7pr, line 7:
```suggestion
#'  Should not be specified when `code` is specified.
```

R/ddl.R
```diff
@@ -0,0 +1,223 @@
+#' DDL object
+#'
+#' Object to execute custom DDL code in the shiny session
```
m7pr, line 3:
```suggestion
#' Object that executes custom DDL code in `shiny` session.
```

R/ddl.R
```diff
@@ -0,0 +1,223 @@
...
+#'
+#' @param expr (`expression`)\cr
+#'  Syntatically valid R code to be executed in the shiny session.
```
m7pr, line 6:
```suggestion
#'  Syntatically valid R code to be executed in the `shiny` session.
```

R/ddl.R
```diff
@@ -0,0 +1,223 @@
...
+#'  shouldn't be specified when `code` is specified.
+#'
+#' @param code (`character`, `language`)\cr
+#'  Object containing code to be evaluated to load data. Shouldn't be specified when `expr`
```
m7pr, line 10: it's more visible is you split to `verb NOT`
```suggestion
#'  Object containing code to be evaluated to load data. Should not be specified when `expr`
```

R/ddl.R
m7pr: I subconsciously understood the purpose of `input_mask` after the first time I saw it. I would eventually propose `input_masking` as an alternative name, but it's longer.

R/ddl.R
```diff
@@ -0,0 +1,223 @@
...
+#'   `shiny` server module returning data. This server should execute
+#'   `code` and return a reactive data containing necessary data. To handle
+#'   evaluation and code masking process it is recommended to use `ddl_run`.
+#'   Package provides universal `username_password_server` which
+#'   runs `ddl_run` function, which returns `teal_data` object.
+#'   Details in the the example
+#'
+#' @param input_mask (`list` named)\cr
+#'   arguments to be substituted in the `code`. These
+#'   argument are going to replace arguments set through
```
m7pr, line 28:
```suggestion
#'   arguments are going to replace arguments set through
```

R/ddl.R
```diff
@@ -0,0 +1,223 @@
...
+#'   `ui` and `server`. Example use case is when app user
```
m7pr, line 29:
```suggestion
#'   `ui` and `server`. Example use case is when ab app user
```

R/ddl.R
```diff
@@ -0,0 +1,223 @@
...
+#'   is asked to input a password and we'd like to skip this
+#'   input in the reproducible code. Typically users password
```
m7pr, line 31:
```suggestion
#'   input in the reproducible code. Typically user's password
```

R/ddl.R
```diff
@@ -0,0 +1,223 @@
...
+    join_keys = join_keys,
+    class = "ddl"
+  )
+
+  ddl_object
+}
+
+#' @name ddl_module
+#' @keywords internal
+NULL
```
m7pr, line 79: what is that for? is it a placeholder for future implementation?

R/ddl.R
```diff
@@ -0,0 +1,223 @@
...
+    tdata <- eventReactive(input$submit, {
+      ddl_run(input = input, ...)
+    })
+
+    # would need to make sure we handle reactivity correctly here as teal::init expects not reactive teal_data...
+    return(tdata)
+  })
+}
+
+#' substitute inputs in the code
```
m7pr, line 131:
```suggestion
#' Substitute inputs in the code
```

R/ddl.R
```diff
@@ -0,0 +1,223 @@
...
+#'
+#' Function replaces symbols in the provided code prefixed with `input$` or `input[["`
+#' by values of the `args` argument.
```
m7pr, line 134:
```suggestion
#' by values of the `args` parameter.
```

R/ddl.R
```diff
@@ -0,0 +1,223 @@
...
+    })
+
+    # would need to make sure we handle reactivity correctly here as teal::init expects not reactive teal_data...
+    return(tdata)
+  })
+}
+
+#' substitute inputs in the code
+#'
+#' Function replaces symbols in the provided code prefixed with `input$` or `input[["`
```
m7pr, line 133:
```suggestion
#' Replace symbols in the provided code prefixed with `input$` or `input[["`
```

R/ddl.R
```diff
@@ -0,0 +1,193 @@
...
+  ddl_run <- function(input = list()) {
+    checkmate::assert_list(input)
+    if (inherits(input, "reactivevalues")) {
+      input <- shiny::reactiveValuesToList(input)
+    }
+    env <- list2env(list(input = input))
+    # substitute by online args and evaluate
+    eval(code, envir = env)
+
+    if (identical(ls(env), character(0))) {
```
m7pr: What if some objects have `.` a dot prefixed to their names, so that they are hidden?

```suggestion
    if (identical(ls(env, all.names = TRUE), character(0))) {
```

```r
> ls()
character(0)
> .x <- 5
> ls()
character(0)
> ls(all.names = TRUE)
[1] ".Random.seed" ".x"  
```

m7pr: `.Random.seed` is only in `.GlobalEnv`
in `new.env()` you don't have that

```r
> x <- new.env()
> ls(x)
character(0)
> ls(x, all.names = TRUE)
character(0)
```

R/ddl.R
m7pr:
```suggestion
      warning("DDL code returned NULL. Returning empty object.")
```

R/ddl.R
m7pr: hell yeah

NAMESPACE
```diff
@@ -18,6 +20,7 @@ S3method(ui_nested_tabs,teal_modules)
 export("%>%")
 export(TealReportCard)
 export(as.teal_slices)
+export(ddl)
```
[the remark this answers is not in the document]
m7pr: Added a PR for this, and few more documentation cleanup that should happen on automated update https://github.com/insightsengineering/teal/pull/948

R/ddl.R
```diff
@@ -0,0 +1,223 @@
...
+  ddl_object <- structure(
+    list(ui = ui, server = server),
+    code = code,
+    input_mask = input_mask,
+    datanames = datanames,
+    join_keys = join_keys,
+    class = "ddl"
+  )
+
+  ddl_object
```
kartikeyakirar, line 74:
```suggestion
  structure(
    list(ui = ui, server = server),
    code = code,
    input_mask = input_mask,
    datanames = datanames,
    join_keys = join_keys,
    class = "ddl"
  )
```

gogonzo:
We are considering two approaches now to implement as DDL. Both are not exclusive, but it need to choose what approach should we promote and expose in our documentation.

# 1. `ddl` wrapper to create a module

Doesn't differ so much from the initial design. App developer can specify `code`, `input_mask`, `ui` and `server` and `datanames` and `join_keys`. Inside of `server` one needs to call `ddl_run` which create teal_data object, run the code and masks shiny inputs.

```r
library(scda)
pkgload::load_all("teal.data")
pkgload::load_all("teal.slice")
pkgload::load_all("teal")
data <- ddl(
  code = quote({
    open_conn(username = input$username, password = input$password)
  }),
  input_mask = list(
  ),
  ui = function(id) {
      ns <- NS(id)
      tagList(
      )
  },
  server = function(id, ...) {
        eventReactive(input$submit, {
          ddl_run(input = input, ...)
        })
      })
  },
  datanames = c("ADSL", "ADTTE"),
  join_keys = cdisc_join_keys("ADSL", "ADTTE")
)

app <- teal::init(
  data = data,
  modules = list(
    teal.modules.general::tm_data_table("Data Table")
  )
)

runApp(app)

```

# 2. self sufficient shiny module

Alternative solution expects that app developer write everything in `ui` `server` and doesn't call `ddl_run` but creates a `teal_data` in the own way. I use `ddl` constructor because at this moment it is implemented in teal, but this could be a single list. 

```r
library(scda)
pkgload::load_all("teal.data")
pkgload::load_all("teal.slice")
pkgload::load_all("teal")
data <- ddl(
  ui = function(id) {
    ns <- NS(id)
    tagList(
    )
  },
  server = function(id, ...) {
      eventReactive(input$submit, {
        data <- teal_data() |>
          within(
            {
              open_conn(username = username, password = password)
            },
            username = input$username,
            password = input$password
          )


        datanames(data) <- c("ADSL", "ADTTE")
        get_join_keys(data) <- cdisc_join_keys("ADSL", "ADTTE")

        data@code <- '
            open_conn(login = askpass::askpass(), password = askpass::askpass())
            ADTTE <- scda::synthetic_cdisc_data("latest")$adtte'

        data
      })
    })
  }
)

app <- teal::init(
  data = data,
  modules = list(
    teal.modules.general::tm_data_table("Data Table")
  )
)

runApp(app)
```


Any thoughts?

R/ddl.R
```diff
@@ -0,0 +1,227 @@
...
+}
+
+#' substitute inputs in the code
+#'
+#' Function replaces symbols in the provided code prefixed with `input$` or `input[["`
+#' by values of the `args` argument.
+#'
+#' @param code (`language`) code to substitute
+#' @param args (`list`) named list or arguments
+.substitute_inputs <- function(code, args) {
```
chlebowa, line 145:
```suggestion
.mask_code <- function(code, mask) {
```

pawelru:
I do believe that masking is not a requirement that applies for all of the use cases that we aim to support (e.g. a module with files load without any auth) so I will go with simpler examples without it. I personally find it way easier / more effective to learn the basic on a simpler examples. 
But then: why not both? We can include the topic of masking as an extra and elaborate on `input_mask` argument / recommend `substitute` for doing this.

R/ddl.R
```diff
@@ -0,0 +1,227 @@
...
+#' @param input_mask (`list`) containing inputs to be masked in the `code`
+#' @param datanames (`character`) names of the objects to be created from the code evaluation
+#' @param join_keys (`join_keys`) object
+#'
+#' @return `teal_data` object
+#'
+#' @export
+ddl_run <- function(input = list(), code, input_mask, datanames, join_keys) {
+  checkmate::assert_list(input)
+  if (inherits(input, "reactivevalues")) {
```
chlebowa, line 204:
```suggestion
  if (inherits(input, "reactiveValues")) {
```

R/ddl.R
```diff
@@ -0,0 +1,227 @@
...
+  # evaluate code and substitute input
+  data <- teal.code::eval_code(data, .substitute_inputs(code, args = input))
+
+  if (identical(ls(data@env), character(0))) {
+    warning("DDL code returned NULL. Returning empty object")
+  }
+
+  if (length(input_mask) > 0) {
+    # mask dynamic inputs with mask
+    input <- utils::modifyList(input, input_mask)
```
chlebowa, line 218:
```suggestion
    input_masked <- utils::modifyList(input, input_mask)
```

R/ddl.R
```diff
@@ -0,0 +1,227 @@
...
+
+    # replace code of teal_data with masked code
+    #  question: warnings and errors are not masked, is it ok?
+    data@code <- format_expression(.substitute_inputs(code, args = input))
```
chlebowa, line 222:
```suggestion
    data@code <- format_expression(.substitute_inputs(code, args = input_masked))
```

chlebowa, line 222: I don't think `format_expression` is needed here. It seems to be called solely for pasting the calls in `code` into a single string, which can very well be done with `paste`. Note that `.substitute_inputs` already unwraps the code into simple calls.

Side note: Maybe we should put `teal.code:::lang2calls` in `.substitute_inputs`?

chlebowa:
My opinion in short:
0. The topic is difficult enough with one version, introducing two versions will confuse the user even further.
1. Introducing the option to mask code is crucial. If I have to write a silly little module to pull my data and I see my credentials end up on the report, I will look for a different framework and I will tell everyone to do the same. Emphasis on _option_. If you want to upload a file, you don't provide a mask, simple as that. Having an optional argument will not interrupt learning if we write docs properly.
2. I dislike the initial design (1.1) for reasons I have explained at length. 
3. I don't like (1.2) either because it expects the user to manipulate code on their own. It also expects the user to manipulate slots, which we should (and do) discourage. The latter is a matter of implementation and can be dealt with but the former is a matter of purpose. I believe the user should not be able to decouple code from data, hence [a leaner `qenv` constructor](https://github.com/insightsengineering/teal.code/issues/145). `teal_data` is an unfortunately necessary departure from that principle. I would rather it be done behind the scenes by our code than explicitly by the user.

gogonzo:
> My opinion in short:

What about 2.1 and 2.2?

> 2. I dislike the initial design (1.1) for reasons I have explained at length.

What's wrong about 1.1 specifically? Calling `ddl` with `...`?

> 3. I don't like (1.2) either 

I don't like it either. Handling NSE inside of the server module was hard for me to write a concise code (example is a simplest I could do)

-------

Apart from the details. `ddl` is a shiny module to create `teal_data` object. I have a feeling that we will expose two vignettes or two cases to the user. 
1. Creating data in the teal session - here we will show simple example without masking
2. Creating data in the teal session (advanced) - here we will demonstrate `ddl_run` and how to handle masking

I'm pretty sure that in cases without masking it's less confusing to `eval_code/within` than using `ddl_run`

m7pr:
I like 1.1 as this is the solution I understood the motivation for and I see a possibility to pass authentication credentials, ans I guess that will probably be the most popular use case. The input_mask is visible for the end user. For 1.2 I had a flash thought that I see a benefit or this approach because we do not need to introduce ddl_run, however the chance to a user to edit the `@code` slot on their own, IMHO, defeats the purpose of what we are doing. The user should not need to know that our structures have specific slots, because this is already complicated. Also the user should not edit them, as our pipeline assumes those fields are created on specific occasions in specific ways.

m7pr:
For 2.1 and 2.2 are we going to not support masking at all? Does it mean there's going to be 'input$version' or 'input$password' in the final returned code, because that's the `code` parameter specification of `ddl`?

We could pre-pend the final resulted code with values of `input` so all is visible and runnable. This could be extended with an option `hide/show input values` so that input is either prefilled in the resulted code or there is just a placeholder to fill it.

Example:

```r
#show input values = TRUE
input <- list() 
input$version = '2023-07-17'
...
qenv@code
```

or for the `FALSE` value

```r
#show input values = FALSE
input <- list() 
input$version <- 'PLEASE FILL' 
... 
qenv@code
```

gogonzo:
> are we going to not support masking at all?

My intention was different. I wanted to present both cases (masked and unmasked) and ask following questions:

1. Do you prefer to use `ddl_run` to support masked code?
2. Do you prefer to use `ddl_run` for unmasked code?

My answers are:
1. I definitely prefer ddl_run 
2. I definitely prefer manual

kartikeyakirar:
Given the priority placed on security and robust code management, my inclination is toward approach 1.1, due to its input masking feature that ensures the protection of credentials and sensitive data.

While approach 1.2 might simplify the user experience by not using  the ddl_run, users to modify the code directly  might lead to additional confusion and complication.

chlebowa:
> > > 2. I dislike the initial design (1.1) for reasons I have explained at length.
> > 
> > 
> > What's wrong about 1.1 specifically? Calling `ddl` with `...`?
> 
> The fact that the user has to write a server and has to use `ddl_run` in it.

We have talked and I now retract my reservations. This is unavoidable.

R/ddl-modules.R
```diff
@@ -0,0 +1,59 @@
...
+submit_button_server <- function(id, ...) {
+  moduleServer(id, function(input, output, session) {
+    tdata <- eventReactive(input$submit, {
+      ddl_run(input = input, ...)
+    })
+
+    # would need to make sure we handle reactivity correctly here as teal::init expects not reactive teal_data...
+    return(tdata)
+  })
+}
```
vedhav, line 31: Am I right to assume that right now they are not complete and we will implement the shiny module such that the UI disappears after clicking the submit button?

gogonzo:
What about alternative solution of having special `teal_module_data` which can create ddl module? `tm_teal_data()` function will have the same arguments as current `ddl` and in the future it could [be another module tab](https://github.com/insightsengineering/teal/issues/860) 

Such "ddl" server will be called in the same way as other `teal_module` - by passing `server_args` to the `server`. `server_args` will be `code`, `input_mask`, `datanames`, `join_keys`.

R/ddl-modules.R
chlebowa: More or less, yes.

vedhav:
> What about alternative solution of having special teal_module_data which can create ddl module? 
I would be okay with such an implementation as long as it's clear that the ddl module is a "special" module. Ideally, I would like fewer changes to create a `ddl` class compared to the typical `teal_data` class that the users create. Right now (at least my thought is) all the teal modules are independent from the "code" and it's taken from the teal_data but if we implement a custom ddl module that violates these fundamental rules I would often question myself if there are exceptions to such fundamental rules about teal that I already learned. So, I would advocate for an additional argument in the `teal::init()` for ddl.

chlebowa:
> > What about alternative solution of having special teal_module_data which can create ddl module?
> 
> Ideally, I would like fewer changes to create a `ddl` class compared to the typical `teal_data` class that the users create. 

I agree.

> So, I would advocate for an additional arguments in the `teal::init()` for ddl.

I disagree. `init` is **the** topmost user facing function in the entire framework and should not be modified lightly.

R/init.R
```diff
@@ -114,11 +114,10 @@ init <- function(data,
                  footer = tags$p(),
                  id = character(0)) {
   logger::log_trace("init initializing teal app with: data ({ class(data)[1] }).")
-
-  if (!inherits(data, c("TealData", "teal_data"))) {
+  if (!inherits(data, c("TealData", "teal_data", "ddl"))) {
```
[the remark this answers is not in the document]
vedhav: Ah! just saw that we're not looking for doc suggestions at the moment.

chlebowa:
> What about alternative solution of having special `teal_module_data` which can create ddl module?

Some more details for clarity's sake:
This would be a third special class for `teal_module` objects, like `teal_module_previewer` and `teal_module_landing`. 
It would be created by its own module template function (with arguments: `ui`, `server`, `code`, `datanames`, `join_keys`, and optional `input_mask`). 
The user may provide UI and server logic but with some constrains: 1) the server must return a reactive expression that returns a `teal_data` object; 2) if the user wishes to mask code, they must create the `teal_data` by calling `ddl_run(input = input, ...)`(*) (ellipsis mandatory).
The module would create data that is supplied to the filter panel and on to analytical modules.

An important change would be that the data could be supplied to the application either by passing `teal_data` or a `teal_module_data` module. Whether we put the latter in the `data` argument or in `modules`, is on the table.


(*) the `ddl_run` function will be renamed

m7pr:
> What about alternative solution of having special teal_module_data which can create ddl module?

Thanks @chlebowa for the explanation in your last comment!
I like the idea of the new module, which would be a new class. Because we would already have a template, this would have a separate documentation, this could get a separate tab as well and also this gets it's own arguments and assertions.

m7pr:
And I guess `ddl_run` is only used on `Show R Code` action only for this new module?

chlebowa:
> And I guess `ddl_run` is only used on `Show R Code` action only for this new module?

No. `ddl_run` creates `teal_data` and masks code. What ends up in `Show R Code` is the masked code from `teal_data` (modified by the filter panel and analytical modules).

m7pr:
If `Show R Code` is the only place where we can expose the code to the user using the app, isn't that a good idea to work on the code only there and just work on avaluated that that is now just a character `teal_data()/qenv@code`? What's point of pre-masking if user might never hit `Show R Code`?

chlebowa:
Because it's difficult to arrange masking at the very end of the analysis as opposed to upon data creation.
```
* create ---> |oo| mask ---> + filter ---> + analysis ---> <o> view
* create ---> + filter ---> + analysis ---> |oo| mask ---> <o> view
```

m7pr:
I think it depends on where/when you pass masking parameters, instead passing to to `ddl` you could pass them to the module definition, that can pass that to `Show R Code` module.

chlebowa:
Oh, even better. Which module? You never want to show unmasked code so would you rather mask it once, upon creation, or multiple times, in analytical modules?

m7pr:
No, it's gonna be masked only in `Show R Code Modal` (`show_rcode_modal()`) that takes `rcode` which I presume is a character string (taken from `teal_data/qenv@code`) and you only substitute there. The masking could be a parameter of `show_rcode_modal`

m7pr:
Alrighty @chlebowa so on `Show R Code` and `Add Report Card` actions. Two actions would trigger the substitution, and you'd need to work on a character string (probably the same way as we are currently working on regexes in this PR).

chlebowa:
I don't see how this is in any wat better than masking in the `teal_data` upon creation.

m7pr:
You postpone the masking part until it's really necessary. If a person does not create a report or does not check the Shown R Code, then the substitution action is not triggered at all.

chlebowa:
Conversely, if you do want it, you have to do it twice.
And the action takes milliseconds.

m7pr:
Ok forget it. You can just not include it in DDL lol.

R/ddl-run.R
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  code_strings <- vapply(code, deparse1, character(1L))
+  # Replace input$ with .()
+  code_strings <- gsub("input\\$(\\w+\\.?\\w*)", "\\.(\\1)", code_strings)
+  code_strings <- gsub("(input\\$)(`[^`]+`)", "\\.(\\2)", code_strings)
+
+  # Replace input[[ with .()
+  code_strings <- gsub("(input\\[\\[\")(\\w+\\.?\\w*)(\"\\]\\])", "\\.(\\2\\)", code_strings)
+  code_strings <- gsub("(input\\[\\[\")(\\w+\\-\\w+)\"\\]\\]", ".(`\\2`)", code_strings)
```
kartikeyakirar, line 74:
```suggestion
  code_strings <- gsub("(input\\$)([a-z0-9_.`-]+)", "\\.(\\2\\)", code_strings)
  code_strings <- gsub("(input\\[\\[\")([a-z0-9_.`-]+)(\"\\]\\])", "\\.(\\2\\)", code_strings)
  code_strings <- gsub("\\.\\(([^)`]*?)-([^)`]*?)\\)", ".(`\\1-\\2`)", code_strings)
```
`input[["var-name"]]` are converted `.(var_name)`,  addtional regaular expression add backticks around variable names to to ensure they are valid for substitution in R.

R/init.R
pawelru, line 117: Update to roxygen argument description should follow this change

donyunardi:
I like the masking option, and the combination of "ddl/ddl_run" makes sense to me, so I prefer 1.1. 

Even though 1.2 and 2.2 are simpler, I think we should encourage the use of all DDL arguments correctly to effectively conceal the pre-processing aspect of DDL.

I am okay to create vignette of the simpler way to cater our more advance users.

> What about alternative solution of having special teal_module_data which can create ddl module? tm_teal_data() function will have the same arguments as current ddl and in the future it could https://github.com/insightsengineering/teal/issues/860

This is interesting. Instead of using a splash screen, we will present the DDL as a module as a new tab. How does this affect other modules though? Since ddl() creates teal_data which feeds to modules, if we serve ddl module as a new tab together with the rest of modules, how do we ensure that this module will be process first? 

I suppose we could implement a similar approach with the landing popup. We could assess the class of the module, and if it's a DDL module, we can temporarily block or disable the other modules. After the submit button is clicked and authentication is successful, we can then re-enable the other modules.

m7pr:
@donyunardi 

> This is interesting. Instead of using a splash screen, we will present the DDL as a module as a new tab. How does this affect other modules though? Since ddl() creates teal_data which feeds to modules, if we serve ddl module as a new tab together with the rest of modules, how do we ensure that this module will be process first?

> I suppose we could implement a similar approach with the landing popup. We could assess the class of the module, and if it's a DDL module, we can temporarily block or disable the other modules. After the submit button is clicked and authentication is successful, we can then re-enable the other modules.

I had the very same intuition and I follow your idea as well. Could be handled very similarly to the way we handle `landing_popup` now.

gogonzo:
I think we have all agree that this class is less favourable and `ddl_run` does too much (creates `teal_data`, eval code, mask code, assign `datanames` and `join_keys`). This is not possible to extend and leaves us developers in the situation where potential feature request will need a refactor.

m7pr:
less favourable or more favourable ;)?

gogonzo:
Closing this to focus on https://github.com/insightsengineering/teal/pull/957/files

PR #929, opened 2023-10-05
928 upload snapshot file

### LINKED ISSUE #928: Snapshot Manager: Upload snapshot file
### Summary
With the latest release, users are able to take a snapshot of a teal's app filter from the ~~filter~~ snapshot manager and download the JSON file. 
The next step is to allow users to upload the JSON file to the teal app.

### Acceptance Criteria
* User will be able to upload ~~the filter manager JSON~~ a snapshot file.

chlebowa:
Closes #928 

Added possibility to upload a snapshot file to the snapshot manager module.
The snapshot manager control bar (top row in table) receives a third button. Clicking it opens a modal dialog that allows for selecting a `.json` file to upload and naming the snapshot.
The new snapshot is added to the snapshot list and the filter manager dialog is reopened so that the user may immediately apply the new snapshot.

To ensure that the uploaded snapshot matches the app, `teal_slices` class receives the `app_id` attribute, which can be set with the `app_id` argument to `teal::teal_slices`. The attribute will be set by `init` to stamp a `teal_slices`. The attribute will be stored along with other attributes. Upon uploading a snapshot, the app id of the upload will be compared to that of the app.

`slices_restore` had to be re-defined in the `teal` namespace because it calls `teal_slices` and that must be the one from `teal`.

#### TESTING
```
options(teal.log_level = "WARN", teal.show_js_log = FALSE)

library(teal.modules.general)
pkgload::load_all("../teal")

rm(list = ls())

funny_module <- function (label = "Filter states", datanames = "all") {
  module(
    label = label,
    datanames = datanames,
    ui = function(id, ...) {
      ns <- NS(id)
      div(
        h2("The following filter calls are generated:"),
        verbatimTextOutput(ns("filter_states")),
        verbatimTextOutput(ns("filter_calls")),
        actionButton(ns("reset"), "reset_to_default")
      )
    },
    server = function(input, output, session, data, filter_panel_api) {
      checkmate::assert_class(data, "tdata")
      observeEvent(input$reset, set_filter_state(filter_panel_api, default_filters))
      output$filter_states <-  renderPrint({
        logger::log_trace("rendering text1")
        filter_panel_api %>% get_filter_state()
      })
      output$filter_calls <- renderText({
        logger::log_trace("rendering text2")
        attr(data, "code")()
      })
    }
  )
}

dead_module <- function(label = "empty module", datanames = NULL) {
  module(label = label,
         datanames = datanames,
         ui = function(id) {
           ns <- NS(id)
           tagList(
             h4("this is just text")
           )
         },
         server = function(id, filter_panel_api, reporter) {
           message("hello")
         }
  )
}

default_filters <- teal::teal_slices(
  teal_slice("iris", "Sepal.Length"),
  teal_slice("iris", "Sepal.Width"),
  teal_slice("iris", "Species", fixed = TRUE),
  teal_slice("mtcars", "mpg"),
  exclude_varnames = list(
    iris = c("Petal.Length"),
    mtcars = c("qsec", "drat")
  ),
  module_specific = TRUE,
  mapping = list(
    table = "iris Species",
    funny1 = c("iris Sepal.Length", "iris Sepal.Width", "iris Species"),
    funny2 = "iris Species",
    global_filters = "mtcars mpg"
  )
)

app <- init(
  data = teal_data(
    dataset("iris", iris),
    dataset("mtcars", mtcars)
  ),
  modules = modules(
    tm_data_table(
      "table",
      variables_selected = list(),
      dt_args = list()
    ),
    modules(
      label = "tab1",
      funny_module("funny1", datanames = NULL),
      funny_module("funny2", datanames = "iris"),
      dead_module("empty", datanames = NULL)
    )
  ),
  filter = default_filters
)

shinyApp(app$ui, app$server, options = list(launch.browser = TRUE))
```


chlebowa:
I have two questions:
1. Should the user be able to upload several files at once? Currently they are not. Naming multiple snapshots can be bothersome and I think naming is important.
2. Should the newly loaded snapshot be automatically applied? Currently it is not.

Note that (1) and (2) are mutually exclusive.

donyunardi:
Thanks for working on this so quickly, Aleks.

> 1. Should the user be able to upload several files at once? Currently they are not. Naming multiple snapshots can be bothersome and I think naming is important.

I agree that naming is important. Let's stick with uploading one file at a time for now and show this to the user. If a valid business case arises for loading multiple files, then we can consider it.


> 2. Should the newly loaded snapshot be automatically applied? Currently it is not.

I did expect the newly loaded snapshot to be automatically uploaded, especially since we're uploading one file at a time. Is it a simple fix to enable this? If so, I would prefer/vote for the snapshot to be automatically uploaded.

donyunardi (changes requested):
Code looks good to me.
Maybe adding NEWS.md and test cases?

R/teal_slices.R
```diff
@@ -132,6 +132,21 @@ c.teal_slices <- function(...) {
 }
 
 
+# this function must be defined in the `teal` namesspace so that it calls `teal::teal_slices`
+#' @noRd
+#' @keywords internal
+#'
+slices_restore <- function(file) {
```
m7pr, line 139: @chlebowa should we remove from teal.slice?

chlebowa: I believe `teal.slice` should contain everything that can be done with a `teal_slice` or a `teal_slices`. The fact that we need a copy in `teal` is an unfortunate consequence of the `teal` extension of the `teal_slices` class.

m7pr: ok, good to know. slices_restore will be extended by https://github.com/insightsengineering/teal.slice/pull/432 so it's good to keep that in mind, which I know you had!

chlebowa: Yes, I will copy the changes as soon as they are merged into `teal.slice`.

chlebowa:
@donyunardi 
I'm not sure I can write tests for this.
NEWS is amended.

What is missing is ensuring that the uploaded `teal_slices` can be applied to the app where it was uploaded. We need an app ID of a sort.

chlebowa:
One way to provide an app id (from @gogonzo) is to have `init` hash all of its arguments.

I imagine something like 
```
rlang::hash(c(mget(names(formals(init))), match.call()))
```
The hash could be added to the `teal_slices` coming from the `filter` argument as an additional attribute. `teal::teal_slices` would then get an optional argument like `app_id` and that way the hash would be able to be stored and restored in the `teal_slices`.

There could be trouble, e.g. due to the `teal_data` refactor the same data will be passed in a different form, likely resulting in a different hash.

We don't even have to go that far, that line up there gives me a different string every time I run the same example app. I guess it might be caused by the fact that we use environments and they are never identical.

chlebowa:
As I suspected, the different hash results from the fact that `data` and `filter` arguments in `init` effectively take environments (`TealDataset` and `teal_slices`, respectively`).
Here is how to obtain a reproducible hash: 
```
hashables <- c(mget(names(formals(init))), call = match.call())
hashables$data <- lapply(hashables$data$get_datanames(), function(ds) hashables$data$get_dataset(ds)$get_raw_data())
hashables$filter <- as.list(hashables$filter, recursive = TRUE)
rlang::hash(hashables)
```

R/module_snapshot_manager.R
```diff
@@ -144,7 +155,63 @@ snapshot_manager_srv <- function(id, slices_global, mapping_matrix, filtered_dat
...
+        )
+      )
+    })
+    # Store new snapshot to list and restore filter states.
+    observeEvent(input$snaphot_file_accept, {
+      snapshot_name <- trimws(input$snapshot_name)
+      if (identical(snapshot_name, "")) {
+        snapshot_name <- tools::file_path_sans_ext(input$snapshot_file$name)
+      }
+      if (is.element(make.names(snapshot_name), make.names(names(snapshot_history())))) {
```
kartikeyakirar, line 183:
```suggestion
      if (tolower(tools::file_ext(input$snapshot_file$name)) != "json") {
    showNotification(
      "Please select a JSON file.",
      type = "error"
    )
  } else if (is.element(make.names(snapshot_name), make.names(names(snapshot_history())))) {
```
if user add any- other format it crash the application.

kartikeyakirar:
I experimented with a JSON file, making manual changes such as modifying and deleting fields or incorrect field name. When an invalid JSON file is uploaded, it causes the application to crash. Is it feasible to implement a mechanism to verify the presence of required fields and, if they are missing, display a notification indicating what's wrong with the JSON file?

R/module_snapshot_manager.R
vedhav: Interesting. I thought limiting the fileInput to accept only JSON files should not allow the fileInput UI to do this.
What file did you upload?

chlebowa: This is covered by `fileInput(ns("snapshot_file"), "Choose snapshot file", accept = ".json", width = "100%")`

vedhav: exactly

chlebowa:
> I experimented with a JSON file, making manual changes such as modifying and deleting fields or incorrect field name. When an invalid JSON file is uploaded, it causes the application to crash. Is it feasible to implement a mechanism to verify the presence of required fields and, if they are missing, display a notification indicating what's wrong with the JSON file?

Please be more specific.
Any errors raised in such cases probably arise in `slices_restore`. If anything, we should add guards there. @m7pr

R/module_snapshot_manager.R
kartikeyakirar: Indeed, the system accepts JSON as the designated format, but users can still switch to a different format while uploading file. HEre I changed it to all files and uploaded Zip. 
image
image

vedhav: imo I think that would be a feature request for the `shiny::fileInput` and out of scope for us.

chlebowa: Oh, by the way, the app crashes with 
```
Warning: Error in slices_restore: Assertion on 'file' failed: File extension must be in {'json'} (case insensitive), but file name is '/tmp/RtmpmTbIfP/e9f1938f5618b837dde5618a/0.txt'.
```

kartikeyakirar: In my opinion, we should incorporate an additional failsafe for users instead of allowing the app to crash. The above error is not visible to the user anyway.

vedhav: I agree that the app should not crash. I would go with a general tryCatch and show a meaningful error message (like "Error while using this file. Please make sure you upload a ....") that the user can act upon rather than us fixing for every user-based edge case that is simply outside the scope of "our" package.
This is the same case as using external API. fileUploads are beyond our control and tryCatch makes sense here.

kartikeyakirar:
> Please be more specific.
> Any errors raised in such cases probably arise in slices_restore.

My point was to ensure that no user action would cause the application to crash, even if they made manual changes or accidentally deleted a field.  and Yes all the errors seem to originate from the `slices_restore` function. 

some of the scenarios I checked was : passing an empty JSON file and changing the JSON field names. These actions led to application crashes.

image

image

chlebowa:
> some of the scenarios I checked was : passing an empty JSON file and changing the JSON field names. These actions led to application crashes.

This is what I was asking about :+1:

donyunardi (approved):
It felt solid to me. 👍🏼

kartikeyakirar (approved):
I've tested the JSON from sample app to teal.gallery app, and the inclusion of hashes is functioning well. The pull request provides excellent functionality. 👍 💯

I

image

vedhav:
Can someone also review the [related PR #930](https://github.com/insightsengineering/teal/pull/930)

chlebowa:
Thanks to @m7pr for testing with a file sent over email. I believe the requested functionality is satisfied.

R/teal_slices.R
averissimo: Can we re-export the `teal.slices::slices_restore`?  I'm sure there's a reason, wondering what it is :grin: 

For example as `{dplyr}` [does](https://github.com/tidyverse/dplyr/blob/73742713bc61d4ef079e72b806143b23cb7c4871/R/reexport-tibble.R#L1C1-L3C19)?

averissimo: `{teal.slices}`

```R
slices_restore <- function(file, teal_slices_fun = teal_slices) {
  ....
}
```

`{teal}`

```R
slices_restore <- function(file) {
  teal.slices(file, teal_slices)
}
```

chlebowa: hmmm :thinking:

chlebowa: As neat as it is, it does not resolve the underlying problem (which I had missed). Consider this:
There are two versions of `teal_slices` and two respective versions `slices_restore`. Start a fresh session.

A `teal_slices` object created under `teal.slice` can be restored by both `restore` funcitons.
```
x <- teal.slice::teal_slices(
  teal.slice::teal_slice("iris", "Species")
)
teal.slice::slices_store(x, "testfile.json")
y <- teal.slice::slices_restore("testfile.json")
z <- teal:::slices_restore("testfile.json")
```
However, a similar object created under `teal` can only be loaded by the function from `teal`.
```
X <- teal::teal_slices(
  teal.slice::teal_slice("iris", "Species")
)
teal.slice::slices_store(X, "testfile.json")
Y <- teal.slice::slices_restore("testfile.json")   # FAILS
Z <- teal:::slices_restore("testfile.json")
```

After some discussion with @gogonzo and @m7pr we decided to move `slices_store` and `slices_restore` to `teal`. Only then will there be no opportunity for error.

Thank you for raising this @averissimo :+1:

chlebowa:
This PR is now **blocked** until 1) the change to `slices_restore` [is resolved](https://github.com/insightsengineering/teal.slice/pull/432) and 2) `slices_store` and `slices_restore` are moved to `teal` in a separate PR.

R/init.R
```diff
@@ -164,6 +164,13 @@ init <- function(data,
   # convert teal.slice::teal_slices to teal::teal_slices
   filter <- as.teal_slices(as.list(filter))
 
+  # Calculate app hash to ensure snapshot compatibility. See ?snapshot. Raw data must be extracted from environments.
+  hashables <- mget(c("data", "modules"))
+  hashables$data <- lapply(hashables$data$get_datanames(), function(dn) {
+    try(hashables$data$get_dataset(dn)$get_raw_data(), silent = TRUE)
+  }) # `try` handles errors from remote data
+  attr(filter, "app_id") <- rlang::hash(hashables)
```
gogonzo, line 172: There is no raw data in case of connector or DDL. This hash could be calculated after pull (teal_with_splash_srv). If it stays here it could be calculated just on datanames.

chlebowa: You're right. But aren't we due to remove `teal_with_splash`?

gogonzo: (future) Removal or keeping the `teal_with_splash` shouldn't influence this solution. DDL can't provide this information in the `teal::init` as  shiny session hasn't started yet

chlebowa: That's true. I will move it but note that this makes the process a whole lot less clear.

chlebowa: So in `teal_with_splash` the data is still not pulled and only a temporary `NULL` is present. I'm not sure how deep this will have to go.

chlebowa: Oh, it's even worse. Down in `module_teal` `raw_data()`  is NULL, there aren't even dataset names.

gogonzo: Yup, I just noticed it on my refactor branch. This makes sense because data is returned when submit button is clicked, otherwise it is NULL. I think in case of DDL hash could be built on the reproducible code instead?

chlebowa: [How's this?](https://github.com/insightsengineering/teal/pull/929/commits/d0e86d7973821473dcc6c308d189cbbad1d8cfd2)?

gogonzo: better ;)

PR #930, opened 2023-10-06
Improve the modal UX for the snapshot module

vedhav:
### Problem:
It takes a small amount of time to close the modal and reopen it causing a bad UX, it would be nice if we could just overlay the secondary modal popups over the primary one creating a more intuitive UX for the user.

For testing I'm using the same app code mentioned in the PR #929 

### Changes:

1. Introduction of `custom_modal_dialog()` to create a custom modal dialog that can be placed along with other modal dialogs without issue, allowing these "custom modal dialogs" to be easily shown and hidden using JS.
2. Fix the namespace issue because of duplicate HTML ids for `snapshot_name` causing the `snapshot_name` in the file upload dialog box to not have any value (now we have ids `new_snapshot_name` and `new_file_snapshot_name`).

Please have a look at the [GIF](https://github.com/insightsengineering/teal/assets/49812166/4c455527-75ff-4130-8961-0779e9a38647) to see the new change. (~7MB might take a while to load)
Kapture 2023-10-06 at 16 16 12


chlebowa:
Thanks for this, I'll have a look.

As a general comment, could you please not upload gifs? Their constant animation is distracting, especially not as fluid as this one. Screencasts are much better, you play them when you want.

m7pr:
What about gif+screencast?

chlebowa:
I'm not sure I like invisible inputs. But then that is what `conditinalPanel` does and we have a lot of those in the modules.

What do you think @insightsengineering/nest-core-dev?

vedhav:
I'm not sure what you mean by "invisible inputs" but if you're referring to the bottom right `Computing ...` which is the only `conditionalPanel` UI I can see over here then I totally agree and I'm not a fan of them. I would love to have component-specific loading with something like the waiter package if I had to choose.

chlebowa:
I mean input widgets that have `display: none`.
And by `conditinalPanel` I mean `shiny::conditionalPanel` in several module UI functions.

kartikeyakirar:
Should we make this as widget in teal.widget ?

vedhav:
> Should we make this as widget in teal.widget ?

If you're proposing just the `custom_modal_dialog` then it makes perfect sense to me. But nothing more than that because in my head `teal.widgets` help teal module developers create UI that is very common to be reused in many modules. And the only reusable component I see from here would be `custom_modal_dialog`, We'd have to come up with a better name.
But I stand against moving the snapshots module into teal.widgets.

chlebowa:
While I like having them side by side, we shouldn't get too attached to that layout because in apps with many modules the mapping matrix will push the snapshots down anyway. [See here](https://github.com/insightsengineering/teal.slice/issues/394).

chlebowa:
> I'm not sure I like invisible inputs. But then that is what `conditinalPanel` does and we have a lot of those in the modules.
> 
> What do you think @insightsengineering/nest-core-dev?

I would really like to hear some opinions on this.

kartikeyakirar:
> with many modules the mapping matrix will push the snapshots down anyway

Ah, I see. In that case, I suppose this is bound to happen anyway. From a user's perspective, having a visible snapshot manager could make it more accessible. should we consider adding tabs in modal ?

vedhav:
> I'm not sure I like invisible inputs. But then that is what conditinalPanel does and we have a lot of those in the modules.

I don't like when the hidden elements persist to stay in the DOM forever. But, I'm not bothered by the hidden elements if they are removed when they are not needed which is the case with this. The nested popup is part of the main popup and will be completely removed once the main popup is removed.

vedhav:
> should we consider adding tabs in modal ?

How about moving it to the top? and having a scroll to the table as suggested [here](https://github.com/insightsengineering/teal.slice/issues/394#issuecomment-1721160586)

kartikeyakirar:
> I'm not sure I like invisible inputs. But then that is what conditinalPanel does and we have a lot of those in the modules.

Invisible inputs may have their advantages in specific scenarios, but personally, share both yours and vedha opinion on this. This is primarily due to the added complexity they introduce when managing visibility. Even when hidden, these inputs persist in the DOM, consuming resources and potentially causing latency in your application.

kartikeyakirar:
@vedhav moving the snapshot manager table  won't resolve this because a similar edge case exists there too. Users can encounter the same problem when adding multiple snapshots and pushing the filter table down.

chlebowa:
This discussion is branching into two. Let's keep the layout part in the relevant issue, shall we?

chlebowa:
While I like the UX improvement, I feel DX is severely deteriorated. Modals are created in the ui function and their handling (showing, hiding, updating) is done in the server. It could be just me because I like having everything in one place but it does feel disjointed.

Also, there is no `showModal` and `removeModal` so it still seems to me that the modals are not created and destroyed in response to buttons but rather persist in an hidden state. They may be removed when the parent modal is closed but that is only because the snapshot manager placed in a modal (which is created by a different module), what happens if we decide to take it out at some point?

vedhav:
> what happens if we decide to take it out at some point?

You're right, Removing the modal is an effect of being inside the modal which is the intended use of this "nested" modal. It's supposed to be nested inside a modal.

> Modals are created in the ui function and their handling (showing, hiding, updating) is done in the server

I would disagree that the modals are created in the UI. Shiny still creates them on the server side, so if you're disconnected from the server even the showModal/hideModal does not work, only the easyClose by clicking outside or pressing escape works because they are implemented on the JS side.

chlebowa:
> I would disagree that the modals are created in the UI.

OK, _defined_.

PR #946, opened 2023-10-26
test for teal_data_to_filtered_data

m7pr:
CC @gogonzo just wondering what should be the input for `teal_data_to_filtered_data` in this test


chlebowa:
The code is fine but I think it needs a 
```
testthat::expect_true(inherits(teal:::teal_data_to_filtered_data(data), "FilteredData"))
```
or something.

m7pr:
I get 

```r
Error in (function (dataset_args, dataname) :
Assertion on 'dataset_args' failed: Must be of type 'list', not 'data.frame'.
```

on my side. Will double check if my branches are accurate.

m7pr:
Sorry guys. Ignore me. I had wrong version of teal.code

PR #947, opened 2023-10-27
Update modules.R

### LINKED ISSUE #942: printing `teal_module(s)` objects
Printing `teal_module` and `teal_modules` is implemented by `print` calling `cat(toString(x))`. This is unusual, typically it would be done with `cat(format(x))`. Perhaps we should change it.

ryanyu9:
teal_module print

# Pull Request

Fixes #942


donyunardi:
Thanks for suggesting this. This was completed in 
https://github.com/insightsengineering/teal/pull/1054/files

PR #956, opened 2023-11-01
replace ddl run() with eval_and_mask()

kartikeyakirar:
Replacing ddl_run with eval_and_mask() in function and documentation.


gogonzo:
Not actual anymore. Thanks for suggestion

PR #957, opened 2023-11-01
Introduce the new DDL

gogonzo:

Another variation for `ddl`. Key changes:
- `teal::init` additionally accepts `list(ui = , server = )`. Both only with `id` argument.
- ~~`delayed_data()` function creates `list(ui=, server=)` and allows to call `server` function with additional arguments. `delayed_data` accepts additional arguments in `...` and they are used in `server()` evaluation. These `...` must be named and matching `server` formals.~~
- We expose `eval_and_mask(code, input, input_mask)` function to support evaluation-with-masking. We need to think about naming of a function and its arguments.
- Since `eval_and_mask` is used in the similar way as `eval_code` and `within` I decided to not use `input$xyz` for masking but `xyz` instead. Please share your feelings about this.
 
<details>
<summary>App code </summary>

```r
library(scda)
pkgload::load_all("teal")

# initial data
data <- teal_data() |> within({
  iris <- iris
  mtcars <- mtcars
})

code <- quote({
  cat("\nusername: ", username, "\npassword: ", password) # mock connection function
})

env_mask <- list(
)

ddl_data <- list(
  ui = function(id) {
    ns <- NS(id)
    tagList(
    )
  },
  server = function(id) {
      eventReactive(input$submit, {
        masked_data <- data |> eval_and_mask(code = code, env = reactiveValuesToList(input), env_mask = env_mask)
        # iris, mtcars set in preprocessing; ADSL and ADTTE during teal session
        teal.data::datanames(masked_data) <- c("iris", "mtcars", "ADSL", "ADTTE")
        masked_data
      })
    })
  }
)

app <- teal::init(
  data = ddl_data,
  modules = list(
    teal.modules.general::tm_data_table("Data Table")
  )
)

runApp(app)


```

</details>


vedhav:
I like the decoupled version of DDL.
In the server module you meant `data` instead of `teal_data()` right? Over here 👇🏽 
```r
masked_data <- teal_data() |> eval_and_mask(code = code, input = input, input_mask = input_mask)
```

R/data-transform_module.R
```diff
@@ -0,0 +1,23 @@
+#' Transform module for `teal_data`
+#'
+#' Function creates object of class `teal_trnasform_module` which allows
```
m7pr, line 3:
```suggestion
#' Function creates object of class `teal_transform_module` which allows
```

R/data-transform_module.R
```diff
@@ -0,0 +1,23 @@
...
+#' `teal` app developer to transform freely `teal_data` object passed to `data` argument in
```
m7pr, line 4:
```suggestion
#' `teal` app developer to transform  `teal_data` object passed to `data` argument in
```

m7pr:
I don't prefer this approach over https://github.com/insightsengineering/teal/pull/955 because here we introduce a `function` that has a confusing name with `teal.transform` package. Maybe `teal_data_tranfsorm` so that it is clear it is meant for `teal_data` transformations, which would be an extension of `teal_data` object. I don't like the way `ui` and `server` are split and created separately. When they were created together in one module, like in https://github.com/insightsengineering/teal/pull/955, it was clearer to follow what is their purpouse.

But as @pawelru mentioned in here https://github.com/insightsengineering/teal/pull/955#issuecomment-1790401336
>  (transform module) should be an extension

we can have https://github.com/insightsengineering/teal/pull/955 as basic specification of `teal_data()` module and a `teal_data_transform` module for any extra customized adjustments.

gogonzo:
Example of the usage. I was able to create exactly the same constructor as in the [**solution 2**](https://github.com/insightsengineering/teal/pull/955) based on this design. It proves that on the lowest level we just need a `ui` and `server` and additional args to be passed when `server` is evaluated.
This means that potential `ddl` class is very similar to `teal_module` with some distinctions:
- `ddl` doesn't need `datanames`
- `ddl` doesn't need `label`?
 

```r
obj <- ddl(
  code = quote({
    open_conn(username = username, password = password)
  }),
  input_mask = list(
  ),
  ui = function(id) {
    ns <- NS(id)
    tagList(
    )
  },
  server = function(id, code, input_mask) {
      eventReactive(input$submit, {
        masked_data <- teal_data() |> eval_and_mask(code = code, input = input, input_mask = input_mask)
      })
    })
  }
)


app <- init(
  data = obj,
  modules = modules(
    teal.modules.general::tm_data_table(label = "yolo")
  ),
  filter = teal_slices(
    teal_slice(dataname = "ADSL", varname = "AGE", selected = c(18, 60)),
    teal_slice(dataname = "inexisting", varname = "AGE")
  )
)

runApp(app)
```

vedhav:
Is it possible to have an implementation where we do not pass a `teal_data` object with it's own code and additional masked code for DDL? I think it would be great if the DDL code is passed as part of the teal_data object.

 <details>
<summary>code passed along with `teal_data`</summary>

```r
ui <- function(id) {
  ns <- NS(id)
  tagList(
  )
}

server <- function(id, data) {
  input_mask <- list(
  )

    eventReactive(input$submit, {
      masked_data <- data |> eval_and_mask(input = input, input_mask = input_mask)
    })
  })
}

app <- init(
  data = delayed_data(
    ui = ui,
    server = server,
    data = teal_data(
      iris = iris,
      code = quote({
        iris <- iris
        open_conn(username = username, password = password)
      })
    )
  ),
  modules = modules(
    teal.modules.general::tm_data_table(label = "yolo")
  ),
  filter = teal_slices(
    teal_slice(dataname = "ADSL", varname = "AGE", selected = c(18, 60)),
    teal_slice(dataname = "inexisting", varname = "AGE")
  )
)

runApp(app)
```
</details>

gogonzo:
> Is it possible to have an implementation where we do not pass a teal_data object with it's own code and additional masked code for DDL?

Now, yes, just specify `ui` and `server` (only with `id`), and don't pass any data.

<details>
<summary> code </summary> 

```r
options(teal.log_level = "TRACE", teal.show_js_log = TRUE)
library(scda)
pkgload::load_all("teal")

ui <- function(id) {
  ns <- NS(id)
  tagList(
  )
}

server <- function(id) {
  code <- quote({
    open_conn(username = username, password = password)
  })

  input_mask <- list(
  )

    eventReactive(input$submit, {
      masked_data <- teal_data() |> eval_and_mask(code = code, input = input, input_mask = input_mask)
    })
  })
}

app <- init(
  data = delayed_data(ui, server),
  modules = modules(
    teal.modules.general::tm_data_table(label = "yolo")
  ),
  filter = teal_slices(
    teal_slice(dataname = "ADSL", varname = "AGE", selected = c(18, 60)),
    teal_slice(dataname = "inexisting", varname = "AGE")
  )
)

runApp(app)


```

</details>

R/data-data-utils.R
```diff
@@ -0,0 +1,98 @@
...
+#'
+#' @return `teal_data` object
+#'
+#' @export
+eval_and_mask <- function(data,
+                          code,
+                          input = list(),
+                          input_mask = list()) {
+  # todo: do we need also within_and_mask?
+  checkmate::assert_list(input)
```
kartikeyakirar, line 15:
```suggestion
  checkmate::assert_class(data, "teal_data")
  checkmate::assert_list(input)
```

gogonzo:
@pawelru 

Here is basic example with app using `data = list(ui = ui, server = server)`

<details>
<summary> app code</summary>

```r
options(teal.log_level = "TRACE", teal.show_js_log = TRUE)
library(scda)
pkgload::load_all("teal")

ui <- function(id) {
  ns <- NS(id)
  tagList(
  )
}

server <- function(id) {
  code <- quote({
    open_conn(username = username, password = password)
  })

  input_mask <- list(
  )

    eventReactive(input$submit, {
      masked_data <- teal_data() |> eval_and_mask(code = code, input = input, input_mask = input_mask)
    })
  })
}

app <- init(
  data = list(ui = ui, server = server),
  modules = modules(
    teal.modules.general::tm_data_table(label = "yolo")
  ),
  filter = teal_slices(
    teal_slice(dataname = "ADSL", varname = "AGE", selected = c(18, 60)),
    teal_slice(dataname = "inexisting", varname = "AGE")
  )
)

runApp(app)

```


</details>

gogonzo:
I think there is an agreement that `teal::init(data =)` should accept also `list(ui = , server = )`. Everything on top of `list(ui, server)` is something extra, but assertion is still valid.

I consider few things to decide on:

1. If needed one can open `server` function to accept any object `data = delayed_data(ui, server, x, y, z)` which can be passed by app developer.   
2. Should `ddl(code, input_mask, ui, server)` be an exported function or should we just show this case in the documentation? Won't work without (1).
3. we need to make a decision about function/class names (if any).

R/data-ddl-utils.R
```diff
@@ -0,0 +1,99 @@
...
+        list(expr = code)
+      ),
+      env = args
+    )
+  )
+}
+
+#' Convenience wrapper for ddl
+#' @export # todo: do we want to export this?
+ddl <- function(code, input_mask, ui, server) {
```
gogonzo, line 72: Could be used in this way 

<details>
<summary> app code </summary>
 
```r
options(teal.log_level = "TRACE", teal.show_js_log = TRUE)
library(scda)
pkgload::load_all("teal")

ui <- function(id) {
  ns <- NS(id)
  tagList(
  )
}

server <- function(id, code, input_mask) {
    eventReactive(input$submit, {
      masked_data <- teal_data() |> eval_and_mask(code = code, input = input, input_mask = input_mask)
    })
  })
}

data <- ddl(
  code = quote({
    open_conn(username = username, password = password)
  }),
  input_mask = list(
  ),
  ui = ui,
  server = server
)

app <- init(
  data = data,
  modules = modules(
    teal.modules.general::tm_data_table(label = "yolo")
  ),
  filter = teal_slices(
    teal_slice(dataname = "ADSL", varname = "AGE", selected = c(18, 60)),
    teal_slice(dataname = "inexisting", varname = "AGE")
  )
)

runApp(app)
```

</details>

kartikeyakirar: Question: Do we actually require the function ddl()? It only utilize `delayed_data` with an additional argument . Perhaps it would be more suitable as a additional example in documentation or  in the vignette?

R/data-ddl-utils.R
```diff
@@ -0,0 +1,99 @@
+#' Function runs the `code`, masks the `code` and creates `teal_data` object.
+#' @param data (`teal_data`) object
+#' @param code (`language`) code to evaluate
+#' @param input (`list`) containing inputs to be used in the `code`
+#' @param input_mask (`list`) containing inputs to be masked in the `code`
+#'
+#' @return `teal_data` object
+#'
+#' @export
+eval_and_mask <- function(data,
```
pawelru, line 10: this should belong to `teal.code` as a separate method or wrapper

R/data-ddl-utils.R
```diff
@@ -0,0 +1,99 @@
...
+                          code,
+                          input = list(),
+                          input_mask = list()) {
+  # todo: do we need also within_and_mask?
+  checkmate::assert_list(input)
+  if (inherits(input, "reactivevalues")) {
+    input <- shiny::reactiveValuesToList(input)
+  }
```
pawelru, line 18: you can do `shiny::reactiveValuesToList` on function call and expect a list - will be simpler

R/data-ddl-utils.R
gogonzo: I consider `eval_and_mask` to be exception which should be avoided in standard workflow. I thought it would be better if we demonstrate `eval_and_mask` only in case of `ddl` because in contrary it might be used in other cases to break reproducibility.

R/data-ddl-utils.R
gogonzo: Exactly, I think the same

R/data-ddl-utils.R
pawelru: How this can be avoided for connectors with auth?

R/data-ddl-utils.R
kartikeyakirar: > We can export delayed_data (name to decide) and expose "delayed data loading" in the documentation as an extension of delayed_data

my preference would be exporting `delayed_data`  as it might be more helpful to understand the class and concept itself to user.

kartikeyakirar: > Not sure why I can not see my ques above so pasting  here again,

> Question: Do we require the function ddl()? It only utilize `delayed_data` with an additional argument . Perhaps it would be more suitable as additonal example in documentation or  in the vignette?

R/data-ddl-utils.R
```diff
@@ -0,0 +1,99 @@
...
+
+  if (!missing(input_mask)) {
+    # mask dynamic inputs with mask
+    input <- utils::modifyList(input, input_mask)
+
+    # replace last code entry with masked code
+    # format_expression needed to convert expression into character(1)
+    #  question: warnings and errors are not masked, is it ok?
+    data@code[length(data@code)] <- format_expression(.substitute_code(code, args = input))
```
pawelru, line 39: Yet another reason to move it to teal.code is that here you are coupling here to the class implementation (internal to teal.code) and not using public package API.

R/data-ddl-utils.R
gogonzo: > my preference would be exporting delayed_data as it might be more helpful to understand the class and concept itself to user.

I have the same opinion. Besides, `delayed_data` might have much wider application

R/data-ddl-utils.R
gogonzo: Connectors will use `eval_and_mask`, but it will be developer by us, so we know it will be used in a right context.

R/data-ddl-utils.R
gogonzo: Sounds good tbh

R/data-ddl-utils.R
```diff
@@ -0,0 +1,99 @@
...
+#' @param input (`list`) containing inputs to be used in the `code`
+#' @param input_mask (`list`) containing inputs to be masked in the `code`
+#'
+#' @return `teal_data` object
+#'
+#' @export
+eval_and_mask <- function(data,
+                          code,
+                          input = list(),
+                          input_mask = list()) {
```
pawelru, line 13: rename needed? Proposition: `env` and `env_mask`
This is so specific to the `shiny` and if you do https://github.com/insightsengineering/teal/pull/957/files#r1381599897 then there will be no link to the shiny at all. I feel this functionality is of general use.

gogonzo: Good proposition, `env` and `env_mask` make sense as it is related directly to `substitute`.

R/data-transform_module.R
```diff
@@ -0,0 +1,46 @@
...
+
+  extra_formals <- setdiff(server_formals, c("id", names(server_args)))
+  if (length(extra_formals) > 0) {
+    stop(
+      "Missing arguments specified in delayed_data(): ",
+      toString(extra_formals),
+      "\n arguments specified in `...` should be the same as in `server` function",
+      call. = FALSE
+    )
+  }
```
pawelru, line 39: why we are blocking usage of `...` in the provided `server` function?

pawelru: I don't clearly see why one would use it but on the other hand I also don't see why one can't do this

gogonzo: I will include this. I was speed-coding to give a working version to the discussion and I skipped `...` to not loose so much time then.

m7pr:
Isn't `data = list(ui = ui, server = server)` confusing in this example https://github.com/insightsengineering/teal/pull/957#issuecomment-1792237632 ? There is barely any data. Is the name `data` still relevant, if the input is an ui+server(that has the data)?

gogonzo:
> Isn't data = list(ui = ui, server = server) confusing in this example https://github.com/insightsengineering/teal/pull/957#issuecomment-1792237632 ? There is barely any data. Is the name data still relevant, if the input is an ui+server(that has the data)?

I guess it might be confusing since it is for you. I think that having `data` in the function name is relevant because `server` must return reactive `teal_data`

R/data-ddl-utils.R
m7pr: I don't think exposing makes it easier to understand as there is `server_formals`, `extra_args`, `server_args`, `extra_formals`. If app developer understand that much of R and Shiny, would he be able to craft his own tool for his needs, instead of using teal? Even though I vote to expose

gogonzo: `extra_args` and `extra_formals` are just helpful to print relevant error message. There are `server` formals and `...` which need to be matched. We need to throw meaningful message to alert that they don't match ;]

m7pr:
I guess this is a tradeoff of simplicity, functionality, flexibility and applications. The basic user might get overwhelmed, but if we do a great job with documentation, templates and vignettes this can be understandable, even though one needs to understand Shiny Modules to pass data to the tool that supposted to be simple :P

R/data-ddl-utils.R
m7pr: I think dark smog clouds has overcome my brain today :)

donyunardi:
> I see three ways out:
> 
> We stay with list(ui, server) and we export only eval_and_mask(). ddl case will be exposed only in the vignette as a special case.
> We can export delayed_data (name to decide) and expose "delayed data loading" in the documentation as an extension of delayed_data
> We can export ddl <- function(code, input_mask, ui_server) (we need delayed_data to create such function).

I don't see a significant distinction between `data = delayed_data(ui, server),` and `data = list(ui = ui, server = server),` from user's perspective. The arguments are the same, and I can't tell how one function is more beginner-friendly than the other.

IMO, I think `ddl` could be the more user-friendly approach for beginners.
I am currently thinking to expose `eval_and_mask()` for our advanced users and `ddl()` for our beginner users.

R/data-data-utils.R
```diff
@@ -0,0 +1,98 @@
+#' Function runs the `code`, masks the `code` and creates `teal_data` object.
```
kartikeyakirar:
```suggestion
#' Evaluate and mask code in the `teal_data` environment
#'
#' Function evaluates the given code in the `teal_data` environment,
#' applies input masking, and updates the `teal_data` object.
```

R/data-ddl-utils.R
```diff
@@ -0,0 +1,104 @@
+#' Function runs the `code`, masks the `code` and creates `teal_data` object.
```
kartikeyakirar, line 1:
```suggestion
#' Evaluate code and mask input in a `teal_data` object
#'
#' Function evaluates R code in the provided `teal_data` object's environment
#' and then masks specified variables.
```

chlebowa:
1. As I said previously, I do not condone putting code masking in the hands of the user. I believe it should be done behind the scenes.
2. If `ddl_run` is bad because it does more than one thing, so is `eval_and_mask`.
3. `eval_and_mask` or any equivalent should _not_ reside in `teal.code` because `teal.code` is on its way to enforce strict reproducibility and this is a mechanism to violate it. Since `teal_data` constructor also violates reproducibility, it makes sense to keep all mechanisms like that in `teal.data`. `eval_and_mask` is basically a helper function for constructing `teal_data` anyway.
4. Users should not be burdened with the knowledge _how_ our functions work. By this I mean they should understand _what_ they do but not _how_ they do it. Therefore, adjusting argument names in `eval_and_mask` to those of `substitute` is a moot point. Superficially `eval_and_mask` deals with _code_, not environments and deriving argument names from "environment" is confusing.
5. We wanted to limit/eliminate the use of `eval_code`, but `eval_and_mask` is a variant of it, one that features prominently.
6. I fail to see the benefit of allowing `list(ui, server)` in `init(data`. I think it makes more sense to allow `teal_data` and this special module (we are sort of departing from the module concept) and if the user only wants a ui and a server with nothing fancy, they can pass them to the constructor instead of `list`. :+1: @donyunardi 

Overall I have a negative opinion on this solution. I do have some constructive criticism as well but it needs refinement.

pawelru:
Hey @chlebowa please find below my comments to your points. Happy to meet and discuss more if needed.

> 1. As I said previously, I do not condone putting code masking in the hands of the user. I believe it should be done behind the scenes.

My intention here is to provide extreme flexibility and allow users to define their own modules with whatever logic. I do want to process new data loading related feature request by creating modules on to top layer as opposed to changing the code on the bottom one (e.g. ddl_run) (and then go through the release cycle). For some of the cases, data loading involves auth and the code needs to be obfuscated in order to present it as a text. If we allow app-developer to define app specific data module then this implies that he needs to obfuscate the code (if auth is included). Now we can debate whether we can give an utility function or make them deal with that on their own but my point here is that this is unavoidable.

> 2. If `ddl_run` is bad because it does more than one thing, so is `eval_and_mask`.

Probably yes but I cannot came up with a better alternative. 

> 3. `eval_and_mask` or any equivalent should _not_ reside in `teal.code` because `teal.code` is on its way to enforce strict reproducibility and this is a mechanism to violate it. Since `teal_data` constructor also violates reproducibility, it makes sense to keep all mechanisms like that in `teal.data`. `eval_and_mask` is basically a helper function for constructing `teal_data` anyway.

This functionality could live outside of `teal.code` as long as we are using public teal.code API. The current implementation of that function is tightly coupled with the class implementation inside teal.code package - especially nasty things like accessing and overwriting the object slots. If done using public package interface then I am perfectly fine to have it elsewhere.

> 4. Users should not be burdened with the knowledge _how_ our functions work. By this I mean they should understand _what_ they do but not _how_ they do it. Therefore, adjusting argument names in `eval_and_mask` to those of `substitute` is a moot point. Superficially `eval_and_mask` deals with _code_, not environments and deriving argument names from "environment" is confusing.

Please see environment as an unordered list. This is also easier for documenting as we can refer to the used function parameters and the function itself.

> 5. We wanted to limit/eliminate the use of `eval_code`, but `eval_and_mask` is a variant of it, one that features prominently.

That's a valid point. But I cannot find a better alternative.

> 6. I fail to see the benefit of allowing `list(ui, server)` in `init(data`. I think it makes more sense to allow `teal_data` and this special module (we are sort of departing from the module concept) and if the user only wants a ui and a server with nothing fancy, they can pass them to the constructor instead of `list`. 👍 @donyunardi

This is for backward design compatibility. In the current released version we accept both `TealDataset` and `TealDatasetConnector` that represents data.frame and module respectively. I don't agree that `data = teal_data() |> teal_data2teal_data_module()` simplifies it.
On the other thing there is something in what you have raised. That's input class inconsistency. It's always better to accept one and only one class of an argument. There is a way to address it which is to introduce a new function argument `data_module` but this opens (at least) two big issues: (i) what's the relation between `data` and `data_module` argument? (ii) isn't it too big of a change? I don't know the answers but I haven't thought about it a lot.
I personally raised this a while ago and my idea was to open a data loading app and on success take what's returned and move on to the main app. But I failed on redirect on deployed app (locally it was working). My attempts resulted in lost connection and grey-out app :(

chlebowa:
@pawelru 

Points (1), (2), (5), and (6) have been addressed and we more or less agree on the direction to take.

Ad (3) That's the thing we _don't_ want to have a public API that violates reproducibility, at least not in `teal.code`. Nevertheless, a need exists in `teal.data` and we probably will have to create a masking functionality that is public.

Ad (4) 
>> Users should not be burdened with the knowledge how our functions work. By this I mean they should understand what they do but not how they do it. Therefore, adjusting argument names in eval_and_mask to those of substitute is a moot point. Superficially eval_and_mask deals with code, not environments and deriving argument names from "environment" is confusing.

> Please see environment as an unordered list. In python terminology it's a _tuple_ instead of a _array_. This is also easier for documenting as we can refer to the used function parameters and the function itself. You can always do `list2env` or even some functions supports lists.
> UPDATE: Environments obviously are much more complex because of inheritance mechanism but I deliberately ignoring this in that context.

More like a dictionary, I think.
I have no problem understanding the concept, however, I believe that app developers shouldn't have to. They should not know, nor care, that `substitute` is called and that's where we take argument names from. The function masks code, not environments, so `code_mask` makes more sense semantically. Introducing `env` to the mix is an unnecessary distraction.

chlebowa:
On a side note, I object to the use of "obfuscate". The modifications we make to the code are not meant to hide what it does but to hide sensitive information.

R/data_module.R
```diff
@@ -0,0 +1,40 @@
...
+#'     })
+#'   }
+#' )
+#' @export
+data_module <- function(ui, server) {
+  checkmate::assert_function(ui, args = "id", nargs = 1)
+  checkmate::assert_function(server, args = "id", nargs = 1)
+  structure(
+    list(ui = ui, server = server),
+    class = "data_module"
```
pawelru, line 38: `teal_data_module`? That would fit nicely to the `teal_data` class name.

chlebowa: I would opt for `teal_module_data` to be consistent with `teal_module_landing` and `teal_module_previewer`. Unless this is not an extension of `team_module`.

pawelru: > I would opt for `teal_module_data` to be consistent with `teal_module_landing` and `teal_module_previewer`. Unless this is not an extension of `team_module`.

Disagree. This class is different to the one you mentioned. First of all it does not inherit from `teal_module`. Second of all this is not meant to be passed as `module` argument value of `init()` - it is meant to be a `data` argument value instead. IMHO we should keep them separate to avoid confusion.

gogonzo: We have some unwritten convention:
- Analytical modules have a prefix `tm_`

And we have additional two special modules which are "inheriting" from `"teal_module"`:
- "teal_module_landing" produced by `landing_popup_module()`
- "teal_module_previewer" produced by `reporter_previewer_module()`

I support `teal_data_module`, do we all agree here?

vedhav: I prefer `teal_data_module` over `teal_module_data`

chlebowa: Like I said, unless. I am pretty sure they will be confused anyway. Nomenclature throughout NEST is not entirely consistent and this difference will likely be seen as an inconsistency rather than a sign of a different entity.

I don't mind `teal_data_module` as such.

R/teal_data_module.R
chlebowa, line 1: Please describe return value.

R/init.R
```diff
@@ -142,26 +142,12 @@ init <- function(data,
...
-  filter <- as.teal_slices(as.list(filter))
-
   # Calculate app hash to ensure snapshot compatibility. See ?snapshot. Raw data must be extracted from environments.
   hashables <- mget(c("data", "modules"))
   hashables$data <- if (inherits(hashables$data, "teal_data")) {
     as.list(hashables$data@env)
-  } else if (inherits(hashables$data, "ddl")) {
-    attr(hashables$data, "code")
+  } else if (inherits(data, "teal_data_module")) {
+    # what?
```
chlebowa, line 150: body of server function?
```suggestion
    body(data$server)
```

R/module_teal_with_splash.R
```diff
@@ -22,14 +22,17 @@ ui_teal_with_splash <- function(id,
...
   ns <- NS(id)
 
   # Startup splash screen for delayed loading
   # We use delayed loading in all cases, even when the data does not need to be fetched.
   # This has the benefit that when filtering the data takes a lot of time initially, the
   # Shiny app does not time out.
-  splash_ui <- if (inherits(data, "teal_data")) {
+
+  splash_ui <- if (inherits(data, "teal_data_module")) {
+    data$ui(ns("data"))
```
chlebowa, line 34: something more specific?
```suggestion
    data$ui(ns("teal_data_module"))
```

R/module_teal_with_splash.R
```diff
@@ -56,17 +59,27 @@ ui_teal_with_splash <- function(id,
...
 
     if (getOption("teal.show_js_log", default = FALSE)) {
       shinyjs::showLog()
     }
 
     # raw_data contains teal_data object
     # either passed to teal::init or returned from ddl
-    raw_data <- if (inherits(data, "teal_data")) {
+    raw_data <- if (inherits(data, "teal_data_module")) {
+      ddl_out <- do.call(
```
chlebowa, line 74:
```suggestion
      do.call(
```

chlebowa (commented):
Looks all right.
Will read vignettes tomorrow.

R/module_teal_with_splash.R
```diff
@@ -56,17 +59,27 @@ ui_teal_with_splash <- function(id,
...
-    raw_data <- if (inherits(data, "teal_data")) {
+    raw_data <- if (inherits(data, "teal_data_module")) {
+      do.call(
+        data$server,
+        append(
+          list(id = "teal_data_module"),
+          attr(data, "server_args") # might be NULL or list() - both are fine
+        ),
+        quote = TRUE
+      )
```
chlebowa, line 81:
```suggestion
      data$server(id = "teal_data_module")
```

Server function must only take `id`, right?

R/module_teal_with_splash.R
```diff
@@ -102,8 +115,44 @@ srv_teal_with_splash <- function(id, data, modules, filter = teal_slices()) {
       raw_data
     }
 
-    res <- srv_teal(id = "teal", modules = modules, raw_data = raw_data, filter = filter)
-    logger::log_trace("srv_teal_with_splash initialized module with data { toString(get_dataname(data))}.")
+    raw_data_checked <- reactive({
+      data <- raw_data()
+      if (inherits(data, "qenv.error")) {
+        #
+        showNotification(sprintf("Error: %s", data$message), type = "error")
```
chlebowa, line 122: These messages may potentially be very long (in case of a long piped expression) and the notification box is rather small. Perhaps something generic?

R/module_teal_with_splash.R
```diff
@@ -102,8 +115,44 @@ srv_teal_with_splash <- function(id, data, modules, filter = teal_slices()) {
...
+        logger::log_error(data$message)
+        return(NULL)
+      }
+      if (!inherits(data, "teal_data")) {
+        msg <- "Error: server must return 'teal_data' object"
```
chlebowa, line 127:
```suggestion
        msg <- "Error: data module failed to return 'teal_data' object."
```

R/module_teal_with_splash.R
```diff
@@ -102,8 +115,44 @@ srv_teal_with_splash <- function(id, data, modules, filter = teal_slices()) {
...
+        showNotification(msg, type = "error")
+        logger::log_error(msg)
+        return(NULL)
+      }
+
+      is_modules_ok <- check_modules_datanames(modules, teal.data::datanames(data))
+      is_filter_ok <- check_filter_datanames(filter, teal.data::datanames(data))
+
+      if (!isTRUE(is_modules_ok)) {
+        showNotification(is_modules_ok, type = "error")
```
chlebowa, line 137: Result of `check_modules_datanames` and `check_modules_datanames` may also be too long for the notification box.

R/module_teal_with_splash.R
chlebowa, line 1: `raw_data_checked` stops the app but only displays `showNotification`, which is small and temporary. Since all cases covered by this check are fatal errors and beyond the control of the app user, perhaps `validate` would be a better choice? It could easly fit longer messages, including instructions to contact the app dev, and it would not disappear.

chlebowa (changes requested):

R/module_teal_with_splash.R
gogonzo: We can't send validate further - I'll need to find the way for this to not crash `teal` (filter panel etc.)

R/module_teal_with_splash.R
gogonzo: true, this is a remaining from the previous implementation. I'll fix this

R/utils.R
```diff
@@ -140,3 +140,51 @@ resolve_modules_datanames <- function(modules, datanames, join_keys) {
...
+  check_datanames <- unlist(recursive_check_datanames(modules, datanames))
+  if (length(check_datanames)) {
+    paste(check_datanames, collapse = "\n")
+  } else {
+    TRUE
+  }
+}
+
+
+check_filter_datanames <- function(filters, datanames) {
```
kartikeyakirar, line 170:
```suggestion
#' Check `datanames` in filters
#'
#' This function check `datanames` in filters correspond to those in `data`,
#' returning character vector with error messages or TRUE if all checks pass.
#'
#' @param filters (`teal_slices`) object
#' @param datanames (`character`) names of datasets available in the `data` object
#'
#' @return A character vector containing error messages or TRUE if validation passes.
#' @keywords internal
check_filter_datanames <- function(filters, datanames) {
```

Adding documentation text

R/module_teal_with_splash.R
gogonzo: I think i fixed all possible gaps possible in `teal_data_module`.

R/utils.R
```diff
@@ -140,3 +140,51 @@ resolve_modules_datanames <- function(modules, datanames, join_keys) {
     modules
   }
 }
+
+check_modules_datanames <- function(modules, datanames) {
```
kartikeyakirar, line 144:
```suggestion
#' Check `datanames` in modules
#' 
#' This function ensures specified `datanames` in modules match those in the data object, 
#' returning error messages or TRUE for successful validation.
#'
#' @param modules (`teal_modules`) object
#' @param datanames (`character`) names of datasets available in the `data` object
#'
#' @return A character vector containing error messages or TRUE if validation passes.
#' @keywords internal
check_modules_datanames <- function(modules, datanames) {
```
Adding documentation text.

vignettes/data-as-shiny-module.Rmd
```diff
@@ -0,0 +1,127 @@
...
+```
+
+_See `?within.qenv` for a detailed explanation of how to use the `within` method._
+
+Note that running preprocessing code in a module as opposed to the global environment will increase app loading times.
+It is recommended to keep the constant code in the global environment and to move only the dynamic parts to a data module.
+
+###### WARNING
+
+When using `teal_data_moduel` to modify a pre-existing `teal_data` object it is crucial that the server function and the data object are defined in the same environment as otherwise the server function will not be able to access the data object.
```
kartikeyakirar, line 126:
```suggestion
When using `teal_data_module` to modify a pre-existing `teal_data` object it is crucial that the server function and the data object are defined in the same environment as otherwise the server function will not be able to access the data object.
```

vignettes/data-as-shiny-module.Rmd
```diff
@@ -0,0 +1,127 @@
...
+  %\VignetteEncoding{UTF-8}
+---
+
+## Introduction
+
+Proper functioning of any `teal` application requires presence of a `teal_data` object.
+Typically, a `teal_data` object created in the global environment will be passed to the `data` argument in `init`.
+This `teal_data` object should contain all elements necessary for successful execution of the application's modules. 
+In some scenarios, however, application developers may opt to postpone some data operations until the application run time.
+This can be done by passing a special _`shiny` module_ to the `data` argument.
```
kartikeyakirar, line 19:
```suggestion
This can be done by passing a special _`shiny module`_ to the `data` argument.
```

R/module_teal_with_splash.R
```diff
@@ -95,15 +106,82 @@ srv_teal_with_splash <- function(id, data, modules, filter = teal_slices()) {
...
+        )
+      )
+
+      is_modules_ok <- check_modules_datanames(modules, teal.data::datanames(data))
+      is_filter_ok <- check_filter_datanames(filter, teal.data::datanames(data))
+
+      validate(need(isTRUE(is_modules_ok), is_modules_ok))
+
+      if (!isTRUE(is_filter_ok)) {
+        showNotification(is_filter_ok, type = "warning")
```
chlebowa, line 169: The message needs fixing (`NULL NULL`?).
image

I still think this shows up for too short a time, I can bearly read it with a single message, let alone with multiple.
image

R/dummy_functions.R
```diff
@@ -88,21 +89,31 @@ example_datasets <- function() { # nolint
 #'   shinyApp(app$ui, app$server)
 #' }
 #' @export
-example_module <- function(label = "example teal module", datanames = "all") {
```
pawelru, line 91: why?

gogonzo: Example didn't have show-r-code before so vignettes explaining how to make app reproducible with `teal_data |> within()` etc. were not very helpful. Would you hardcode `src` inside instead?

R/module_teal_with_splash.R
```diff
@@ -95,15 +106,82 @@ srv_teal_with_splash <- function(id, data, modules, filter = teal_slices()) {
...
+
+      # to handle module non-qenv errors
+      if (inherits(data, "error")) {
+        validate(
+          need(
+            FALSE,
+            paste0(
+              "Error when executing `teal_data_module`:",
+              attr(data, "condition")$message,
+              "\n Check your inputs or contact app developer if error persist"
```
chlebowa, line 147:
```suggestion
              "Error when executing `teal_data_module`:\n ",
              data$message,
              "\n Check your inputs or contact app developer if error persists."
```

R/dummy_functions.R
pawelru: Yes - I doubt this argument will be used at all. This function is for intended for internal use only.

R/module_teal_with_splash.R
```diff
@@ -95,15 +106,82 @@ srv_teal_with_splash <- function(id, data, modules, filter = teal_slices()) {
...
+
+      # to handle qenv.error
+      if (inherits(data, "qenv.error")) {
+        validate(
+          need(
+            FALSE,
+            paste(
+              "Error when executing `teal_data_module`:\n",
+              data$message,
+              "\n Check your inputs or contact app developer if error persist"
```
chlebowa, line 133:
```suggestion
              "Error when executing `teal_data_module`:\n ",
              data$message,
              "\n Check your inputs or contact app developer if error persists."
```

chlebowa (changes requested):
When running app with empty data:
```
devtools::load_all("../teal")

td <- teal_data()

app <- init(
  data = td,
  modules = example_module(label = "example", datanames = "all", src = TRUE),
  title = "break me",
  id = "mock"
)

runApp(app, launch.browser = TRUE)

Listening on http://127.0.0.1:3432
Warning: Error in teal.slice::get_filter_expr: Assertion on 'datanames' failed: Must be of type 'character', not 'NULL'.
  1: runApp
```

Perhaps add assertion on `datanames` in `data` not being `NULL`?

R/module_teal_with_splash.R
gogonzo: 5 seconds to 10 seconds?

R/utils.R
```diff
@@ -140,3 +140,72 @@ resolve_modules_datanames <- function(modules, datanames, join_keys) {
...
+  recursive_check_datanames <- function(modules, datanames) {
+    # check teal_modules against datanames
+    if (inherits(modules, "teal_modules")) {
+      sapply(modules$children, function(module) recursive_check_datanames(module, datanames = datanames))
+    } else {
+      extra_datanames <- setdiff(modules$datanames, c("all", datanames))
+      if (length(extra_datanames)) {
+        sprintf(
+          "- Module '%s' uses datanames not available in the 'data': (%s) not in (%s)",
```
chlebowa, line 163:
```suggestion
          "- Module '%s' uses datanames not available in 'data': (%s) not in (%s)",
```

R/module_teal_with_splash.R
chlebowa: Well...

image


You be the judge.

R/module_teal_with_splash.R
chlebowa: What about this?

R/module_teal_with_splash.R
gogonzo: Maybe: mistake of app developer shouldn't be show to app user at all. I'll just write "some filters were not applied because of incompatibility with data" and the details will be in logs

chlebowa: I think that might be better. BUT. What if a remote data source drops a dataset between app sessions?
App dev builds app with downloaded data and with filters for `dataset1`, `dataset2`, `dataset3`.
App works.
Database removes `dataset3` for whatever reason.
App fails.

Is that possible?

On the other hand, maybe that is something the app dev should handle.

R/module_teal_with_splash.R
gogonzo: This is outdated. Blame GH for not showing this explicitly ;)

chlebowa: It's marked as outdated but the change hasn't been made.

gogonzo: Ok, I see now, I changed line[1] and [3] line but ommited [2]. Just pushed the update

R/module_teal_with_splash.R
```diff
@@ -95,15 +106,89 @@ srv_teal_with_splash <- function(id, data, modules, filter = teal_slices()) {
...
-      if (!is.reactive(raw_data)) {
-        stop("The delayed loading module has to return a reactive object.")
-      }
       raw_data
     }
 
-    res <- srv_teal(id = "teal", modules = modules, raw_data = raw_data, filter = filter)
-    logger::log_trace("srv_teal_with_splash initialized module with data { toString(get_dataname(data))}.")
+    if (!is.reactive(raw_data)) {
+      stop("The `teal_data_module` has to return a reactive object.", call. = FALSE)
```
chlebowa, line 113:
```suggestion
      stop("`teal_data_module` must return a reactive expression containing a `teal_data` object.", call. = FALSE)
```

I don't remember if I put this in my previous comments.

tests/testthat/test-init.R
```diff
@@ -23,7 +23,7 @@ testthat::test_that("init data accepts teal_data object", {
   )
 })
 
-testthat::test_that("init data throws an error with input other than TealData, teal_data and ddl", {
+testthat::test_that("init data throws an error with input other than TealData, teal_data and list(ui, server)", {
```
chlebowa, line 26: This passes:
```
init(
  data = teal_data_module(ui = function(id) div(), server = function(id) NULL),
  modules = example_module()
)
```

but this fails:
```
init(
  data = list(ui = function(id) div(), server = function(id) NULL),
  modules = example_module()
)


Error in to_relational_data.list(data = data) :    Assertion on 'data' failed: May only contain the following types: {dataset,data.frame,MultiAssayExperiment,TealDataset,TealDatasetConnector}, but element 1 has type 'function'.
```

tests/testthat/test-init.R
```diff
@@ -140,6 +140,32 @@ testthat::test_that("init data accepts a list of TealDatasetConnector object", {
...
+    "Must have exactly 1 formal arguments"
+  )
+  testthat::expect_error(
+    init(
+      data = teal_data_module(ui = function(id) div(), server = function(id, x) NULL),
+      modules = modules(teal:::example_module())
+    ),
+    "Must have exactly 1 formal arguments"
+  )
+})
```
chlebowa, line 167: Arguments of module functions are checked by `teal_data_module`, is this not redundant? Unless there are no unit tests for `teal_data_module`...

tests/testthat/test-init.R
gogonzo: Sorry, I will not address it. `to_relational_data` will be removed and thus `teal::init` will accept only `teal_data` and `teal_data_module`. Currently you're theoretically able to pass `list(ui = function(), server = function())` and  also `list(ui = iris, server = MAE)`. This state is just temporary and will be solved soon

chlebowa: ok then

tests/testthat/test-init.R
gogonzo: Good point. I'll push a change soon

tests/testthat/test-module_nested_tabs.R
```diff
@@ -485,10 +485,10 @@ testthat::test_that("calculate_hashes takes a FilteredData and vector of datanam
...
 
   datasets <- teal.slice::init_filtered_data(
-    teal.data::cdisc_data(
-      teal.data::cdisc_dataset("ADSL", adsl),
-      teal.data::cdisc_dataset("ADAE", adae),
-      teal.data::cdisc_dataset("ADTTE", adtte)
+    list(
+      ADSL = list(dataset = head(adsl)),
+      ADAE = list(dataset = head(adae)),
+      ADTTE = list(dataset = head(adtte))
```
chlebowa, line 491: `head` is redundant, they all have 1 row

tests/testthat/test-module_teal_with_splash.R
```diff
@@ -1,27 +1,131 @@
...
+      modules = modules(example_module())
+    ),
+    expr = {
+      testthat::expect_is(raw_data, "reactive")
+      testthat::expect_identical(raw_data(), "whatever")
+    }
+  )
+})
+
+testthat::test_that("srv_teal_with_splash passes teal_data to reactiveVal", {
```
chlebowa, line 45:
```suggestion
testthat::test_that("srv_teal_with_splash passes teal_data to reactive", {
```

chlebowa (approved):
S O L I D :ok_hand: :rocket: :tada:

pawelru:
I can see that this got approval. Do you want me to go through this again? (I haven't checked recent changes)

chlebowa:
> I can see that this got approval. Do you want me to go through this again? (I haven't checked recent changes)

Won't hurt to have another look.

chlebowa:
@gogonzo 
https://github.com/insightsengineering/teal.data/pull/187 was merged so if you want to include it here, bump the version requirement for `teal.data` to `0.3.0.9009`.

R/init.R
```diff
@@ -142,26 +143,13 @@ init <- function(data,
...
-  # Calculate app hash to ensure snapshot compatibility. See ?snapshot. Raw data must be extracted from environments.
+  # Calculate app hash to ensure snapshot compatibility.
+  # See ?snapshot. Raw data must be extracted from environments.
   hashables <- mget(c("data", "modules"))
   hashables$data <- if (inherits(hashables$data, "teal_data")) {
     as.list(hashables$data@env)
-  } else if (inherits(hashables$data, "ddl")) {
-    attr(hashables$data, "code")
+  } else if (inherits(data, "teal_data_module")) {
+    body(data$server)
```
pawelru, line 152: I don't quite get the idea behind `hashables` but it looks to be an unique identifier of the input data (e.g. you do calculate hash on env with data tables in it). Here you are extracting server function body and calculate hashes on this. Please note that the same server function might return different object (vide: `runif()` but more realisticaly here: module with file upload). I don't think this is what we expect in the next step.

(The same mistake below where we call `hashables$data$get_code()`)

R/init.R
```diff
@@ -213,6 +189,26 @@ init <- function(data,
...
+      stop(is_modules_ok)
+    }
+
+    is_filter_ok <- check_filter_datanames(filter, teal.data::datanames(data))
+    if (!isTRUE(is_filter_ok)) {
+      logger::log_warn(is_filter_ok)
+      # we allow app to continue if applied filters are outside
+      # of possible data range
+    }
+  }
```
pawelru, line 210: `stop()` convert to `checkmate` to have better consistency of error handling

(also same in the block above but you haven't touched this)

R/module_nested_tabs.R
```diff
@@ -112,8 +112,7 @@ ui_nested_tabs.teal_module <- function(id, modules, datasets, depth = 0L, is_mod
   checkmate::assert_class(datasets, class = "FilteredData")
   ns <- NS(id)
 
-  args <- isolate(teal.transform::resolve_delayed(modules$ui_args, datasets))
```
pawelru, line 115: are you sure?
if yes - is this the only place where we use this function?
if yes - shall we remove it?

R/module_teal.R
```diff
@@ -180,11 +186,16 @@ srv_teal <- function(id, modules, raw_data, filter = teal_slices()) {
         } else if (isTRUE(attr(filter, "module_specific"))) {
           # we should create FilteredData even if modules$datanames is null
           # null controls a display of filter panel but data should be still passed
-          datanames <- if (is.null(modules$datanames)) teal.data::get_dataname(raw_data()) else modules$datanames
-          # todo: subset tdata object to datanames
+          datanames <- if (is.null(modules$datanames) || modules$datanames == "all") {
+            include_parent_datanames(raw_data()@datanames, raw_data()@join_keys) # todo: use methods instead
```
pawelru, line 190: you left `"todo"` - are those methods ready now?

R/module_teal.R
```diff
@@ -199,29 +210,23 @@ srv_teal <- function(id, modules, raw_data, filter = teal_slices()) {
...
-    if (is_arg_used(modules, "reporter") && length(extract_module(modules, "teal_module_previewer")) == 0) {
-      modules <- append_module(modules, reporter_previewer_module())
-    }
-
     # Replace splash / welcome screen once data is loaded ----
     # ignoreNULL to not trigger at the beginning when data is NULL
     # just handle it once because data obtained through delayed loading should
     # usually not change afterwards
     # if restored from bookmarked state, `filter` is ignored
-    observeEvent(datasets_reactive(), ignoreNULL = TRUE, once = TRUE, {
```
pawelru, line 218: why remove `ignoreNULL` and `once` arguments?

R/module_teal_with_splash.R
```diff
@@ -56,17 +64,20 @@ ui_teal_with_splash <- function(id,
...
+    logger::log_trace("srv_teal_with_splash initializing module with data.")
 
     if (getOption("teal.show_js_log", default = FALSE)) {
       shinyjs::showLog()
     }
 
     # raw_data contains teal_data object
     # either passed to teal::init or returned from ddl
-    raw_data <- if (inherits(data, "teal_data")) {
+    raw_data <- if (inherits(data, "teal_data_module")) {
```
pawelru, line 78: rename to `teal_data_rv`

R/module_teal_with_splash.R
```diff
@@ -56,17 +64,20 @@ ui_teal_with_splash <- function(id,
...
   moduleServer(id, function(input, output, session) {
-    logger::log_trace("srv_teal_with_splash initializing module with data { toString(get_dataname(data))}.")
+    logger::log_trace("srv_teal_with_splash initializing module with data.")
 
     if (getOption("teal.show_js_log", default = FALSE)) {
       shinyjs::showLog()
     }
 
     # raw_data contains teal_data object
     # either passed to teal::init or returned from ddl
```
pawelru, line 77:
```suggestion
    # either passed to teal::init or returned from teal_data_module
```

R/module_teal_with_splash.R
```diff
@@ -95,15 +106,89 @@ srv_teal_with_splash <- function(id, data, modules, filter = teal_slices()) {
...
 
-    res <- srv_teal(id = "teal", modules = modules, raw_data = raw_data, filter = filter)
-    logger::log_trace("srv_teal_with_splash initialized module with data { toString(get_dataname(data))}.")
+    if (!is.reactive(raw_data)) {
+      stop("The `teal_data_module` must return a reactive expression containing a `teal_data` object.", call. = FALSE)
+    }
+
+    raw_data_checked <- reactive({
+      # custom module can return error
+      data <- tryCatch(raw_data(), error = function(e) e)
```
pawelru, line 118: if you don't have custom handlers at all then just `try` should be enough

R/module_teal_with_splash.R
```diff
@@ -95,15 +106,89 @@ srv_teal_with_splash <- function(id, data, modules, filter = teal_slices()) {
...
+
+      # to handle qenv.error
+      if (inherits(data, "qenv.error")) {
+        validate(
+          need(
+            FALSE,
+            paste(
+              "Error when executing `teal_data_module`:\n ",
+              data$message,
+              "\n Check your inputs or contact app developer if error persists"
```
pawelru, line 133: please assure full stop on error messages

R/module_teal_with_splash.R
```diff
@@ -95,15 +106,89 @@ srv_teal_with_splash <- function(id, data, modules, filter = teal_slices()) {
...
-        stop("The delayed loading module has to return a reactive object.")
-      }
       raw_data
     }
 
-    res <- srv_teal(id = "teal", modules = modules, raw_data = raw_data, filter = filter)
-    logger::log_trace("srv_teal_with_splash initialized module with data { toString(get_dataname(data))}.")
+    if (!is.reactive(raw_data)) {
+      stop("The `teal_data_module` must return a reactive expression containing a `teal_data` object.", call. = FALSE)
+    }
```
pawelru, line 114: The error message indicates condition on both (i) reactive and (ii) class but what you check against is only (i).

R/module_teal_with_splash.R
```diff
@@ -95,15 +106,89 @@ srv_teal_with_splash <- function(id, data, modules, filter = teal_slices()) {
...
+      }
+
+      # to handle qenv.error
+      if (inherits(data, "qenv.error")) {
+        validate(
+          need(
+            FALSE,
+            paste(
+              "Error when executing `teal_data_module`:\n ",
+              data$message,
```
pawelru, line 132: I'm not 100% sure whether error message is one string only. For most of the cases it's true but not sure if true for all possible cases. Would be safer to paste-collapse this as well.

R/module_teal_with_splash.R
```diff
@@ -95,15 +106,89 @@ srv_teal_with_splash <- function(id, data, modules, filter = teal_slices()) {
...
+
+      validate(need(teal.data::datanames(data), "Data has no datanames. Contact app developer"))
+
+
+      is_modules_ok <- check_modules_datanames(modules, teal.data::datanames(data))
+      is_filter_ok <- check_filter_datanames(filter, teal.data::datanames(data))
+
+      validate(need(isTRUE(is_modules_ok), is_modules_ok))
+
+      if (!isTRUE(is_filter_ok)) {
```
pawelru, line 171: reorder?
```suggestion
      is_modules_ok <- check_modules_datanames(modules, teal.data::datanames(data))
      validate(need(isTRUE(is_modules_ok), is_modules_ok))
      
      is_filter_ok <- check_filter_datanames(filter, teal.data::datanames(data))
      if (!isTRUE(is_filter_ok)) {
```

R/module_teal_with_splash.R
```diff
@@ -95,15 +106,89 @@ srv_teal_with_splash <- function(id, data, modules, filter = teal_slices()) {
...
+        showNotification(
+          "Some filters were not applied because of incompatibility with data. Contact app developer",
+          type = "warning",
+          duration = 10
+        )
+        logger::log_warn(is_filter_ok)
+      }
+
+      raw_data()
```
pawelru, line 180: Can we not return (i.e. return `NULL`) from `raw_data_checked` / `teal_data_rv_validate `?

You might be concerned about this being re-executed on `raw_data` / `teal_data_rv()` changes (if we allow this). Please note that this reactive is used only in `output$error` so what's the point of wrapping this in reactive? Can we move this all in there? Then that problem would be gone.

R/teal_data_module.R
```diff
@@ -0,0 +1,52 @@
+#' Data module for `teal` applications
+#'
+#' Create `shiny` module to supply or modify data in a `teal` application.
```
pawelru, line 3:
```suggestion
#' Creates `teal_data_module` object - a `shiny` module to supply or modify data in a `teal` application.
```

R/teal_data_module.R
```diff
@@ -0,0 +1,52 @@
...
+#'
+#' This function creates a `shiny` module that allows for running data pre-processing code after the app starts.
+#' The body of the server function will be run in the app rather than in the global environment.
+#' This means it will be run every time the app starts, so use sparingly.
+#'
+#' Pass this module instead of a `teal_data` object in a call to `init`.
```
pawelru, line 9: Introduce a link to init() instead of just the code formatting.

R/teal_data_module.R
```diff
@@ -0,0 +1,52 @@
...
+#'
+#' See vignette "Data as shiny Module" for more details.
```
pawelru, line 11: Introduce a link.

R/module_nested_tabs.R
gogonzo: Yup, too early for this. I was testing something and pushed by mistake. Reverting.

FYI we will remove `resolve_delay` from `ui` functions anyway soon.

R/init.R
```diff
@@ -15,10 +15,11 @@
 #' an end-user, don't use this function, but instead this module.
 #'
 #' @param data (`TealData` or `TealDataset` or `TealDatasetConnector` or `list` or `data.frame`
-#' or `MultiAssayExperiment`, `teal_data`)\cr
+#' or `MultiAssayExperiment`, `teal_data`, `teal_data_module`)\cr
```
pawelru, line 18:
```suggestion
#' or `MultiAssayExperiment` or `teal_data` or `teal_data_module`)\cr
```

R/init.R
```diff
@@ -15,10 +15,11 @@
...
 #' `R6` object as returned by [teal.data::cdisc_data()], [teal.data::teal_data()],
 #' [teal.data::cdisc_dataset()], [teal.data::dataset()], [teal.data::dataset_connector()] or
-#' [teal.data::cdisc_dataset_connector()] or a single `data.frame` or a `MultiAssayExperiment`
+#' [teal.data::cdisc_dataset_connector()] or [teal::teal_data_module()] or a single `data.frame` or
+#' a `MultiAssayExperiment`
 #' or a list of the previous objects or function returning a named list.
```
pawelru, line 23: to add:
or `teal_data_module` object created by `teal_data_module()`

(the latter with link)

R/teal_data_module.R
```diff
@@ -0,0 +1,52 @@
...
+#'     moduleServer(id, function(input, output, session) {
+#'       eventReactive(input$submit, {
+#'         data <- within(
+#'           teal.data::teal_data(),
+#'           {
+#'             dataset1 <- iris
+#'             dataset2 <- mtcars
+#'           }
+#'         )
+#'         teal.data::datanames(data) <- c("iris", "mtcars")
```
pawelru, line 37: why prefix `teal.data` (on which `teal` depends on) and not prefix `teal.code` (which is _just_ in imports)?
it should be the other way around right?
Would be the best to avoid prefixing at all if possible

R/init.R
gogonzo: Hashables are for different reason. Hashes are not used for reproducibility purposes, and they are not returned in the R code. This single hash is needed to identify an app. At this moment it is used as "id" for `filter` argument, to check whether filter is applied to the same app. This means:
- if you have an app with dynamic data source you'd like to make sure that filters which you upload are done for this app.
- filters done for other app (possibly for other modules and data) should not be possible to upload for this app

chlebowa: Should we update the comment a few lines up to be more descriptive?

pawelru: > Should we update the comment a few lines up to be more descriptive?

Yes - that would be great for a person from the outside.

> This single hash is needed to identify an app

So I was wrong with my interpretation but the problem stays the same. The way how you calculate the hash now identify whether the data loading module is the same. You can have multiple _apps_ with the same _data loading module_. 

I have other objections for other cases of that if but it's going beyond the scope of this PR so will keep quiet here.

R/utils.R
```diff
@@ -140,3 +140,72 @@ resolve_modules_datanames <- function(modules, datanames, join_keys) {
...
+#' @param datanames (`character`) names of datasets available in the `data` object
+#'
+#' @return A `character(1)` containing error message or `TRUE` if validation passes.
+#' @keywords internal
+check_modules_datanames <- function(modules, datanames) {
+  recursive_check_datanames <- function(modules, datanames) {
+    # check teal_modules against datanames
+    if (inherits(modules, "teal_modules")) {
+      sapply(modules$children, function(module) recursive_check_datanames(module, datanames = datanames))
+    } else {
```
pawelru, line 159: just for safety -> add: else if inherits modules "teal_module" ... else stop(...)

R/init.R
chlebowa: If app1 and app2 use the same data module, then app2 _can_ use a filter snapshot created in app1.
Note that `modules` is also hashed.

pawelru: > If app1 and app2 use the same data module, then app2 _can_ use a filter snapshot created in app1. Note that `modules` is also hashed.

So then few more comments:

1. Just to clarify as I wanted to understand this better.
IMO those two sentences conflicting with each other. If app1 and app2 uses the same data module but different analysis modules then the has won't be identical (as we include modules as well) and we cannot re-use the filter. Am I missing something here?

2. Also: are you sure?

Let's leave 1. and assume same data module and same analysis modules. Let's assume the following:
- module X that allows to load multiple csv files and apply datanames according to file names
- APP1: data module X that returns datasets A, B, C + module on "all" datanames
- APP2: data module X that returns datasets K, L, M + module on "all" datanames

So the hash here will be the same and are you sure that we can re-apply filters from APP1 to APP2? Am I right here?

I have the impression that this becomes imperfect as we introduce the concept of teal_data_module in which the same module can return different datasets. I am just trying to find a good test case.

R/module_teal_with_splash.R
gogonzo: `srv_teal` requires `teal_data` object and we can't send `qenv.error` there or `simpleError`. There (srv_teal) is `observer` which waits for non-NULL value from "raw_data". When non-null value is propagated further than we have unexpected `insertUI`, and app failure.

vignettes/data-as-shiny-module.Rmd
```diff
@@ -0,0 +1,127 @@
...
+vignette: >
+  %\VignetteIndexEntry{Data as shiny Module}
+  %\VignetteEngine{knitr::rmarkdown}
+  %\VignetteEncoding{UTF-8}
+---
+
+## Introduction
+
+Proper functioning of any `teal` application requires presence of a `teal_data` object.
+Typically, a `teal_data` object created in the global environment will be passed to the `data` argument in `init`.
```
pawelru, line 16: Does it have to be a _global env_? I propose to rephrase it into just: It's required as a `data` argument in `init`.

R/module_teal.R
gogonzo: I'll revert, `once = TRUE`, `ignoreNULL` is TRUE by default

R/module_teal_with_splash.R
gogonzo: You're right, `simpleCondition(c("a", "b"))$message` returns a vector. Will fix

R/module_teal_with_splash.R
gogonzo: What do you mean by "full stop"?

vignettes/data-as-shiny-module.Rmd
pawelru: There are few more occurrences of "global environment". I feel that what we want to say is: "before the app defition" (or rephrase of this). We can include following article to better understand the difference: https://shiny.posit.co/r/articles/improve/scoping/

UPDATE: actually we can re-use the terminology from there which is: "global objects" / "per-session objects".

R/module_teal_with_splash.R
pawelru: Full sentence with a dot at the end.

vignettes/data-as-shiny-module.Rmd
```diff
@@ -0,0 +1,127 @@
...
+- a `server` function: accepts only one argument, `id`; defines server logic for the data module, including data creation; must return a reactive expression containing a `teal_data` object
+
+`teal` will run this module when the application starts and the resulting `teal_data` object that will be used throughout all `teal` (analytic) modules. 
+
+## Creating Data In-App
+
+One case for postponing data operations are data sets that are dynamic, frequently updated.
+Such data cannot be created once and kept in the global environment.
+Using `teal_data_module` allows to create a data set from scratch every time the user starts the application.
+This the user will always have access to the most recent version of the data.
```
pawelru, line 32:
```suggestion
This way, the user will always have access to the most recent version of the data.
```

R/init.R
chlebowa: Ad 1. Sorry for the confusing phrasing. I did mean that in order to use an uploaded snapshot, it must come from an app that used the same data (module) **and** modules. You are right, those are not exactly different apps. Which is the point ot the hash.

Ad 2. For the data module to return different (or even differently named) datasets it would have to have either a) different server code or b) the same server code that _modifies_ data (see the new vignette for examples), where different data is supplied. The former seems safe enough. The latter, not necessarily.

vignettes/data-as-shiny-module.Rmd
pawelru: Also: I don't understand this. What is "version of the data"? If the data is static (no change in time) - how the modules assures that it's "most recent".

I feel that what you mean is: "data will be recreated for each application session". But you already have such statement before. So maybe add: "In some of the cases, this will assure that the most recent version of the data is used".

vignettes/data-as-shiny-module.Rmd
```diff
@@ -0,0 +1,127 @@
...
+    moduleServer(id, function(input, output, session) {
+      reactive({
+        data <- within(
+          teal.data::teal_data(),
+          {
+            dataset1 <- iris
+            dataset2 <- mtcars
+          }
+        )
+        teal.data::datanames(data) <- c("dataset1", "dataset2")
```
pawelru, line 52: same as in https://github.com/insightsengineering/teal/pull/957/files/422efa6288c33105a6c94ee3681f2f877e5a6344#r1390896679

vignettes/data-as-shiny-module.Rmd
```diff
@@ -0,0 +1,127 @@
...
+        )
+        data_modified
+      })
+    })
+  }
+)
+
+app <- init(
+  data = data_mod,
+  module = list(example_module())
```
pawelru, line 111: does it have to be wrapped in a list?

gogonzo: no, I'll fix this

R/init.R
```diff
@@ -142,26 +143,13 @@ init <- function(data,
...
-    # list_to_teal_slices is lifted from teal.slice package, see zzz.R
-    # This is a temporary measure and will be removed two release cycles from now (now meaning 0.13.0).
-    filter <- list_to_teal_slices(filter)
-  }
-  # convert teal.slice::teal_slices to teal::teal_slices
-  filter <- as.teal_slices(as.list(filter))
-
-  # Calculate app hash to ensure snapshot compatibility. See ?snapshot. Raw data must be extracted from environments.
+  # Calculate app hash to ensure snapshot compatibility.
+  # See ?snapshot. Raw data must be extracted from environments.
```
chlebowa, line 147:
```suggestion
  # Calculate app id that will be used to stamp filter state snapshots.
  # App id is a hash of the app's data and modules.
  # See "transferring snapshots" section in ?snapshot.
```

R/init.R
pawelru: AD 1. Thanks. Now it's more clear.

AD 2. I am actually thinking about yet another possibility of dynamically named object -> things like: `x <- foo_named_list_of_df(input$foo, input$bar); list2env(x, env = current_env())`. The server code is the same but it can return different things based on user inputs. This essentially means that our identifier mechanism based on hashing of server fun body is imperfect (also taking into account use case b) you mentioned).

pawelru (commented):
OK. I am done for this round. Please ping me after you finish pushing corrections.

vignettes/data-as-shiny-module.Rmd
gogonzo: Global environment is still a correct, precise term. "global objects" is an informal name used by `shiny` but they still explain as

```r
Objects defined in global.R (...) they are loaded into the global environment of the R session
```

We can say also, "objects created before `teal::init`"?

R/init.R
chlebowa: Right. So in the particular case you describe snapshot compatibility is upheld. Data names may be different but the data objects themselves are the same. If you decide to change column names, _then_ you're in trouble.

The question is, is this a valid use case? Are dataset names (let alone variable names) ever left up to the app user? We don't even provide for the app user to modify variable labels (yet).

vignettes/data-as-shiny-module.Rmd
pawelru: I actually find other, much simpler statement that I find useful here:

> outside of the server function definition

I think I can do the following
```
e <- new_env()
e$x <- teal_data(...)
init(data = e$x, ...)
```

Whether `e$x` is a global env object or not is debatable but IMHO it fulfils the definition of a "global object".

Or the following:
```
init(data = teal_data(...), ...)
```

No assignment to global env.

Maybe we can have a short glossary at the very beginning? My point is ~not to use~ limit the use an "environment" term throughout the content of this article as this might be too complex for an average user. My idea is to refer that article first and say that teal-data-module is an implementation of per-session data objects whereas global objects are still allowed (see other vignette here: XYZ).

vignettes/data-as-shiny-module.Rmd
chlebowa: If you read the whole paragraph, not every sentence in isolation, the last sentence clearly refers to a dynamic data situation. Or is this my author bias?

gogonzo: I've removed last sentence and seems to be fine.

vignettes/data-as-shiny-module.Rmd
chlebowa: I agree that we don't want to burden the app dev with environments but "global environment" is a basic concept of interactive R.

I don't think a glossary would solve the problem. You would have to explain it in terms of environments anyway. Plus it's clutter, better to use footnotes.

> I think I can do the following
> 
> ```
> e <- new_env()
> e$x <- teal_data(...)
> init(data = e$x, ...)
> ```

You _can_ but would you? It seems a little far-fetched. Please note that the vignette says "_Typically_, a `teal_data` object created in the global environment will be passed to the `data` argument in `init`."

R/init.R
pawelru: I think it is. To me, that's one of the greatest things about teal-data-module concept. It essentially unlocks total flexibility of data names, column names, column types and everything else. We even say in the docs that (the only) limitation is that the returned object has to be of `teal_data` class. How you achieve that - it's up to the module developer. And that's great.

This feature does not fit well with our ~concept~ implementation of filters re-applicability which I am highlighting here. Now I don't know if we should invest in finding appropriate way of implement this this (here) or revisit this as a whole.

R/utils.R
gogonzo: added assert on `modules` to be `teal_modules` - it is enough as modules can't contain anything else.

R/init.R
gogonzo: Let's skip it. We will fix this param description anyway soon

vignettes/data-as-shiny-module.Rmd
pawelru: yeah right - don't want to replace technicality with a clutter

My point is to make it more _simpler_ and avoid frequent use of "environment". Thought we can "delegate" that problem to the aforementioned article. If you think it's ok then let's leave it.

R/teal_data_module.R
```diff
@@ -0,0 +1,52 @@
...
+#'
+#' Creates `teal_data_module` object - a `shiny` module to supply or modify data in a `teal` application.
+#'
+#' This function creates a `shiny` module that allows for running data pre-processing code after the app starts.
+#' The body of the server function will be run in the app rather than in the global environment.
+#' This means it will be run every time the app starts, so use sparingly.
+#'
+#' Pass this module instead of a `teal_data` object in a call to [init()].
+#'
+#' See vignette \code{vignette("data-as-shiny-module", package = "teal")} for more details.
```
chlebowa, line 11: How about linking to `https://insightsengineering.github.io/teal/latest-tag/articles/`? 
It seems to me vignettes are not built during a default installation. I installed with `install.packages` and I get 
```
> library(teal)
> browseVignettes("teal")
No vignettes found by browseVignettes("teal")
> vignette("data-as-shiny-module", package = "teal")
Warning message:
vignette ‘data-as-shiny-module’ not found 
```

vignettes/data-as-shiny-module.Rmd
gogonzo: If we don't have a clear conclusion I suggest to leave it aside and open an issue to revisit this vignette. This PR blocks a lot and each day postpones even bigger issues in teal which we have to solve in 1 month.

chlebowa: > If we don't have a clear conclusion I suggest to leave it aside and open an issue to revisit this vignette. This PR blocks a lot and each day postpones even bigger issues in teal which we have to solve in 1 month.

I agree. We can seek an outsider's opinion in the mean time.

vignettes/data-as-shiny-module.Rmd
pawelru: I'm ok with this. My vignettes comments are not that _strong_ anyway.

vignettes/data-as-shiny-module.Rmd
pawelru: I'm ok with this. My vignettes comments are not that strong anyway.

R/teal_data_module.R
gogonzo: vignette is fine, you need to build the vignette when installing from source. Normally on CRAN or other repos they are build.

chlebowa: Oh, that's why it happens. Very well :ok_hand:

pawelru (approved):
This is good. Thank you. Let's take hashing discussion elsewhere not to block everything that depends on this.

Issue #601, opened 2022-04-01
Fix note


pawelru:
BEE installation as of 01/04/2022
```
* checking for unstated dependencies in vignettes ... NOTE
'library' or 'require' call not declared from: ‘ggplot2’
```

PR #962, opened 2023-11-10
Extend `validate_has_data` to accept vector input

### LINKED ISSUE #961: [Bug]: tm_outliers module - not able to use categorical_var
### What is your question?

Hi,

In the  "tm_outliers" module I'm not able to add "categorical_var" parameter,  because if I do there is an error in the app "no data loaded". I looked at the example https://insightsengineering.github.io/teal.modules.general/latest-tag/articles/usingoutliers-module.html but it does not help to understand an issue. Could you advise what is wrong? Below a reproducible example using iris data

```r
library(teal.modules.general)
library(dplyr)
library(tidyr)
options(shiny.useragg = FALSE)

# prepare data
iris2 <- iris %>%
  mutate(
    grp_slength = cut(Sepal.Length, breaks = c(0, 5, 8, Inf)),
    grp_swidth = cut(Sepal.Width, breaks = c(0, 1, 2, 3, 4, Inf)),
    petal_color = as.factor(rep(c("orange", "blue", "magenta"), 50))
  )

# App
app <- init(
  data = teal_data(
    dataset("iris2", iris2)
  ),
  modules = modules(
    tm_data_table("Data Table"),
    tm_outliers(
      "Outliers",
      outlier_var = data_extract_spec(
        dataname = "iris2",
        select = select_spec(
          choices = variable_choices(iris2, subset = names(Filter(isTRUE, sapply(iris2, is.numeric)))),
          selected = "Sepal.Length",
          multiple = FALSE,
          fixed = FALSE
        )
      ),
      categorical_var = data_extract_spec(
        dataname = "iris2",
        select = select_spec(
          choices = variable_choices(iris2, subset = names(Filter(isTRUE, sapply(iris, is.factor)))),
          selected = "Species",
          multiple = FALSE,
          fixed = FALSE
        )
      )
    )
  ),
  header = "Reproducible example"
)
shinyApp(app$ui, app$server)
```
### Code of Conduct

### Contribution Guidelines

### Security Policy

vedhav:
Closes insightsengineering/teal.modules.general#600


chlebowa (changes requested):
I disagree with the applied solution.
`validate_has_data` is a function specialized to deal with datasets, i.e. `data.frame`s.
There is nothing wrong with the function, it was only used incorrectly by the module developer who passed a vector.

I think it is better to assert argument type here and fix the offending call (calls?) by adding `drop = FALSE`. If a vector is passed by accident (as is clearly the case) the app should fail immediately and the bug should be caught in the console by the app developer, not in the app by a user.

Also, the changes to the documentation do not reflect changes to the functionality, and the changes to the code are more of a hack than an adjustment.

vedhav:
Thanks @chlebowa reverted the changes and now it only has an assert to check for `data.frame` so it fails.
Only `teal.modules.general` had two places that could potentially return vectors to this function so created a PR for that [teal.modules.general PR #601](https://github.com/insightsengineering/teal.modules.general/pull/601)

R/validations.R
```diff
@@ -46,6 +46,7 @@ validate_has_data <- function(x,
   stopifnot(
     "Please provide a character vector in msg argument of validate_has_data." = is.character(msg) || is.null(msg)
   )
+  checkmate::assert_data_frame(x)
   validate(need(!is.null(x) && is.data.frame(x), "No data left."))
```
chlebowa, line 50:
```suggestion
  checkmate::assert_data_frame(x)
```
I guess we don't need the `validate` any more.

vedhav: True

R/validations.R
```diff
@@ -46,7 +46,7 @@ validate_has_data <- function(x,
   stopifnot(
     "Please provide a character vector in msg argument of validate_has_data." = is.character(msg) || is.null(msg)
   )
```
chlebowa, line 48:
```suggestion
  checkmate::assert_string(msg, null.ok = TRUE)
```
What do you think?

chlebowa (approved):
:+1:

PR #983, opened 2023-11-28
Adds `within` and `eval_code` to `teal_data_module`

averissimo:
# Pull Request

Fixes #nnn

### Changes description

- Adds `within` and `eval_code` to `teal_data_module`
- Adds tests for new functions (100% coverage)
- Updates `data-as-shiny-module` vignette
- 🟠 Moves `{teal.code}` from `Suggests` to `Imports`


R/teal_data_module-eval_code.R
```diff
@@ -0,0 +1,87 @@
...
+    expr <- call("{", expr)
+  }
+
+  calls <- as.list(expr)[-1]
+
+  # Inject extra values into expressions.
+  calls <- lapply(calls, function(x) do.call(substitute, list(x, env = extras)))
+
+  eval_code(object = data, code = as.expression(calls))
+}
```
averissimo, line 87: The body is a copy from `teal.code:::within.qenv`, so in practice it could be replaced with code below _(reducing duplicated code accross `{teal.*}`)_:

Thoughts?

```suggestion
within.teal_data_module <- function(data, expr, ...) {
  within.qenv <- getFromNamespace("within.qenv", "teal.code")
  within.qenv(data, expr, ...)
}
```

chlebowa: 1. I think `getS3method("within", "qenv", envir = as.environment(getNamespace("teal.code")))` would be a better choice.
2. It shouldn't be necessary anyway because `teal.code` must be installed when `teal` is.
3. At the moment `within` only processes the expression and calls `eval_code`. In this context this is a valid simplification. I'm not sure, though :thinking:

averissimo: > 1. I think getS3method("within", "qenv", envir = as.environment(getNamespace("teal.code"))) would be a better choice.
> 2. It shouldn't be necessary anyway because teal.code must be installed when teal is.

That's better! Is the `envir` parameter required? _(as you said, `teal.code` is imported and it seems to find the within)_

> 3. At the moment within only processes the expression and calls eval_code. In this context this is a valid simplification. I'm not sure, though 🤔

@vedhav brough a smilar concern, I'm on the fence as I prefer to minimize repetitive code, but I don't think this solution is "best practice" (calling `within.qenv` with a different data object seems like a red flag). 

I would prefer if there was `within.default` (non-exported) or `.generic_within` in `{teal.code}` that was the backbone of both of those calls and still allow for deviations in the future of `within.qenv` that wouldn't affect `within.teal_data_module`.

chlebowa: > That's better! Is the `envir` parameter required?

It isn't but just in case there is a _different_ `within.qenv` defined somewhere...

chlebowa: > I would prefer if there was `within.default`

The `within` generic is a `base` function so defining a `default` that suits our particular classes would cause a mess.

R/zzz.R
```diff
@@ -32,3 +32,7 @@ setdiff_teal_slices <- getFromNamespace("setdiff_teal_slices", "teal.slice")
 coalesce_r <- getFromNamespace("coalesce_r", "teal.slice")
 # all *Block objects are private in teal.reporter
 RcodeBlock <- getFromNamespace("RcodeBlock", "teal.reporter") # nolint
+
+# Use non-exported function(s) from teal.code
+# This one is here because format_expression should not be exported from teal.code
+format_expression <- getFromNamespace("format_expression", "teal.code")
```
averissimo, line 38: This will be repaced with `lang2calls` when we merge the `refactor` branch due to https://github.com/insightsengineering/teal.code/pull/176

R/teal_data_module-eval_code.R
averissimo: Using the `.default` was a suggestion if it didn't interfere with `base::within` (by not exporting it, but even then it might be "dangerous")

chlebowa: Ah, but S3 methods must be exported :wink:

tests/testthat/test-teal_data_module-eval_code.R
```diff
@@ -0,0 +1,118 @@
...
+  testthat::expect_setequal(
+    c(names(iris), "id"),
+    colnames(shiny::isolate(tdm2$server("test")()[["IRIS"]]))
+  )
+
+  # Original teal_data_module was left untouched
+  testthat::expect_setequal(
+    c(names(iris)),
+    colnames(shiny::isolate(tdm$server("test")()[["IRIS"]]))
+  )
```
gogonzo, line 117: Please don't make indirect tests. 

```suggestion
  testthat::expect_identical(
    testthat::expect_identical(
      shiny::isolate(tdm2$server("test")()[["IRIS"]]),
      within(iris, id <- nrow(.data))
    )
  )
```

gogonzo (commented):
There are missing tests and poc for teal application. [There are already tests](https://github.com/insightsengineering/teal/blob/main/tests/testthat/test-module_teal_with_splash.R) on `data` (teal_data_module) output. I wonder what would be the output of `eval_code(teal_data_module, code)` if there is an error in `teal_data_module` or in the `code`.

R/teal_data_module-eval_code.R
```diff
@@ -0,0 +1,98 @@
...
+    ui = function(id) {
+      ns <- NS(id)
+      object$ui(ns("mutate_inner"))
+    },
+    server = function(id) {
+      moduleServer(id, function(input, output, session) {
+        data <- object$server("mutate_inner")
+        eventReactive(data(),
+          {
+            eval_code(data(), code)
```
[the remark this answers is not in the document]
averissimo: > data() returns NULL (I can see it is just ignored by eventReactive) which means that it doesn't go to srv_teal_with_splash but I wonder what are the consequences.

It reaches `srv_teal_with_splash` with a `shiny.silent.error`. The same as if it was non-mutated `teal_data_module`

I've pushed a protection that will send all error handling upstream, i.e. only handles reactive with `teal_data`

```r
        eventReactive(teal_data_rv(),
          {
            data <- tryCatch(teal_data_rv(), error = function(e) e)

              eval_code(data, code)
            } else {
              data
            }
          },
          ignoreNULL = TRUE
        )
```

<details>

<summary>Click to expand a testing application</summary>

```r
tdm <- teal_data_module(
  ui = function(id) {
    ns <- NS(id)
    tags$ul(
      tags$li(actionButton(ns("submit"), label = "Load data")),
      tags$li(actionButton(ns("submit_0"), label = "🟢 Load data (silent shiny error -- empty)")),
      tags$li(actionButton(ns("submit_1"), label = "🟥 Load data (silent shiny error)")),
      tags$li(actionButton(ns("submit_2"), label = "🔵 Load data (qenv.error)")),
      tags$li(actionButton(ns("submit_3"), label = "🔴 Load data (error)")),
      tags$li(actionButton(ns("submit_4"), label = "🍎 Load data (non-teal_data qenv)"))
    )
  },
  server = function(id) {
      # Mock shiny.silent.error
      result <- reactiveVal(structure(
        list(message = ""),
        class = c("shiny.silent.error", "validation", "error", "condition")
      ))

      # Valid result
      observeEvent(input$submit, {
        data <- within(teal_data(), {
          dataset1 <- iris
          dataset2 <- mtcars
        })
        datanames(data) <- c("dataset1", "dataset2")
        result(data)
      })

      # shiny.silent.error with empty message
      observeEvent(input$submit_0, {
        result(structure(
          list(message = ""), 
          class = c("shiny.silent.error", "validation", "error", "condition")
        ))
      })

      # shiny.silent.error with **NON**-empty message
      observeEvent(input$submit_1, {
        result(structure(
          list(message = "🟥 shiny.silent.error with non-empty message"),
          class = c("shiny.silent.error", "validation", "error", "condition")
        ))
      })

      # qenv.error
      observeEvent(input$submit_2, result(teal_data() |> within(a + "🔵")))

      # error class
      observeEvent(input$submit_3, {
        result(structure(
          list(message = "🔴 is an object of class `error"),
          class = "error"
        ))
      })

      # valid qenv (that is not teal.data)
      observeEvent(input$submit_4, {
        result(teal.code::new_qenv() |> within({
          message("🍎")
          dataset1 <- iris
        }))
      })

      result
    })
  }
)

tdm_new <- eval_code(tdm, "dataset1$id <- seq_len(nrow(dataset1))") |>
  within(
    dataset1$checksum <- apply(dataset1, 1, rlang::hash)
  )

init(
  data = tdm_new,
  modules = list(
    example_module()
  )
) |> shiny::runApp()
```

</details>

image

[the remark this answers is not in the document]
averissimo: Duplicating validation logic here seemed counter-productive.

This solution will process all `within`s and stop on the first that fails (propagating that up to `srv_teal_with_slash`

tests/testthat/test-teal_data_module-eval_code.R
```diff
@@ -0,0 +1,118 @@
...
+
+  tdm2 <- eval_code(tdm, quote(IRIS$id <- seq_len(nrow(IRIS)))) # nolint: object_name.
+
+  # Columns were added via eval_code.teal_data_module
+  testthat::expect_setequal(
+    c(names(iris), "id"),
+    colnames(shiny::isolate(tdm2$server("test")()[["IRIS"]]))
+  )
+})
```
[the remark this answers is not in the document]
averissimo: I agree with you, this was mainly to keep coverage at 100%

(without the test)

image

tests/testthat/test-teal_data_module-eval_code.R
averissimo: Done in 51dcadff6

R/zzz.R
averissimo: No reason to not use it right now, done in a861bec97

R/zzz.R
```diff
@@ -32,3 +32,7 @@ setdiff_teal_slices <- getFromNamespace("setdiff_teal_slices", "teal.slice")
 coalesce_r <- getFromNamespace("coalesce_r", "teal.slice")
 # all *Block objects are private in teal.reporter
 RcodeBlock <- getFromNamespace("RcodeBlock", "teal.reporter") # nolint
+
+# Use non-exported function(s) from teal.code
+# This one is here because lang2calls should not be exported from teal.code
+lang2calls <- getFromNamespace("lang2calls", "teal.code")
```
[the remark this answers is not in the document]
chlebowa: @pawelru NB, we _only_ do it for private functions because public ones don't need it.

gogonzo (approved):
Nothing to add in my opinion. Good job!

R/teal_data_module-eval_code.R
```diff
@@ -0,0 +1,124 @@
...
+              data
+            } else {
+              validate(
+                need(
+                  FALSE,
+                  paste(
+                    sep = "\n",
+                    "Error when executing `teal_data_module`:",
+                    paste0(
+                      "It must always return a reactive with `teal_data`, it returns object of class(es): ",
```
chlebowa, line 59: As in `srv_teal_with_splash`, I believe this error message is for the app dev, not for the app user. It should shop up in the console, not in the app.

R/teal_data_module-eval_code.R
```diff
@@ -0,0 +1,124 @@
...
+
+        if (!is.reactive(teal_data_rv)) {
+          stop("The `teal_data_module` must return a reactive expression.", call. = FALSE)
+        }
+
+        eventReactive(teal_data_rv(),
+          {
+            data <- tryCatch(teal_data_rv(), error = function(e) e)
+
+            if (inherits(data, "teal_data")) {
```
chlebowa, line 47:
```suggestion
            if (inherits(data, "teal_data", "qenv.error")) {
```

R/teal_data_module-eval_code.R
```diff
@@ -0,0 +1,124 @@
...
+              eval_code(data, code)
+            } else if (inherits(data, "error")) {
+              data
+            } else {
```
chlebowa, line 51: I don't understand when this clause is triggered.

gogonzo: It is possible when `server` of `teal_data_module` returns error (not related to qenv.error). There is an example app which @averissimo provided:

https://github.com/insightsengineering/teal/pull/983#discussion_r1412098320


@chlebowa I think it might be removed tbh. When "evaluated" `teal_data_module` throws then `eventReactive` throws as well and error would be caught by `srv_teal_with_splash`.

gogonzo: @averissimo I think we can remove `else if (inherits(data, "error")) {` and keep only `else`. Reason for this is that `inherits(data, "error")` expects this error to be already **handled**.  I expect handled errors to be very rare output of `teal_data_module` and we can throw them to the `else` or remove else at all and just return `data` (not evaluated).

averissimo: This clause catches both `shiny.silent.errors` and `qenv.errors` and pushes those up to be handled.

We could make it more specific to specifically catch those 2, however, we risk having error messages that look like this:

image

gogonzo: Problem with your app is that server returns objects of class errors but it doesn't really throw errors. If you 

There is a difference between
```t
structure(
  list(message = "🔴 is an object of class `error"),
  class = "error"
)

# and 
stop("this is error")
```

This app for example is done in a way which would never get inside of the `eventReactive` in `eval_code`. 

```r

tdm <- teal_data_module(
  ui = function(id) NULL,
  server = function(id) {
      # Mock shiny.silent.error
      reactive(stop("error"))
    })
  }
)

tdm_new <- eval_code(tdm, "dataset1$id <- seq_len(nrow(dataset1))") |>
  within(
    dataset1$checksum <- apply(dataset1, 1, rlang::hash)
  )

init(
  data = tdm_new,
  modules = list(
    example_module()
  )
) |> shiny::runApp()
```

Same case with `shiny.error`

```r
tdm <- teal_data_module(
  ui = function(id) NULL,
  server = function(id) {
      # Mock shiny.silent.error
      reactive(validate(need(FALSE, "this is shiny.error")))
    })
  }
)

tdm_new <- eval_code(tdm, "dataset1$id <- seq_len(nrow(dataset1))") |>
  within(
    dataset1$checksum <- apply(dataset1, 1, rlang::hash)
  )

init(
  data = tdm_new,
  modules = list(
    example_module()
  )
) |> shiny::runApp()

```

averissimo: It will probably be very rare as you said, but this way we allow for those ocassions to produce a readable message.

R/teal_data_module-eval_code.R
averissimo: This is currently handled in the next `else if`, do we really need to continue and eval when an error exists?

R/teal_data_module-eval_code.R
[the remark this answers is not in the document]
gogonzo: Locally I changed code to this. And it still returns meaningful message for all the cases. Please try

```r
        eventReactive(teal_data_rv(),
          {
            print("I'm here")
            if (inherits(teal_data_rv(), c("teal_data", "qenv.error"))) {
              eval_code(teal_data_rv(), code)
            } else {
              teal_data_rv()
            }
          },
          ignoreNULL = TRUE
        )
```

R/teal_data_module-eval_code.R
chlebowa: The `qenv.error` mechanism is built specifically for evaluation errors such that an error rides along all subsequent evaluations and is interpreted at the end. This is why we have `eval_code` methods for `qenv.error`. I wanted to keep to that mechanism rather than handle `qenv.errors` differently. In my mind, I want extra handling for things that _aren't_ covered by this mechanism.

R/teal_data_module-eval_code.R
pawelru: It depends. Let's say we have teal_data_module that reads csv files only and an app user select e.g. pdf file then we probably need to expose error message for app user.

R/teal_data_module-eval_code.R
pawelru: One interesting pattern that I have found when working on the `verdepcheck` is to make a _simple_ (i.e. non S3 / non S4 etc.) (implementation) function for each method and these methods are essentially a one-liner calling those functions. Please find an example in R6 context: https://github.com/r-lib/pkgdepends/blob/main/R/pkg-plan.R - all public methods are calling private (implementation) functions.

Then you can re-use those implementation functions from multiple classes but there are some obvious drawbacks though for package size, documentation, maintenance, debugging etc.

This however is not addressing an exporting / importing issue mentioned here 😈

R/zzz.R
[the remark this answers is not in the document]
pawelru: I know I know :)

tests/testthat/test-teal_data_module-eval_code.R
```diff
@@ -0,0 +1,169 @@
...
+  tdm2 <- eval_code(tdm, "1 + 1")
+
+  shiny::testServer(
+    app = tdm2$server,
+    expr = {
+      testthat::expect_null(td())
+    }
+  ) %>%
+    testthat::expect_no_error()
+})
```
gogonzo, line 169:
```suggestion
})

testthat::test_that("eval_code.teal_data_module handles a `error`", {
  tdm <- teal_data_module(
    ui = function(id) div(),
    server = function(id) {
      shiny::moduleServer(id, function(input, output, session) {
        stop("I am an error")
      })
    }
  )

  tdm2 <- eval_code(tdm, "1 + 1")

  testthat::expect_error(
    shiny::testServer(
      app = tdm2$server,
      expr = {}
    )
  )
})
```

tests/testthat/test-teal_data_module-eval_code.R
```diff
@@ -0,0 +1,169 @@
...
+
+  tdm2 <- eval_code(tdm, "1 + 1")
+
+  shiny::testServer(
+    app = tdm2$server,
+    expr = {
+      testthat::expect_null(td())
+    }
+  ) %>%
+    testthat::expect_no_error()
```
gogonzo, line 168: Please don't ever use `%>%` in the developer code. Should be only in the vignettes/docs

R/teal_data_module-eval_code.R
```diff
@@ -0,0 +1,124 @@
...
+#'   ui = function(id) div(id = shiny::NS(id)("div_id")),
+#'   server = function(id) {
+#'     shiny::moduleServer(id, function(input, output, session) {
+#'       shiny::reactive(teal_data(IRIS = iris))
+#'     })
+#'   }
+#' )
+#' \dontrun{
+#' eval_code(tdm, "IRIS <- subset(IRIS, Species == 'virginica')")
+#' }
```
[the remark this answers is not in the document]
gogonzo: Comment still valid ⬆️

gogonzo (approved):
Please address comments

tests/testthat/test-teal_data_module-eval_code.R
averissimo: Ok, I used them in testthat as they are well integrated in testthat https://testthat.r-lib.org/news/index.html?q=pipe#new-expectations-1-0-0 and help create clean tests.

I'll remove them though if that is the standard for `{teal}`

averissimo: Done in b1625ba52 (also removed a native pipe in tests that was useless)

PR #989, opened 2023-11-29
removed additional line break from show code.

kartikeyakirar:
this fixes https://github.com/insightsengineering/teal.gallery/issues/113

### DIFF
--- a/R/get_rcode_utils.R
+++ b/R/get_rcode_utils.R
@@ -49,7 +49,7 @@ get_datasets_code <- function(datanames, datasets, hashes) {
   if (length(str_prepro) == 0) {
     str_prepro <- "message('Preprocessing is empty')"
   } else if (length(str_prepro) > 0) {
-    str_prepro <- paste0(str_prepro, "\n\n")
+    str_prepro <- paste(paste0(str_prepro, collapse = "\n"), "\n\n")
   }
 
   str_hash <- paste(


R/get_rcode_utils.R
```diff
@@ -49,7 +49,7 @@ get_datasets_code <- function(datanames, datasets, hashes) {
   if (length(str_prepro) == 0) {
     str_prepro <- "message('Preprocessing is empty')"
   } else if (length(str_prepro) > 0) {
-    str_prepro <- paste0(str_prepro, "\n\n")
+    str_prepro <- paste(paste0(str_prepro, collapse = "\n"), "\n\n")
```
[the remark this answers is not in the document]
kartikeyakirar: I will change it to` str_prepro <- paste0(paste(str_prepro, collapse = "\n"), "\n\n")` my bad.

chlebowa:
I wonder if the logic is correct. If there is no pre-processing code, something line is added, so the two empty lines should be appended always, not just in the `else`. What do you think?

kartikeyakirar:
Added changes here [e39f0df](https://github.com/insightsengineering/teal/pull/989/commits/e39f0df4cda7cc382ca43e97ff3a2684791e189b)

R/get_rcode_utils.R
```diff
@@ -46,29 +46,25 @@ get_rcode_str_install <- function() {
 #' @keywords internal
 get_datasets_code <- function(datanames, datasets, hashes) {
   str_prepro <- teal.data:::get_code_dependency(attr(datasets, "preprocessing_code"), names = datanames)
-  if (length(str_prepro) == 0) {
-    str_prepro <- "message('Preprocessing is empty')"
-  } else if (length(str_prepro) > 0) {
-    str_prepro <- paste0(str_prepro, "\n\n")
-  }
+  str_prepro <- ifelse(length(str_prepro) == 0, "message('Preprocessing is empty')", str_prepro)
```
chlebowa, line 49: This is wrong. `ifelse` always returns a vector the same length as `test`, in this case `1L`. You would lose all but the first element.

`ifelse` is _not_ shorthand for `if (cond) val1 else val2`.

R/get_rcode_utils.R
chlebowa, line 1: How about, rather than adding `"\n\n"` to the first two elements, we `paste` them all with `collapse = "\n\n"` at the end? Would changing the return from character vector of 3 to one string break anything downstream?

chlebowa (changes requested):

R/get_rcode_utils.R
kartikeyakirar: Its better idea as the possibility of the filter code (str_filter) being an empty string, which could lead to more line breaks.  we could process here and send it as one string.  I don't see any issue with this downstream .

kartikeyakirar:
I have resolved the spacing issue. I noticed there were extra line breaks when the filter expression was excluded, so I've corrected those. Additionally, I addressed the issue of duplicate filtering expressions. 
image

chlebowa (approved):
Just a few minor changes.

PR #990, opened 2023-11-29
vignette data

chlebowa:
* adapted to `teal_data` class
* removed vignettes `including-general-data-in-teal`, `including-adam-data-in-teal`, `including-mae-data-in-teal`, and `preprocessing-data`
* added vignette `data-in-teal`
* added `bookdown` to `Suggests` to allow for cross-references within document


vignettes/data-in-teal.Rmd
```diff
@@ -0,0 +1,248 @@
+---
+title: "Including Data in teal Applications"
```
gogonzo, line 2: Please change filename to `including-data-in-teal-applications.R`

vignettes/data-in-teal.Rmd
```diff
@@ -0,0 +1,248 @@
...
+
+```{r, message=FALSE}
+library(teal)
+
+# create data objects
+IRIS <- iris
+CARS <- mtcars
+
+# create teal_data
+data <- teal_data(IRIS = IRIS, CARS = CARS)
```
gogonzo, line 44: We used to do `IRIS = iris` just to have some non-redundant line in show-r-code (redundant: `iris <- iris`, less redundant: `IRIS <- iris`). I admit, both looks redundant.

Here we don't deal with reproducible code in `teal_data` and people might think that they really need to make all datasets uppercase and it is not the case - this limitation is only CDISC datasets because of the join keys. I think we can skip `IRIS <- iris` and specify data in `teal_data` directly to make example as much concise as possible.





```suggestion
library(teal)

# create teal_data
data <- teal_data(iris = iris, mtcars = mtcars)
```

vignettes/data-in-teal.Rmd
```diff
@@ -0,0 +1,248 @@
...
+)
+adtte <- rbind(adsl, adsl, adsl)
+adtte$PARAMCD <- rep(c("OS", "EFS", "PFS"), each = 10)
+adtte$AVAL <- c(
+  rnorm(10, mean = 700, sd = 200), # dummy OS level
+  rnorm(10, mean = 400, sd = 100), # dummy EFS level
+  rnorm(10, mean = 450, sd = 200) # dummy PFS level
+)
+
+data_cdisc <- cdisc_data(ADSL = adsl, ADTTE = adtte)
```
gogonzo, line 86: Let's use `teal.data` dummy data to simplify an example

```suggestion
data_cdisc <- cdisc_data(ADSL = teal.data::rADSL, ADTTE = teal.data::rADSL)
```

vignettes/data-in-teal.Rmd
```diff
@@ -0,0 +1,248 @@
...
+```
+
+The filter panel supports `MAE` data out of the box, but `teal` itself does not guarantee that any module will work with `MAE` data the same way it works with other types of data (e.g. `ADaM`) because `MAE` has a unique structure that requires special consideration when developing a module.
+The package [`teal.modules.hermes`](https://insightsengineering.github.io/teal.modules.hermes/latest-tag/) has been developed specifically with `MAE` in mind and will be more reliable.
+
+### Delayed Data
+
+The one departure from passing a `teal_data` object to `init` is when the data does not exist in the environment where the app is run, e.g. when it has to be pulled from a remote source.
+In those cases a `teal_data_module` must be used.
+See [this vignette](https://insightsengineering.github.io/teal/latest-tag/articles/data-as-shiny-module.html) for a detailed description.
```
gogonzo, line 130: For vignettes in the same package it is enough to set a relative path.

```suggestion
See [this vignette](data-as-shiny-module.html) for a detailed description.
```

gogonzo (commented):
I would change the order of the sections. Content of the sections is fine IMO

```
## Data in `teal` Applications
### General Data
### Reproducible Data (instead Creating Data by Evaluating Code)
~~## Passing Data And Code (to remove)~~
### Delayed Data

## `teal_data` properties
##### `datanames`
##### `join_keys`
##### `verified`

## Supporting Clinical trials 
### ADaM Data
### MultiAssayExperiment Data
## Further Reading
```

I like the idea but I wanted the `properties` section to be placed at the (near) bottom so that one would jump down for clarifications and back up to the main text. Ideally, these would have been footnotes but those cannot handle code and extensive formatting (to my knowlege). What do you think about moving `Clinical Data` up?

vignettes/data-in-teal.Rmd
chlebowa: Perfect!

chlebowa:
> I would change the order of the sections. Content of the sections is fine IMO

I like the idea but I wanted the `properties` section to be placed at the (near) bottom so that one would jump down for clarifications and back up to the main text. Ideally, these would have been footnotes but those cannot handle code and extensive formatting (to my knowlege). What do you think about moving `Clinical Data` up?

vignettes/including-data-in-teal-applications.Rmd
```diff
@@ -0,0 +1,251 @@
...
+
+<br>
+
+## Clinical Data
+
+Currently `teal` supports two specialized data formats.
+
+### `ADaM` Data
+
+The container used for `CDISC` data is the `cdisc_data` class, which is an extension of `teal_data` created with the `cdisc_data` function.
```
gogonzo, line 112: There is no `cdisc_data` class❗

Target for teal is to be data-standard independent and we are very close to achieve this. At this moment `teal.data` supports every kind of data. It is only `teal.slice` and `teal.transform` which really blocks us to announce this wonderful news ;]

```suggestion
For `CDISC` data we recommend to create `teal_data` object using `cdisc_data()` function, which adds default `join_keys` to set proper relationships with `ADSL` dataset.
```

vignettes/including-data-in-teal-applications.Rmd
```diff
@@ -0,0 +1,251 @@
...
+
+data_with_objects_and_code <- teal_data(iris = iris, cars = mtcars, code = expression(iris <- iris, cars <- mtcars))
+data_with_objects_and_code
+
+data_with_objects_and_code_verified <- verify(data_with_objects_and_code)
+data_with_objects_and_code_verified
+```
+
+
+For a detailed explanation of verification, see [this `teal.data` vignette](https://insightsengineering.github.io/teal.data/latest-tag/articles/teal-data.html).
```
gogonzo, line 243: Let's refer to this https://github.com/insightsengineering/teal.data/blob/refactor/vignettes/teal-data-reproducibility.Rmd

```suggestion
For a detailed explanation of verification, see [the "`teal.data` reproducibility" vignette](https://insightsengineering.github.io/teal.data/latest-tag/articles/teal-data-reproducibility.html).
```

vignettes/including-data-in-teal-applications.Rmd
```diff
@@ -0,0 +1,251 @@
+---
+title: "Including Data in teal Applications"
+author: "NEST CoreDev"
+output: 
+  bookdown::html_document2:
```
gogonzo, line 5: why bookdown needed?

chlebowa: > * added `bookdown` to `Suggests` to allow for cross-references within document

gogonzo: Can't you do it with pure markdown syntax https://linuxhint.com/markdown-link-section/?

```
(jump to [`teal_data` properties](#teal-data-properties))
```

```
(back to [Reproducible data](#reproducible-data))
```

chlebowa: Apparently I can.

gogonzo (approved):
👍

PR #991, opened 2023-11-29
update example module

chlebowa:
Modified `example_module` to remove `data` argument from UI function. Data set selection is moved to a `renderUI` element on the server side.

### DIFF
--- a/R/dummy_functions.R
+++ b/R/dummy_functions.R
@@ -96,9 +96,7 @@ example_module <- function(label = "example teal module", datanames = "all") {
       checkmate::assert_class(data(), "teal_data")
       moduleServer(id, function(input, output, session) {
         ns <- session$ns
-        output$datanames <- renderUI({
-          selectInput(ns("dataname"), "Choose a dataset", choices = teal.data::datanames(data()))
-        })
+        updateSelectInput(session, "dataname", choices = isolate(teal.data::datanames(data())))
         output$text <- renderPrint({
           req(input$dataname)
           data()[[input$dataname]]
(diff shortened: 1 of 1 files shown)


R/dummy_functions.R
```diff
@@ -95,20 +95,27 @@ example_module <- function(label = "example teal module", datanames = "all") {
...
       })
     },
-    ui = function(id, data) {
+    ui = function(id) {
       ns <- NS(id)
       teal.widgets::standard_layout(
         output = verbatimTextOutput(ns("text")),
         encoding = div(
-          selectInput(ns("dataname"), "Choose a dataset", choices = teal.data::datanames(data)),
+          uiOutput(ns("datanames")),
```
gogonzo, line 118:
```suggestion
          selectInput(ns("datanames"), "Choose a dataset", choices = NULL),
```

R/dummy_functions.R
```diff
@@ -95,20 +95,27 @@ example_module <- function(label = "example teal module", datanames = "all") {
     server = function(id, data) {
       checkmate::assert_class(data(), "teal_data")
       moduleServer(id, function(input, output, session) {
-        output$text <- renderPrint(data()[[input$dataname]])
+        ns <- session$ns
+        output$datanames <- renderUI({
+          selectInput(ns("dataname"), "Choose a dataset", choices = teal.data::datanames(data()))
+        })
```
gogonzo, line 101: Each time data is filtered renderUI renders input. This means that when you set selection to `"ADTTE"`, after filtering this will change back to `"ADSL"`
```suggestion
        isolate({
          selectInput(ns("dataname"), "Choose a dataset", choices = teal.data::datanames(data()))
        })
```

gogonzo (changes requested):

R/dummy_functions.R
chlebowa: You mean to use `updateSeletInput` here, correct?

gogonzo: yes

gogonzo (approved):

PR #993, opened 2023-11-30
update example modules in vignettes

chlebowa:
adapted example modules to follow `example_module`

### DIFF
--- a/vignettes/adding-support-for-reporting.Rmd
+++ b/vignettes/adding-support-for-reporting.Rmd
@@ -31,19 +31,21 @@ The entire life cycle of objects involved in creating the report and configuring
 Let us consider an example module, based on the example module from `teal`:
 ```{r, message=FALSE}
 library(teal)
-teal_example_module <- function(label = "example teal module") {
+example_module <- function(label = "example teal module") {
   module(
     label,
     server = function(id, data) {
       moduleServer(id, function(input, output, session) {
+        ns <- session$ns
+        updateSelectInput(session, "dataname", choices = isolate(datanames(data())))
         output$text <- renderPrint(data()[[input$dataname]])
       })
     },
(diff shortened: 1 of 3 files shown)


vignettes/adding-support-for-reporting.Rmd
```diff
@@ -31,19 +31,21 @@ The entire life cycle of objects involved in creating the report and configuring
 Let us consider an example module, based on the example module from `teal`:
 ```{r, message=FALSE}
 library(teal)
-teal_example_module <- function(label = "example teal module") {
+example_module <- function(label = "example teal module") {
   module(
     label,
     server = function(id, data) {
+        ns <- session$ns
```
vedhav, line 39:
```suggestion
        checkmate::assert_class(data, "reactive")
        checkmate::assert_class(data(), "teal_data")
        ns <- session$ns
```

vedhav (approved):
LGTM, Just felt that adding data assertions would be good. More then checking for the data it communicates the data type to the users who stumble upon example_module which will motivate app developers to think about creating modules now since its the same object that they create for app ;)

vignettes/adding-support-for-reporting.Rmd
chlebowa: I decided to remove assertions from _this_ vignette because it concentrates on reporting and I wanted less clutter.
The examples in _Creating Custom Modules_ do have these assertions.

vedhav: Ah! please move them above moduleServer.

vedhav: Ah! please move them above moduleServer.

vedhav: Fair enough

PR #998, opened 2023-12-06
remove leftover `tdata` mentions

chlebowa:
Changes `tdata` to `teal_data` in a few places we missed.


gogonzo (approved):
👍 no `tdata` occurrences in other files than `tdata.R`

PR #1003, opened 2023-12-11
1000 fix empty selection in restore

### LINKED ISSUE #1000: [Bug]: Empty selection on saved snapshot cannot be restored
### What happened?

Snapshot with an empty selection on categorical variable will fail to load

### How to reproduce:

1. Run app below
2. Add filter variable "Species"
3. Unselect all species
4. Save snapshot
5. Download snapshot
6. Restore snapshot
    - Error occurs

### Sample code

```r
app <- init(
  data = teal_data(
    new_iris = iris,
    code = "new_iris <- iris"
  ),
  modules = modules(
    example_module()
  ),
  title = "App title"
)
if (interactive()) {
  shinyApp(app$ui, app$server)
}
#> Error in as.Date.default(slice[[field]]) : 
#>   do not know how to convert 'slice[[field]]' to class “Date”
```

### sessionInfo()

_No response_

### Relevant log output

_No response_

### Code of Conduct

### Contribution Guidelines

### Security Policy

m7pr:
Closes #1000 

Hey, I fixed the issue with disability to load a snapshot with empty `selected` field.
This filter can now be created, saved and uploaded.

However on an upload of such filter and selection of such filter the app goes back to empty filters. Is that intended?

### DIFF
--- a/R/teal_slices-store.R
+++ b/R/teal_slices-store.R
@@ -63,7 +63,7 @@ slices_restore <- function(file) {
   tss_json$slices <-
     lapply(tss_json$slices, function(slice) {
       for (field in c("selected", "choices")) {
-        if (!is.null(slice[[field]])) {
+        if (length(slice[[field]]) > 0) {
           date_partial_regex <- "^[0-9]{4}-[0-9]{2}-[0-9]{2}"
           time_stamp_regex <- paste0(date_partial_regex, "\\s[0-9]{2}:[0-9]{2}:[0-9]{2}\\sUTC$")
 
@@ -75,6 +75,8 @@ slices_restore <- function(file) {
             } else {
               slice[[field]]
             }
+        } else {
+          slice[[field]] <- NULL
         }
       }
       slice
--- a/R/teal_slices-store.R
+++ b/R/teal_slices-store.R
@@ -76,7 +76,7 @@ slices_restore <- function(file) {
               slice[[field]]
             }
         } else {
-          slice[[field]] <- NULL
+          slice[[field]] <- character(0)
         }
       }
       slice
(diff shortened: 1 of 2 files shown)


chlebowa:
`selected = NULL` should set all values, `selected = character(0)` should set none.

chlebowa:
`selected = NULL` should set all values, `selected = character(0)` should set none.

m7pr:
Ok, just changed to character(0) and tested it sets filters to empty states which filter out all observations

R/teal_slices-store.R
```diff
@@ -75,6 +75,13 @@ slices_restore <- function(file) {
             } else {
               slice[[field]]
             }
+        } else {
+          slice[[field]] <-
+            if (field == "selected") {
+              character(0)
+            } else {
+              NULL
+            }
```
averissimo, line 84: :heart: the simplicity of solution. 1 line suggestion, but feel free to ignore

```suggestion
          slice[[field]] <- switch(field, selected = character(0), NULL)
```

chlebowa: Don't you mean `slice[[field]] <- switch(field, "selected" = character(0), NULL)` :thinking:

averissimo (approved):
Great fix! :rocket: 

Solved the initial bug and tested a bunch of scenarios without being able to break.

The 1 line was broken into more :smiling_face_with_tear:

chlebowa:
So what about `"choices"`? If an app is started with `choices = character(0)` (for whatever reason) and then a snapshot is created, on upload it will be converted to `choices = NULL`, which means something different.

Should we worry about such a case?

m7pr:
Hey @chlebowa probably uncommon situation but I fixed the code so that it handles `character(0)` and `NULL`s separately. Also added 3 tests to verify that. Thanks for the input!

Issue #999, opened 2023-12-06
[Feature Request]: Set default datanames to `ls(data@env)` 


gogonzo:
### Feature description

We should set default datanames of `<teal_data>` object. Changes should be made in two places:
- init - when the `data` is `teal_data` we check datanames consistency [here](https://github.com/insightsengineering/teal/blob/e01b322c4b9dfd8011b813e33be5733d00562bef/R/init.R#L186)
- srv_teal_with_splash -  when data is `teal_data_module` we check data consistency [here](https://github.com/insightsengineering/teal/blob/e01b322c4b9dfd8011b813e33be5733d00562bef/R/module_teal_with_splash.R#L129) 

Instead of throwing when data is empty, we should warn in the console that we set it to defaults.

```r
if (length(datanames(<teal_data>) == 0) {
  datanames(<teal_data>) <- ls(<teal_data>@env)
}
```

averissimo:
This is what I meant with the idea of discarding the slot, while giving users the convenience of the `datanames<-` function

``` r
library(teal.code) # refactor branch
library(teal.data) # refactor branch

my_datanames <- \(obj) ls(obj@env)

`my_datanames<-` <- function(obj, value) {
  varnames <- setdiff(ls(obj@env), value)
  eval_code(
    obj,
    substitute(
      rm(list = varnames),
      list(varnames = varnames)
    )    
  )
}

td <- teal_data() |>
  within({
    rm_me <- 2
    IRIS <- iris
    cols <- 1:3
    MTCARS <- mtcars[, cols]
  })

td
#> ✅︎ verified teal_data object
#> <environment: 0x55bf06312028> [L]
#> Parent: <environment: package:teal.data>
#> Bindings:
#> • cols: <int> [L]
#> • MTCARS: <df[,3]> [L]
#> • IRIS: <df[,5]> [L]
#> • rm_me: <dbl> [L]
my_datanames(td) <- c("IRIS" , "MTCARS")
td
#> ✅︎ verified teal_data object
#> <environment: 0x55bf04c40168> [L]
#> Parent: <environment: package:teal.data>
#> Bindings:
#> • MTCARS: <df[,3]> [L]
#> • IRIS: <df[,5]> [L]
my_datanames(td)
#> [1] "IRIS"   "MTCARS"
get_code(td) |> cat()
#> rm_me <- 2
#> IRIS <- iris
#> cols <- 1:3
#> MTCARS <- mtcars[, cols]
[...]

PR #1004, opened 2023-12-11
999 set default datanames to ls(data@env)

kartikeyakirar:
this fixes https://github.com/insightsengineering/teal/issues/999

~~Here, I am assigning the default dataset of data types as "data.frame" and "MultiAssayExperiment", which can be altered to any other class type as needed.~~

~~Edit: added function `update_default_dataname`, is designed to retrieve and update default `datanames` from a `teal_data` object. Purpose of this function is to extract environment object from a `teal_data` object. If the object does not already have predefined dataset names, the function sets default names based on the environment.~~

Edit:  Based on the suggestion, the `init` function has been updated to avoid modifying the `teal_data` object. Instead, if `datanames` are not specified, it now passes the app through. Moreover, the `teal_data_to_filtered_data` function has been adjusted to handle cases where `datanames` are not pre-defined in the `teal_data` object. In such case, this function assigns default `datanames`  from the environment of the `teal_data` object.

~~Assertion~~ Different data types in `FilteredData `will be addressed in issue https://github.com/insightsengineering/teal.slice/issues/493


R/utils.R
```diff
@@ -215,3 +215,27 @@ check_filter_datanames <- function(filters, datanames) {
...
+#' @keywords internal
+update_default_dataname <- function(data) {
+  if (length(teal.data::datanames(data)) == 0) {
+  warning("`data` object has no datanames. Default datanames are specified from environment.")
+  datanames <- ls(data@env)
+  classes = sapply(datanames, function(dn) class(data[[dn]]), simplify = "array")
+  valid_datanames <- datanames[classes %in% c("data.frame", "MultiAssayExperiment")]
+  if(length(valid_datanames) == 0) {
+    stop("Data environment does not contains valid data class. Contact app developer.")
+  }
```
chlebowa: allow everything

R/utils.R
```diff
@@ -215,3 +215,27 @@ check_filter_datanames <- function(filters, datanames) {
...
+#' @keywords internal
+update_default_dataname <- function(data) {
+  if (length(teal.data::datanames(data)) == 0) {
+    warning("`data` object has no datanames. Default datanames are specified from environment.")
+    datanames <- ls(data@env)
+    classes <- sapply(datanames, function(dn) class(data[[dn]]), simplify = "array")
+    valid_datanames <- datanames[classes %in% c("data.frame", "MultiAssayExperiment")]
+    if (length(valid_datanames) == 0) {
+      stop("Data environment does not contains valid data class. Contact app developer.")
+    }
```
chlebowa, line 237: allow everything

R/utils.R
kartikeyakirar: functions as well?

R/utils.R
kartikeyakirar: In within some places user define function too. so Are we also permitting functions as dataname ?

chlebowa: everything

kartikeyakirar:
Assertion on FilteredDATA will be addressed in issue https://github.com/insightsengineering/teal.slice/issues/493

chlebowa:
Update initial comment.

pawelru:
I personally have some mixed feelings about this - not implementation per se but rather a design outlined in the issue. It's because those are _implicit_ modifications of the data object which are pretty much hidden behind the scenes for the end-user. As a general rule, _explicit_ is better than _implicit_.

As a direct consequence: `data` object passed to `init()` is different to the one being used which will make debugging and maintenance more difficult. Being super realistic now - it's not a big deal right now as the delta is relatively small but it's more about creating such precedence which is very likely to grow in the future.
I think it would be much better not to modify `data` object once it enters the init.

How about moving this into the constructor? This would address the use case of `teal_data(a = ..., b = ..., ...)` but not `teal_data() |> within(...)`. 

Related to this PR:
- https://github.com/insightsengineering/teal.data/discussions/211 - a wider discussion about `set_datanames()`
- https://github.com/insightsengineering/teal.code/issues/175 - a dedicated method for `ls(x@env)` which seems to be discouraged

chlebowa:
> How about moving this into the constructor? This would address the use case of `teal_data(a = ..., b = ..., ...)` but not `teal_data() |> within(...)`.

And that is the crux of the issue as we keep recommending using `within`.

I agree about explicit > implicit. BUT we are changing the philosophy a little here.
CURRENTLY `datanames` is a statement of which variables are datasets and therefore eligible for filtering.
SOON `datanames` will be a _subset_ of variables that will be _directed_ to filtering _if_ the `module` call takes `datanames = "all"`. If some `datanames` are specified in `module()`, we trust the app dev to name eligible datasets.

pawelru:
> > How about moving this into the constructor? This would address the use case of `teal_data(a = ..., b = ..., ...)` but not `teal_data() |> within(...)`.
> 
> And that is the crux of the issue as we keep recommending using `within`.

That's a valid point. Moving this to the constructor won't address the issue then...

> CURRENTLY `datanames` is a statement of which variables are datasets and therefore eligible for filtering. SOON `datanames` will be a _subset_ of variables that will be _directed_ to filtering _if_ the `module` call takes `datanames = "all"`. If some `datanames` are specified in `module()`, we trust the app dev to name eligible datasets.

Yes. I personally have some problems with this as I used to have a different meaning of datanames but it seems that I need to adapt and/or let's discuss it separately. Sticking with the definition you outlined in this thread for clarity.

So if one does not tell which variables from data object are eligible for filtering I think we shouldn't alter that object. Now I'm thinking that this should be an exception in filter panel init. I'm challenging the place / time we do this.

chlebowa:
At the moment the filter panel can only handle `data.frame` and `MultiAssayExperiment` and will crash if handed anything else. [Not for long](https://github.com/insightsengineering/teal.slice/issues/493).

As for the (`teal_data`) object being modified, note that it is disassembled when creating the filter panel and reassembled after filtering. The modules don't receive the entity that the app dev creates. By the `datanames` are no longer relevant.

pawelru:
> note that it is disassembled when creating the filter panel

I'm exactly suggesting doing this during that step. When disassembling we should throw a message and use all objects.

gogonzo:
> When disassembling we should throw a message and use all objects.

Consequence of this will be triggering each module on change of any dataset (even if dataset isn't used in a module).

> So if one does not tell which variables from data object are eligible for filtering I think we shouldn't alter that object. Now I'm thinking that this should be an exception in filter panel init. I'm challenging the place / time we do this.

`<teal_module>$datanames` indicates which datasets should be sent to the `<teal_module>$server` when initializing. Filtering is done automatically on these `datanames` - there is no way to say "i want these filtered" and "these received by the module$server". What you propose seems to be another argument of the `module()` function, or extension of the `teal_slices` api.

----- 

I would like to highlight importance of `datanames` and that `teal` actually handles situation when `datanames` are not known when the application starts. When all modules have `datanames = "all"` and `init(data = <teal_data_module>)` then `datanames` are not known. `datanames` are necessary during initialization of the modules and filter panel. 

Another scenario, if we consider everything in the `teal_data` as a `dataname` then FilteredData could be initialized with unnecessary objects, like [temporary objects](https://github.com/insightsengineering/teal.gallery/blob/b89806007beef055dcb301cd9334cbd967ae713b/efficacy/app.R#L17) or functions

gogonzo:
> Let me rephrase my point - we should avoid writting to data (and any other received arguments) inside init() as this is an implicit object mutation - we should only read from it.

This could be possible, there will be probably many `if`s in the whole teal.

----

 @kartikeyakirar I propose to initialize app with data but with empty `datanames()`, follow `data` object and see how this can be addressed in the alternative way (by doing `if (length(datanames(<teal_data>)) > 0)` `datanames(<teal data>)`  `else` `# names(ls(<tealdata>@env))`.

chlebowa:
> > CURRENTLY datanames is a statement of which variables are datasets and therefore eligible for filtering
> 
> > there is no way to say "i want these filtered"
> 
> These are somewhat conflicting statements.

To clarify: you cut off part of the latter statement

> here is no way to say "i want these filtered" and "these received by the module$server".

There is no way to say "filter these but pass these to the module".
"I want these filtered" is `modue(datanames)`, whereas `teal_data@datanames` is "these can be filtered".

kartikeyakirar:
> I propose to initialize app with data but with empty datanames(), follow data object and see how this can be addressed in the alternative way

@gogonzo The app can now be initialized with empty `datanames()`. When the app is opened, a notification will appear indicating that the `datanames` have not been specified. Additionally, the `teal_data_to_filtered_data` function has been modified to check if the `datanames` are pre-defined in the `teal_data` object. If the datanames are not defined, the function will assign default datanames from the environment of the `teal_data` object.

R/module_teal_with_splash.R
```diff
@@ -126,11 +126,15 @@ srv_teal_with_splash <- function(id, data, modules, filter = teal_slices()) {
...
-
-
       is_modules_ok <- check_modules_datanames(modules, teal.data::datanames(data))
-      validate(need(isTRUE(is_modules_ok), is_modules_ok))
+      if (!isTRUE(is_modules_ok)) {
+        showNotification(
+          "Object datanames is not specified; defaults will be applied from the environment",
+          type = "warning",
+          duration = 10
+        )
```
chlebowa, line 135: This notification will be shown to the app user but is meant for the app developer. The user knows nothing about `datanames`, either from from `teal_data` or the `modules` side.

R/utils.R
```diff
@@ -58,8 +58,16 @@ include_parent_datanames <- function(dataname, join_keys) {
 #' @keywords internal
 teal_data_to_filtered_data <- function(x, datanames = teal.data::datanames(x)) {
   checkmate::assert_class(x, "teal_data")
-  checkmate::assert_character(datanames, min.len = 1L, any.missing = FALSE)
-  checkmate::assert_subset(datanames, teal.data::datanames(x))
+  checkmate::assert_character(datanames, any.missing = FALSE)
+
+  if (length(datanames) == 0) {
+    datanames <- ls(x@env)
```
averissimo, line 64: Should we add an abstraction layer here instead of directly accessing the slot, neither `teal_data` nor `qenv` objects are defined in {teal}.

If this makes sense, then we could:
- register a S4 method for `teal.code::get_var("qenv", "missing")`
- register a new S4 method `teal.code::get_vars("qenv")`
- apply this only to `teal.data` with a new method `teal.data::get_vars("teal_data")`

R/module_teal_with_splash.R
kartikeyakirar: If the app user is not familiar with the datanames, then notifications are unnecessary, I will removed it.

R/utils.R
kartikeyakirar: there is feature request for this actually https://github.com/insightsengineering/teal.code/issues/175

kartikeyakirar: this will be modified once that is implemented

chlebowa: There is an issue for that somewhere.

R/utils.R
```diff
@@ -58,8 +58,16 @@ include_parent_datanames <- function(dataname, join_keys) {
...
+    if (length(datanames) == 0) {
+      stop("cannot assign datanames as `teal_data` environment is empty. Contact app developer.")
+    }
```
chlebowa, line 67: 1. A `stop` on empty `teal_data` should be triggered much earlier, in `init` probably. This is not the place.

pawelru: Re 1: I would argue that it is the right place with the rationale of keeping the logic encapsulated within a given function. If we move this stop condition elsewhere we will spread this specific logic across multiple functions which is a bad code smell for me. The responsibility of `teal_data_to_filtered_data` is to create one object instance on a basis of another and this stop statement is an information that this cannot be done because of some reason - a specific condition for a specific task. For me this looks good.
On the other hand I do understand a fail-fast that you probably thinking of but in this case (of specific condition for a specific action) I would personally value more better code organization.

chlebowa: No that `init` does this at the moment:
```
    if (length(teal.data::datanames(data)) == 0) {
      stop("`data` object has no datanames. Specify `datanames(data)` and try again.")
    }
```

I would argue that if an empty object cannot be used to build an app, it should fail immediately, after several processing steps.

The purpose of `teal_data_to_filtered_data` is to take a `teal_data` and turn it into a `filtered_data`. Setting `datanames` on that `teal_data` is not in its purview.

The coming changes to `teal` and `teal.slice` will remove the need for this `stop` anyway but that's another story.

kartikeyakirar:
```
 if (length(teal.data::datanames(data)) == 0) {
      stop("`data` object has no datanames. Specify `datanames(data)` and try again.")
    }

```
I added an additional condition to check for environment and restored above code. This is to ensure that the app initialization fails if an empty object is passed. Previously, I intended to minimize these checks and keep them in one place. However, I now favor failing the app on initialization for an empty object. committed change here [6866652](https://github.com/insightsengineering/teal/pull/1004/commits/6866652edddbedbfaa3d7fb928ce791aae5e9462).

pawelru: > No that `init` does this at the moment:
> 
> ```
>     if (length(teal.data::datanames(data)) == 0) {
>       stop("`data` object has no datanames. Specify `datanames(data)` and try again.")
>     }
> ```

I have a similar objection to this and probably more in there but that's a separate issue out of scope of this PR.

> I would argue that if an empty object cannot be used to build an app, it should fail immediately, after several processing steps.

You are risking distributing logic across many functions. 
Let's take a following dummy example:
```
foo <- function(a, b) {
  aa <- a_handler(a)
  bb <- b_handler(b)
  bar(aa, bb)
}
```
If I understand you right you are suggesting the following
```
foo <- function(a, b) {
  stopifnot(check1(a))
  stopifnot(check2(a))
  ...
  (same for `b`)
  (body)
}
```
Then you have part of the logic for `a` in `foo` and the other part in `a_handler`. I am thinking to have it into `a_handler` instead. Both are fully functional but differs from code organization perspective. Following single responsibility paradigm - (broadly speaking) I want to change `a_handler` only on `a` change.

PS. I am actually seeing `teal_data_to_filtered_data` as the _main_ handler of `teal_data` class of object.

I think that the good middle ground is something we used to do: to perform _just_ class checks upfront in the top-level function to fail fast but don't put any specific logic outside of handlers.

I am raising this to your attention as recent refactor (in particular: the need of modify ~a lot of places~ almost everything) just exposed previous mistakes. Obviously those were big changes but still it should had been more condensed.

> The purpose of `teal_data_to_filtered_data` is to take a `teal_data` and turn it into a `filtered_data`. Setting `datanames` on that `teal_data` is not in its purview.

Yes but I'm seeing this more of a validating than setting (even though formal set is called).

> The coming changes to `teal` and `teal.slice` will remove the need for this `stop` anyway but that's another story.

Ok then I'm not pushing anymore on this particular case but I really want you to understand the rationale not to repeat previous mistakes.

chlebowa: > PS. I am actually seeing `teal_data_to_filtered_data` as the _main_ handler of `teal_data` class of object.

Ah, that is where we fundamentally differ. I think we are unlikely to hammer out a consensus in this form.

> I am raising this to your attention as recent refactor (in particular: the need of modify ~a lot of places~ almost everything) just exposed previous mistakes. Obviously those were big changes but still it should had been more condensed.

We have cut some heads off the this hydra but the battle is not over yet :wink: And since we are changing everything incrementally, we have transition stages. I think this is one.

As for moving a `stop` from one place to another, it does seem out of the scope of this issue. I hate to be a broken record but we have had too much of that recently.

R/utils.R
```diff
@@ -58,8 +58,13 @@ include_parent_datanames <- function(dataname, join_keys) {
 #' @keywords internal
 teal_data_to_filtered_data <- function(x, datanames = teal.data::datanames(x)) {
   checkmate::assert_class(x, "teal_data")
-  checkmate::assert_character(datanames, min.len = 1L, any.missing = FALSE)
-  checkmate::assert_subset(datanames, teal.data::datanames(x))
+  checkmate::assert_character(datanames, any.missing = FALSE)
+
+  if (length(datanames) == 0) {
```
gogonzo, line 63: @kartikeyakirar I think you need a new internal teal function to extract datanames from `teal_data` object

```r
#' Wrapper on `teal.data::datanames`
#'
#' Special function used in internals of `teal` to return names of datasets even if `datanames` 
#' has not been set. 
#' @param data (`teal_data`)
#' @return `character`
# @keywords internal
teal_data_datanames <- function(data) {
  if (length(teal.data::datanames(data))) {
    teal.data::datanames(data)
  } else {
    warning("`data` object has no datanames. Default datanames are set using `teal_data`'s environment.")
    ls(data@env)
  }
}
```

This function should probably replace all `teal.data::datanames()` calls in `teal`. At this moment solution is insufficient as most of the asserts will fail (please check teal.gallery/exploratory app for example).

kartikeyakirar: replacing` teal.data::datanames` call  for` teal`  this will probably gives many warnings from different places.

gogonzo: > replacing teal.data::datanames call for teal this will probably gives many warnings from different places.

Let's remove warning from the function and warn once in `srv_teal_with_splash` just before `is_modules_ok <- ...`:

```r
if (!length(teal.data::datanames(data))) {
  warning("`data` object has no datanames. Default datanames are set using `teal_data`'s environment.")
}

```

kartikeyakirar: added suggested changes here [3504669](https://github.com/insightsengineering/teal/pull/1004/commits/3504669b005a2ebdd2687259072d7e9aae587b59)

R/module_teal.R
```diff
@@ -189,7 +189,7 @@ srv_teal <- function(id, modules, teal_data_rv, filter = teal_slices()) {
           # null controls a display of filter panel but data should be still passed
           datanames <- if (is.null(modules$datanames) || modules$datanames == "all") {
             include_parent_datanames(
-              teal.data::datanames(teal_data_rv()),
+              teal_data_datanames(teal_data_rv()),
               teal_data_rv()@join_keys
```
chlebowa, line 193:
```suggestion
              teal.data::join_keys(teal_data_rv())
```

R/utils.R
```diff
@@ -215,3 +214,19 @@ check_filter_datanames <- function(filters, datanames) {
...
+#' has not been set.
+#' @param data (`teal_data`)
+#' @return `character`
+# @keywords internal
+teal_data_datanames <- function(data) {
+  checkmate::assert_class(data, "teal_data")
+  if (length(teal.data::datanames(data))) {
+    teal.data::datanames(data)
+  } else {
+    ls(data@env)
```
chlebowa, line 230:
```suggestion
    names(data@env)
```

R/init.R
```diff
@@ -183,19 +183,17 @@ init <- function(data,
...
+      stop("`data` object has no datanames and its environment is empty. Specify `datanames(data)` and try again.")
     }
-
     # in case of teal_data_module this check is postponed to the srv_teal_with_splash
-    is_modules_ok <- check_modules_datanames(modules, teal.data::datanames(data))
+    is_modules_ok <- check_modules_datanames(modules, teal_data_datanames(data))
     if (!isTRUE(is_modules_ok)) {
       logger::log_error(is_modules_ok)
-      checkmate::assert(is_modules_ok, .var.name = "modules")
+      # we allow app to continue if datanames are not available.
```
gogonzo, line 193: when using `teal_data_datanames` we should stop the app. I don't necessary mean to crash the app but we can't continue loading modules as there are insufficient datanames. 
```suggestion
      checkmate::assert(is_modules_ok, .var.name = "modules") 
      # or
      validate(need(FALSE, is_modules_ok))
```

tests/testthat/test-utils.R
```diff
@@ -36,3 +36,23 @@ testthat::test_that("report_card_template function returns TealReportCard object
...
+  teal_data <- teal.data::teal_data()
+  teal_data <- within(teal_data, {
+    iris <- head(iris)
+    mtcars <- head(mtcars)
+  })
+  testthat::expect_equal(teal_data_datanames(teal_data), c("iris", "mtcars"))
+
+  datanames(teal_data) <- "iris"
+  testthat::expect_equal(teal_data_datanames(teal_data), "iris")
+})
```
gogonzo, line 58: Please always set `test_that(desc=)` with meaningful and precise description. `teal_data_datanames return datanames` is ambiguous. 

```suggestion
test_that("test_that("teal_data_datanames returns names of the @env's objects when datanames not set", {
  teal_data <- teal.data::teal_data()
  teal_data <- within(teal_data, {
    iris <- head(iris)
    mtcars <- head(mtcars)
  })
  testthat::expect_equal(teal_data_datanames(teal_data), c("iris", "mtcars"))
})

test_that("teal_data_datanames returns datanames which are set by teal.data::datanames", 
  teal_data <- teal.data::teal_data()
  teal_data <- within(teal_data, {
    iris <- head(iris)
    mtcars <- head(mtcars)
  })
  datanames(teal_data) <- "iris"
  testthat::expect_equal(teal_data_datanames(teal_data), "iris")
})
```

kartikeyakirar:
When passing `datanames` from the environment to some modules in teal.gallery apps, some issues are observed. e.g  `datanames` are displayed in the `tm_data_table`.
image . I tried to fix this by removing the assertion on `validate_has_data` (its not solution)

also  if have` function` defined in` within` it fails the app
image

Do we need to modify modules as well to filter out valid data to be displayed ?

chlebowa:
> When passing `datanames` from the environment to some modules in teal.gallery apps, some issues are observed. e.g `datanames` are displayed in the `tm_data_table`.

You have to modify the app by setting the modules's `datanames`, I assume it's `"all"` now.

kartikeyakirar:
> You have to modify the app by setting the modules's datanames, I assume it's "all" now.

yes, I was experimenting with teal.gallery app by not setting datanames and running app. I realized this is app-developer responsibility to set correct `datanames`.By default, the data names are set to `ALL`. However, I assumed that when modules were built, the data names were always data frames or multi-assay experiment classes. But now, it can be a function or variable, so that's why I wondered if we needed to modify the modules considering possibilities that if `All` is set then instead of failing the app how to handle the different datatypes.

"I was experimenting with the teal.gallery app without setting data names, and realized that it is the app developer's responsibility to set correct data names. By default, the data names are set to `ALL`. However, I assumed that when modules were built, the data names were always data frames or multi-assay experiment classes. But now, it can be a function or variable, so I wondered if we needed to modify the modules as well."

gogonzo:
> yes, I was experimenting with teal.gallery app by not setting datanames and running app. I realized this is app-developer responsibility to set correct `datanames`.By default, the data names are set to `ALL`. However, I assumed that when modules were built, the data names were always data frames or multi-assay experiment classes. But now, it can be a function or variable, so that's why I wondered if we needed to modify the modules considering possibilities that if `All` is set then instead of failing the app how to handle the different datatypes.

Yes, it is a responsibility of a module to skip unsupported datasets or app developer to correctly specify `module$datanames`. Even now, teal.modules.general::tm_data_table will fail with `MultiAssayExperiment`.

chlebowa:
> It should be okey with this [PR](https://github.com/insightsengineering/teal.slice/pull/497)

This will still happen.

image

kartikeyakirar:
> It should be okey with this https://github.com/insightsengineering/teal.slice/pull/497

I am running app by loading teal.slice from https://github.com/insightsengineering/teal.slice/pull/497

R/init.R
```diff
@@ -183,19 +183,17 @@ init <- function(data,
...
+      stop("`data` object has no datanames and its environment is empty. Specify `datanames(data)` and try again.")
     }
-
     # in case of teal_data_module this check is postponed to the srv_teal_with_splash
-    is_modules_ok <- check_modules_datanames(modules, teal.data::datanames(data))
+    is_modules_ok <- check_modules_datanames(modules, teal_data_datanames(data))
     if (!isTRUE(is_modules_ok)) {
       logger::log_error(is_modules_ok)
-      checkmate::assert(is_modules_ok, .var.name = "modules")
+      validate(need(isTRUE(is_modules_ok), is_modules_ok))
```
gogonzo, line 193:
```suggestion
      validate(need(isTRUE(is_modules_ok), sprintf("%s. Contact app developer.", is_modules_ok)))
```
Skärmavbild 2023-12-14 kl  13 11 39

gogonzo, line 193: init is not a shiny module, you need to hardstop here
```suggestion
      validate(need(isTRUE(is_modules_ok), is_modules_ok))
```

R/module_teal_with_splash.R
```diff
@@ -126,13 +126,16 @@ srv_teal_with_splash <- function(id, data, modules, filter = teal_slices()) {
...
-
+      if (!length(teal.data::datanames(data))) {
+        warning("`data` object has no datanames. Default datanames are set using `teal_data`'s environment.")
+      }
 
-      is_modules_ok <- check_modules_datanames(modules, teal.data::datanames(data))
-      validate(need(isTRUE(is_modules_ok), is_modules_ok))
+      is_modules_ok <- check_modules_datanames(modules, teal_data_datanames(data))
+      if (!isTRUE(is_modules_ok)) {
+        logger::log_warn(is_modules_ok)
```
gogonzo, line 135:
```suggestion
        logger::log_warn(is_modules_ok)
        validate(need(isTRUE(is_modules_ok), sprintf("%s. Contact app developer.", is_modules_ok)))
```

Skärmavbild 2023-12-14 kl  13 11 39

R/init.R
```diff
@@ -183,19 +183,17 @@ init <- function(data,
   }
 
   if (inherits(data, "teal_data")) {
-    if (length(teal.data::datanames(data)) == 0) {
-      stop("`data` object has no datanames. Specify `datanames(data)` and try again.")
+    if (length(teal.data::datanames(data)) == 0 && length(names(data@env)) == 0) {
```
gogonzo, line 186:
```suggestion
    if (length(teal.data_datanames(data)) == 0) {
```

R/init.R
kartikeyakirar: you mean `checkmate::assert(is_modules_ok, .var.name = "modules")`

gogonzo: Yes

gogonzo:
Not setting `datanames` in the `teal_data` object causes problems. `get_code(datanames)` asserts on `datanames`. Can we relax `get_code` to allow any "name"? 
What do you think @m7pr ?

```r
data <- teal_data() |>
  within({
    set.seed(1)
    library(scda)
    library(airway)
    library(MultiAssayExperiment)
    data(miniACC, envir = environment())
    data(airway, envir = environment())
    any_variable <- "elo"
    ADSL <- synthetic_cdisc_data("latest")$adsl
    ADSL$categorical <- sample(letters[1:3], size = nrow(ADSL), replace = TRUE, prob = c(.1, .3, .6))
    ADTTE <- synthetic_cdisc_data("latest")$adtte
    ADRS <- synthetic_cdisc_data("latest")$adrs
  })

get_code(data, datanames = "ADRS")
# Error in .local(object, deparse, ...) : 
#   Assertion on 'datanames' failed: Must be a subset of the empty set, i.e. also empty.
```

chlebowa:
Is this a problem in the app? By the time we get to `get_code`, `datanames` have been set by `teal`.

gogonzo (approved):
> Is this a problem in the app? By the time we get to get_code, datanames have been set by teal.

It is not a Today's problem. We use `get_code_dependency` in `.datasets_to_data` so we are not limited by the fact that `@datanames` of `data` object is not set. 

Normally, if we use `get_code(datanames=)` we wouldn't be able to run `get_code` as in `teal` we avoid to set default `datanames(data) <- names(data@env)` when `datanames(data)` is empty. If we decide to use `get_code()` in the future, we will be forced to either:
1.  Relax `get_code(datanames=)` to accept any name, or
2. Force-set default `datanames` to `data` object in early `teal` stages.

----

This is not ideal but I don't request @kartikeyakirar to resolve all of this now. At this moment PR addresses what we need.

gogonzo:
@kartikeyakirar could you please add `@details` section in `init` documentation explaining `datanames` process?

R/utils.R
```diff
@@ -215,3 +214,19 @@ check_filter_datanames <- function(filters, datanames) {
...
+#' has not been set.
+#' @param data (`teal_data`)
+#' @return `character`
+#' @keywords internal
+teal_data_datanames <- function(data) {
+  checkmate::assert_class(data, "teal_data")
+  if (length(teal.data::datanames(data))) {
+    teal.data::datanames(data)
+  } else {
+    names(data@env)
```
m7pr, line 230: what about `ls(teal.code::get_env(data)` once this is merged https://github.com/insightsengineering/teal.code/pull/187

m7pr:
About `datanames` assertion for `get_code,teal_data` https://github.com/insightsengineering/teal.data/blob/main/R/teal_data-get_code.R#L39-L40 I would prefer to `Force-set default datanames to data object in early teal stages.` so that `get_code` works only for `teal_data` objects with proper specification (non empty `@code` slot and `datanames` attribute)

R/utils.R
chlebowa: `names` keeps hidden items.

m7pr: `ls(teal.code::get_env(data), all.names = TRUE)` : P ?

kartikeyakirar:
> @kartikeyakirar could you please add @details section in init documentation explaining datanames process?

done https://github.com/insightsengineering/teal/pull/1004/commits/f611e5549a6456c883f53995dc992e61ae3e3035

PR #1008, opened 2023-12-13
align args asserts; align args defaults; logger in test

pawelru:
ready for review

Unfortunately it grew more than I expected initially and I should have been split it into few for the review convenience...

- align argument assertions
  Currently on main we have two approaches for assert and convert. Please find below examples (note the order)
  - `modules`:
    - assert all classes: https://github.com/insightsengineering/teal/blob/e98eab0a2465a00e05957cefc408ee10185440aa/R/init.R#L121
    - convert to the desired: https://github.com/insightsengineering/teal/blob/e98eab0a2465a00e05957cefc408ee10185440aa/R/init.R#L133-L138
  - `data`:
    - convert to the desired: https://github.com/insightsengineering/teal/blob/e98eab0a2465a00e05957cefc408ee10185440aa/R/init.R#L105-L108
    - assert classes: https://github.com/insightsengineering/teal/blob/e98eab0a2465a00e05957cefc408ee10185440aa/R/init.R#L120
I decided to follow the latter one. 
- align argument defaults for `header`, `footer`, etc.
- silent logger in tests to avoid the following:
```
✔ |          7 | rcode_utils                                             
⠏ |          0 | report_previewer_module                                 [INFO] 2023-12-13 16:14:23.5635 pid:41602 token:[] teal Initializing reporter_previewer_module
[INFO] 2023-12-13 16:14:23.5663 pid:41602 token:[] teal Initializing reporter_previewer_module
✔ |          5 | report_previewer_module
```
- change error message from "The teal_data_module must ..." into "The data module must ..." as end users don't have to know the name of the class. They are interacting with function arguments like `data` or `modules`.
- I noticed that a few tests expects a string with error / warning message that comes from logger. This is wrong as this actually test logger functionality and assumes certain value of logger threshold, appender, layout etc. I changed this to the base warning that is independent on all of those things. Will create a separate task to discuss handler via logger - https://github.com/insightsengineering/teal/issues/1028


tests/testthat/setup-logger.R
```diff
@@ -0,0 +1 @@
+logger::log_threshold(logger::FATAL, namespace = "teal")
```
pawelru, line 1: I actually don't know how to turn off the logger. FATAL is just temporary that is sufficient for now.

Available options are:
- custom layout to be an empty string
- custom appender function to be NULL
- ...

R/init.R
```diff
@@ -134,40 +146,37 @@ init <- function(data,
...
   checkmate::assert_character(id, max.len = 1, any.missing = FALSE)
 
-  teal.logger::log_system_info()
-
   if (is.character(title)) {
     title <- build_app_title(title)
-  } else {
-    validate_app_title_tag(title)
   }
+  validate_app_title_tag(title)
```
pawelru, line 152: changed this to be unconditional

R/init.R
```diff
@@ -237,13 +241,3 @@ init <- function(data,
...
-
-create_app_id <- function(data, modules) {
-  hashables <- c(data, modules)
-  hashables$data <- if (inherits(hashables$data, "teal_data")) {
-    as.list(hashables$data@env)
-  } else if (inherits(data, "teal_data_module")) {
-    body(data$server)
-  }
-  rlang::hash(hashables)
-}
```
chlebowa: I don't see this block on `main` :thinking:

pawelru: IDK why it's highlighted in `R/init.R` as removed :/

I moved respective code to the function and placed it in the utils script.

pawelru:
This still needs some work - back to the draft.

I need to clear up the way how title is handled:
- align the defaults with other ui_xyz functions
- will probably move it to the `ui_teal` where we similarily handle `header` and `footer` arguments
- possibly encapsulate header and footer as a seperate function - to be consistent with build_title functionality
- maybe other minor stuff as I see the need

R/validate_inputs.R
```diff
@@ -152,7 +152,7 @@ gather_messages <- function(iv) {
     failing_inputs <- Filter(Negate(is.null), status)
     unique(lapply(failing_inputs, function(x) x[["message"]]))
   } else {
-    logger::log_warn("Validator is disabled and will be omitted.")
```
chlebowa, line 155: This was intended to explicitly send this message to the logger rather than the console. The respective unit tests accounted for that.

pawelru: Yes I noticed this and I feel it's not very correct. 

If you put an expectation on the string, you are actually testing logger functionality. It's also dependent on logger threshold, layout formatter, appender function etc, e.g. `"\\[WARN\\].+Validator is disabled and will be omitted."` instead of just `"Validator is disabled and will be omitted."`.
Our tests should be more independent on logger OR if we explicitly test logger then we should pre-set all of those options (testthat::with_option).
When thinking about this I came across a promising logger functionality that could address this issue nicely. I opened a separate issue for a wider discussion: https://github.com/insightsengineering/teal/issues/1028

R/init.R
```diff
@@ -122,52 +121,59 @@ init <- function(data,
...
+  # argument mutation
+  ## `modules` - landing module
   landing <- extract_module(modules, "teal_module_landing")
-  if (length(landing) > 1L) stop("Only one `landing_popup_module` can be used.")
+  landing_module <- NULL
+  if (length(landing) > 1L) {
+    stop("Only one `landing_popup_module` can be used.")
+  } else if (length(landing) == 1) {
+    landing_module <- landing[[1L]]
+  }
```
chlebowa, line 162: I disagree with this change. You want to complete checks in one place, I get it but the check at the place of execution cannot be avoided. Your offset is you are effectively adding one more check on the result of `extract_module`, you introduce a "declare and modify" on `landing_module` here _and_ you remove a variable, which is honestly something I have never seen in a function body before.

R/init.R
```diff
@@ -122,52 +121,59 @@ init <- function(data,
...
   modules <- drop_module(modules, "teal_module_landing")
+  rm(landing)
 
+  ## `filter` - app_id attribute
   # Calculate app id that will be used to stamp filter state snapshots.
   # App id is a hash of the app's data and modules.
   # See "transferring snapshots" section in ?snapshot.
```
chlebowa, line 169: Since this is delegated to a function, these comments will be more appropriate in that function's documentation.
```suggestion
```

R/utils.R
```diff
@@ -271,3 +271,20 @@ build_app_title <- function(title = "Teal app", favicon = "https://raw.githubuse
...
   )
 }
+
+#' Creates App ID as a hash of the input data and modules.
+#'
+#' @param data data object
+#' @param modules modules object
+#'
+#' @return character(1) with hash
+#' @keywords internal
```
chlebowa, line 281:
```suggestion
#' Application ID
#' 
#' Creates App ID used to match filter snapshots to application.
#'
#' Calculate app id that will be used to stamp filter state snapshots.
#' App id is a hash of the app's data and modules.
#' See "transferring snapshots" section in ?snapshot.
#' 
#' @param data `teal_data` or `teal_data_module` as accepted by `init`
#' @param modules `teal_modules` object as accepted by `init`
#'
#' @return A single character string.
#'
#' @keywords internal
```

R/utils.R
```diff
@@ -271,3 +271,20 @@ build_app_title <- function(title = "Teal app", favicon = "https://raw.githubuse
...
+create_app_id <- function(data, modules) {
```
chlebowa, line 282: Since this is a separated function now, please add argument checks.

R/init.R
```diff
@@ -122,52 +121,59 @@ init <- function(data,
...
   checkmate::assert(
     checkmate::check_class(filter, "teal_slices"),
     checkmate::check_list(filter, names = "named")
   )
+
+  ## all other arguments
   checkmate::assert_multi_class(title, c("shiny.tag", "character"))
   checkmate::assert_multi_class(header, c("shiny.tag", "character"))
   checkmate::assert_multi_class(footer, c("shiny.tag", "character"))
   checkmate::assert_character(id, max.len = 1, any.missing = FALSE)
```
chlebowa, line 149: Shouldn't this be `assert_string`?

R/module_teal_with_splash.R
```diff
@@ -72,7 +80,7 @@ srv_teal_with_splash <- function(id, data, modules, filter = teal_slices()) {
     teal_data_rv <- if (inherits(data, "teal_data_module")) {
       data <- data$server(id = "teal_data_module")
       if (!is.reactive(data)) {
-        stop("The `teal_data_module` must return a reactive expression.", call. = FALSE)
+        stop("The `data` module must return a reactive expression.", call. = FALSE)
```
chlebowa, line 83: `teal_data_module` is correct. We don't have a "data module" concept in `teal` ATM. This error message is for the app developer, who is expected to know what a `teal_data_module` is, since it is only the app developer that can pass one to `init`.

Likewise the messages below, they are to point the app developer in the right direction and need not necessarily be readily understood by the app user.

chlebowa (changes requested):
> align argument assertions

I believe asserting and converting is superior to converting and asserting, see specific comment below.

----

> align argument defaults for header, footer, etc.

While we are here, the description of `header` and is borderline confusing:
```
#' @param header (`shiny.tag` or `character`) \cr
#'   the header of the app. Note shiny code placed here (and in the footer
#'   argument) will be placed in the app's `ui` function so code which needs to be placed in the `ui` function
#'   (such as loading `CSS` via [htmltools::htmlDependency()]) should be included here.
```

Could this be improved somewhat?

Also consider combining `header` and `footer` 
```
#' @parem header,footer (`shiny.tag` or `character`) \cr
#'   the header and footer of the app (...)
```

----

Lastly, I would also consider changing the order of the arguments by putting `filter` before `title` but that may be too serious a change.

(Withholding logger discussion until the simple things are agreed upon.)

R/init.R
pawelru: There is a difference for `character(0)` - string will raise whereas character with max.len = 1 will not. `assert_string` does not have appropriate argument to allow `character(0)`. Looking at the argument default value (i.e. `character(0)`) I think this needs to stay as is

R/init.R
```diff
@@ -122,52 +121,59 @@ init <- function(data,
       )
     )
   }
-
+  if (is.list(data) && !inherits(data, "teal_data_module")) {
```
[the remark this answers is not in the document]
pawelru: https://github.com/insightsengineering/teal/issues/1031#issue-2072037423

R/init.R
```diff
@@ -122,52 +121,59 @@ init <- function(data,
...
-  checkmate::assert_multi_class(modules, c("teal_module", "list", "teal_modules"))
+
+  ## `modules`
+  if (inherits(modules, "teal_module")) {
+    modules <- list(modules)
+  }
+  if (inherits(modules, "list")) {
+    modules <- do.call(teal::modules, modules)
+  }
+  checkmate::assert_class(modules, "teal_modules")
```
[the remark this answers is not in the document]
pawelru: > Changing checkmate::assert_multi_class(modules, c("teal_module", "list", "teal_modules")) pre-normalization to checkmate::assert_class(modules, "teal_modules") post-normalization removes the gate on allowed argument types.

Hmm I don't follow. Could you please elaborate more?

My rationale to do convert - assert (as opposed to assert - convert) is that we can also include the results of convert in the assertion that follows (note that we dont do assert on return inside convert).

> Additionally, if modules is list, it wouldn't hurt to check if elements are teal_module(s).

I was thinking about this and this is already covered in here: 
https://github.com/insightsengineering/teal/blob/few_enhancements/R/modules.R#L78
```
checkmate::assert_list(submodules, min.len = 1, any.missing = FALSE, types = c("teal_module", "teal_modules"))
```

R/init.R
pawelru: > but the check at the place of execution cannot be avoided.

Are you referring to this (code w/o my change)?

```
      if (length(landing) == 1L) {
        landing_module <- landing[[1L]]
      }
```

I think it's the same as:

```
if (!is.null(landing)) { ... }
```

(or in the version with changes)
```
if (!is.null(landing_module)) { ... }
```

because we also have:

```
if (length(landing) > 1L) stop("Only one `landing_popup_module` can be used.")
```

The above practically means that we can only have 0 or 1 landing module and 0 is represented by NULL value here.

---

The whole point of my change is that I want to move a following line `landing_module <- landing[[1L]]` upper to keep all the logic together. Right now we have it a little bit distributed - _extract_ and _stop_ in one place and _use first el_ in the other.

In the light of the above - I introduced a second variable to keep interpretation of the return value from extract() next to the actual call. Currently we have `length(landing)>1` in one place and `length(landing) == 1L` together with `landing[[1]]` in the other.

---

I added a `rm()` call not to be confused during debugging which variable means what. Probably better to put better names instead. I am thinking now of having `landing_modules` (return from the extract()) and `landing_module` (as the first element of the former). WDYT?

R/init.R
chlebowa: Oh, sorry, I neglected to look at the default.

R/init.R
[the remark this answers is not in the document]
chlebowa: > > Changing checkmate::assert_multi_class(modules, c("teal_module", "list", "teal_modules")) pre-normalization to checkmate::assert_class(modules, "teal_modules") post-normalization removes the gate on allowed argument types.
> 
> Hmm I don't follow. Could you please elaborate more?

I prefer checking what is passed into the function so that we can fail specifically rather than attempt a conversion from anything and then fail on the attempt. One could pass a `FilteredData` as modules (because why not?) and `do.call(teal::modules, modules)` would fail miserably.

R/init.R
chlebowa: > > but the check at the place of execution cannot be avoided.
> 
> Are you referring to this (code w/o my change)?

Yes. You have to check down there anyway, so creating a NULL just to test for NULL instead of length>1 is redundant.


> The above practically means that we can only have 0 or 1 landing module and 0 is represented by NULL value here.

Yes. Your logic does not improve the current one, merely substitutes the length>1 condition for is.null.

> The whole point of my change is that I want to move a following line `landing_module <- landing[[1L]]` upper to keep all the logic together. Right now we have it a little bit distributed - _extract_ and _stop_ in one place and _use first el_ in the other.
> 
> In the light of the above - I introduced a second variable to keep interpretation of the return value from extract() next to the actual call. Currently we have `length(landing)>1` in one place and `length(landing) == 1L` together with `landing[[1]]` in the other.

You are restoring the logic that was proposed and rejected during review when this change was introduced.

R/init.R
```diff
@@ -122,52 +121,59 @@ init <- function(data,
...
+
+  ## `filter`
   checkmate::assert(
     checkmate::check_class(filter, "teal_slices"),
     checkmate::check_list(filter, names = "named")
```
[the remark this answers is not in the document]
pawelru: At first I just did it but then I noticed that similar lines are in `teal.slice` as well. I Think it should be a separate task to be completed across multiple repos.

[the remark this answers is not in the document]
pawelru: image

R/init.R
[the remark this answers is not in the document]
pawelru: > rather than attempt a conversion from anything and then fail on the attempt

We don't try on anything - note that each conversion is conditional. The goal is to have the following:

```
if (inherits(x, "a")) convert_a_to_x(x)
if (inherits(x, "b")) convert_b_to_x(x)
assert_class(x, "x")
```

But you have discovered one thing - R6 objects inherit from list and `(inherits(modules, "list"))` condition is imperfect. How about changing this into `if (checkmate::test_list(modules, types = c("teal_module", "teal_modules"))) {`?

> One could pass a FilteredData as modules (because why not?) and do.call(teal::modules, modules) would fail miserably.

Great example and love this way of thinking. The goal is that this won't enter any conversion and fall directly into the assertion. That's not the case currently but will be true after proposal.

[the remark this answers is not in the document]
averissimo: I favor the `assert ➟ convert` workflow when we can group the assertions on the same argument together (instead of having 2+ assertions as `data` has).

If we can protect against all allowable types, then we can "safely" convert those. Although I see a point in making sure that in the end, we have what is expected.

<details>

<summary>Sample</summary>

```r
## `data`
checkmate::assert(
  combine = "or",
  .var.name = "data",
  checkmate::check_multi_class(data, c("teal_data", "teal_data_module")),
  checkmate::check_list(data, names = "named")
)

# then convert

## `modules`

checkmate::assert(
  combine = "or",
  .var.name = "modules",
  checkmate::check_multi_class(modules, c("teal_module", "teal_modules")),
  checkmate::check_list(modules, types = c("teal_module", "teal_modules"))
)

# then convert
```

</details>

> But you have discovered one thing - R6 objects inherit from list and (inherits(modules, "list")) condition is imperfect

``` r
Sample <- R6::R6Class("Sample", public = list(sample = function(...) sample(...)))
inherits(Sample$new(), "list")
#> [1] FALSE

fd <- teal.slice:::FilteredData$new(list(iris = iris))
inherits(fd, "list")
#> [1] FALSE
```

R/init.R
```diff
@@ -122,52 +119,56 @@ init <- function(data,
...
+    modules <- list(modules)
+  }
+  if (inherits(modules, "list")) {
+    modules <- do.call(teal::modules, modules)
+  }
+  checkmate::assert_class(modules, "teal_modules")
+
+  ## `filter`
   checkmate::assert(
     checkmate::check_class(filter, "teal_slices"),
```
averissimo, line 139: Nitpicking, but consider manually adding  `.var.name` when performing a generic `assert` on the same variable. (individual messages have it though)

image

```suggestion
    .var.name = "filter",
    checkmate::check_class(filter, "teal_slices"),
```

R/init.R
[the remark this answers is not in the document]
pawelru: Right - true. I was wrong with my statement that R6 instances inherits from list. Then the example of FilteredData as data should behave correctly and no changes needed.

R/init.R
pawelru: > so creating a NULL just to test for NULL instead of length>1 is redundant.

As pointed out in my previous message is rather to keep interpretation of extract() return value closer to the extract() function call. Let's imagine a refactor of extract() to return list() instead of NULL in case of no landing module - it would be much easier perform such refactor if both _execute_ and _use_ returned value is close together. My aim is to have pythonic `landing_module, modules <- extract_landing(modules)` where everything is in one line. Unfortunately it's not possible in R and we have to split it in multiple calls. I have no problem with that but I do have if those calls are spread across the code which I am trying to fix.

> Yes. Your logic does not improve the current one, merely substitutes the length>1 condition for is.null.

I feel that currently the information that there has to be a single landing module is implemented in two distant places in the code - my changes condense it into one if. It also moves usage of extract() return close to the call.

> You are restoring the logic that was proposed and rejected during review when this change was introduced.

Hmm... That would mean that I agree with whoever proposed this :) But if that really have been discussed than I am ok to revert it but I have to admit that I am not super convinced - just bad smell for me for future refactor.

pawelru:
> While we are here, the description of header and is borderline confusing:

Yes - fixed and simplified this. There was an idea of combining this one one but there are reports that it does not work nicely when inheriting - https://stackoverflow.com/questions/39762443/inherit-roxygen2-documentation-for-multiple-arguments-in-r-package. Decided to keep it separate.

> Lastly, I would also consider changing the order of the arguments by putting filter before title but that may be too serious a change.

I agree. This should be a separate issue discussed separately. Didn't want to propose not backwards compatible changes here.

pawelru:
> While we are here, the description of header and is borderline confusing:

Yes - fixed and simplified this. There was an idea of combining this one one but there are reports that it does not work nicely when inheriting - https://stackoverflow.com/questions/39762443/inherit-roxygen2-documentation-for-multiple-arguments-in-r-package. Decided to keep it separate.

> Lastly, I would also consider changing the order of the arguments by putting filter before title but that may be too serious a change.

I agree. This should be a separate issue discussed separately. Didn't want to propose not backwards compatible changes here.

R/init.R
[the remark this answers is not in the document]
chlebowa: @pawelru 
```
 if (inherits(modules, "list")) {
    modules <- do.call(teal::modules, modules)
  }
```
All right, a **list** of `FilteredData`s.

@averissimo 
> If we can protect against all allowable types, then we can "safely" convert those. 

Exactly. :+1: 

> Although I see a point in making sure that in the end, we have what is expected.

Our conversion routines/procedures depend _only_ on the input, so as long their logic is sound, checking results seems redundant to me. Of course, we are open to the problem of shifting definitions, e.g. we use a `foo` to convert `bar` to `pub` but then then `bar` stops being a list and becomes an S4 object and `foo` fails. But we have unit tests for such cases, do we not?

R/init.R
```diff
@@ -122,52 +119,57 @@ init <- function(data,
...
     )
   }
-
+  if (is.list(data) && !inherits(data, "teal_data_module")) {
+    checkmate::assert_list(data, names = "named")
+    data <- do.call(teal.data::teal_data, data)
+  }
   checkmate::assert_multi_class(data, c("teal_data", "teal_data_module"))
-  checkmate::assert_multi_class(modules, c("teal_module", "list", "teal_modules"))
+
```
gogonzo, line 127:
```suggestion
  checkmate::assert(
    checkmate::assert_multi_class(modules, c("teal_module", "teal_modules")),
    checkmate::assert_list(modules, c("teal_module", "teal_modules"))
  )
```

Same to @chlebowa and @averissimo comments below I prefer assert and then convert. I Also prefer assert error to have informative message. I admit that previously `assert_multi_class` wasn't precise enough.

R/init.R
```diff
@@ -122,52 +119,57 @@ init <- function(data,
...
-  } else if (inherits(data, "teal_data_module")) {
-    body(data$server)
+  landing_module <- NULL
+  if (length(landing) > 1L) {
+    stop("Only one `landing_popup_module` can be used.")
+  } else if (length(landing) == 1) {
+    landing_module <- landing[[1L]]
   }
+  modules <- drop_module(modules, "teal_module_landing")
+  rm(landing)
```
gogonzo, line 163: Why to remove `landing` object?  
```suggestion
  landing_module <- NULL
  if (length(landing) == 1) {
    landing_module <- landing[[1L]]
    modules <- drop_module(modules, "teal_module_landing")
  } else if (length(landing) > 1L) {
    stop("Only one `landing_popup_module` can be used.")
  }

```

R/module_teal_with_splash.R
gogonzo: I like better corrected suggestion but I liked current message more. No strong opinion though

chlebowa: How about something very specific?
```
stop("The `teal_data_module` passed to `data` must return a reactive expression.", call. = FALSE)
```

pawelru:
Thank you all for expressing your thoughts. I have applied this as you suggested, i.e. assert - convert. 

When doing this I noticed two possible disadvantages that have not been raised yet.

This is possibly not super DRY as we might repeat similar condition three times. Let me give you an example:

```
foo <- function(x) {
  assert(
    check_additional_class(x),  # <- here
    check_desired_class(x)
  )
  if (test_additional_class(x)) x <- additional_to_desired(x)  # <- here
  ...
}
additional_to_desired <- function(x) {
  assert_additional_class(x)  # <- here
  ...
}
```

as opposed to two times only (which is also not great tbh)
```
foo <- function(x) {
  if (test_additional_class(x)) x <- additional_to_desired(x)  # <- here
  assert(
    check_desired_class(x)
  )
  ...
}
additional_to_desired <- function(x) {
  assert_additional_class(x) # <- here
  ...
}
```

Now, let's imagine that this condition is not wrapped into a nice function but there are multiple combined conditions. Let's go further and imagine that these conditions requires internals from other packages or in general: logic that suppose to be encapsulated elsewhere. Such situation is observable in case of `data` argument which is validated inside `teal.data`. I tried to find a some reasonable middle ground in this case.

---

There is also one more thing. Usually we have one and only one _desired_ class and there are few _additional_ ones that are allowed and then convertible into the _desired_. If we put _additional_ classes into the assertions upfront then those will be listed in the error message. In case there are multiple _additional_ ones, users might interpret that they need to put some effort into converting an _additional_ one into the other _additional_ one whereas we actually want them to convert into the _desired_ one.

Let's take an example (dummy code):
```
r$> data <- list(iris, mtcars)
r$> checkmate::assert(
        .var.name = "data",
        checkmate::check_list(data, names = "named"),
        checkmate::check_multi_class(data, c("teal_data", "teal_data_module"))
      )
Error: Assertion on 'data' failed: One of the following must apply:
 * checkmate::check_list(data): Must have names
 * checkmate::check_multi_class(data): Must inherit from class 'teal_data'/'teal_data_module', but has class 'list'.
```

A valid interpretation of the above could be "I need to put names into the list" whereas I feel we should say "please convert to teal_data" instead.

R/init.R
pawelru: applied your suggestions

R/init.R
pawelru: Applied assert and convert. Please see the new message in the main PR theread. Also converted few assert_multi_class into a assert. If you spot anything more - please kindly put a new comment

R/init.R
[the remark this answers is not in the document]
pawelru: Hey guys. Thanks for all of this. It seems that there is an agreement among developers so I implemented this in a suggested way. There are a few additional small caveats that I have discovered down the way but we can also live with that I guess.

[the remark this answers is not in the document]
pawelru: (resolving this thread)

gogonzo:
> This is possibly not super DRY as we might repeat similar condition three times. Let me give you an example:

You right, probably we will always repeat some code. When the same argument is passed from function to function we should add assert to each function. I like your idea of having asserts grouped and arranged by the argument but In the same time I'm against it. Asserts are filling gap of the week typing in R. I prefer to have all asserts on the top of the function to communicate to developer what is the desired type (@param arg types in the documentation is a derivate of the assert not the opposite).


```
# c++
int foo(int a) {
  ...
}

# R
foo <- function(a) {
  checkmate::assert_int(a)
  ...
}
```


> In case there are multiple additional ones, users might interpret that they need to put some effort into converting

This is interesting point of view. I'm not sure, I think when the wrong type is provided the answer should be `wrong type "z"  - "a", "b" "c" are accepted`. `checkmate` doesn't provide ideal message, but I hope it will become a standard so the app developers can understand output of the `checkmate::assert` - if not we can always use `if () stop()` and write precise message and instructions (similar to what we have in the body `module()`)

R/module_teal_with_splash.R
pawelru: It's definitely more correct than the original one but I feel we shouldn't expect app developers to know what `teal_data_module` is. I would prefer to use more general "module" instead (as proposed).

Taking the dummy example from the above:  `"x argument of <abc> class needs to be of length 1"` is probably too verbose to `"x argument needs to be of length 1"`. Do you agree?

As I mentioned - it's better that the original one so I am also happy to continue with this as a compromise.

chlebowa: It is impossible to get that error message without using a `teal_data_module`. Conversely, `module` - what module?

pawelru: changed as per suggestion

gogonzo (approved):
👍 I don't have any objections anymore

R/init.R
```diff
@@ -107,11 +105,10 @@ init <- function(data,
                  header = tags$p(),
                  footer = tags$p(),
                  id = character(0)) {
-  logger::log_trace("init initializing teal app with: data ({ class(data)[1] }).")
-  if (is.list(data) && !inherits(data, "teal_data_module")) {
-    checkmate::assert_list(data, names = "named")
-    data <- do.call(teal.data::teal_data, data)
-  }
+  logger::log_trace("init initializing teal app with: data ({ toString(sQuote(class(data))) }).")
```
chlebowa, line 108:
```suggestion
  logger::log_trace("init initializing teal app with { 'class(data)' }.")
```
Class of `data` can now only be `teal_data`, `teal_data_module`, and `list` (#1031), so no need to concatenate complicated classes. Unless we want to allow extensions of `teal_data` but that is a different discussion.

I would also move this until after `data` has passed argument checks. After all, app initiation does not begin until well after that.

R/init.R
```diff
@@ -122,52 +119,74 @@ init <- function(data,
...
-  if (is.character(title)) {
-    title <- build_app_title(title)
-  } else {
-    validate_app_title_tag(title)
+  if (is.list(data) && !inherits(data, "teal_data_module")) {
+    data <- do.call(teal.data::teal_data, data)
   }
 
+  ## `modules`
+  checkmate::assert(
```
chlebowa, line 132: No `.var.name = "modules",` here?

R/init.R
```diff
@@ -122,52 +119,74 @@ init <- function(data,
...
-
-  # Calculate app id that will be used to stamp filter state snapshots.
-  # App id is a hash of the app's data and modules.
-  # See "transferring snapshots" section in ?snapshot.
-  hashables <- mget(c("data", "modules"))
-  hashables$data <- if (inherits(hashables$data, "teal_data")) {
-    as.list(hashables$data@env)
-  } else if (inherits(data, "teal_data_module")) {
-    body(data$server)
+  landing_module <- NULL
```
chlebowa, line 174: I still don't like this. You are creating another variable just because you want to test for `NULL` and not `length==0`. 
You are also uncoupling executing the landing module from validating it (i.e. validating that there is only one). This may be down to preference because you _are_ coupling it to the extraction. 
Dropping modules under a condition - I don't know. If we ever add a module class that requires special handling, it will taken right back out.

I am not going to press it further but let it be noted.

R/module_teal_with_splash.R
```diff
@@ -120,8 +140,10 @@ srv_teal_with_splash <- function(id, data, modules, filter = teal_slices()) {
         need(
           inherits(data, "teal_data"),
           paste(
-            "Error: `teal_data_module` did not return `teal_data` object",
-            "\n Check your inputs or contact app developer if error persists"
+            "Error: `teal_data_module` passed to `data` did not return `teal_data` object but",
```
chlebowa, line 143: minor suggestion
```suggestion
            "Error: `teal_data_module` passed to `data` failed to return `teal_data` object, returned",
```

chlebowa (changes requested):

R/init.R
pawelru: moved quotes and added brackets

```suggestion
  logger::log_trace("init initializing teal app with: data ('{ class(data) }').")
```

R/init.R
pawelru:
```suggestion
  checkmate::assert(
    .var.name = "modules",
```

chlebowa (approved):

R/init.R
pawelru: This is not only for replacing length check with NULL. This is rather an indirect consequence. My point was to keep the logic in one place close together. Namely: (i) landing module existence condition as well as (ii) extraction of the landing module used from all the landing modules we identified much closer to the landing modules extraction call. Currently (before this PR) we have both of these down below in `srv_teal_with_splash()` function call with a lot of other unrelated logic in between.
I really believe that this encapsulates the logic much better. This will make reading the code as well as potential future refactor much easier.

pawelru:
Thanks all for the valuable feedback 👍  Much appreciated. I think we ended up with a good conventions to carry forward.

PR #1015, opened 2023-12-19
Removes `library` calls from `get_rcode_libraries` in favor of manual specification

### LINKED ISSUE #950: Remove `library` calls from `get_rcode_header` in favour of manual specification
In the preprocessing app developer can add `library(pkg)` calls in the `eval_code/within` call. Thanks to this code in `qenv/teal_data` will contain `library` calls, so will be more reproducible.
```
data <- teal_data() |> within({
  library(ggplot2)
  ggplot(iris, aes(x = Sepal.Length, y = Petal.Length)) + geom_point()
})
```

However, when sending data to the module another [list of libraries is appended](https://github.com/insightsengineering/teal/blob/60755551478f0f66642399a7b5e78db96dd86471/R/module_nested_tabs.R#L320). We need to decide about who should be responsible for listing `libraries` used in a code (app developer) or (teal).
- app developer doesn't know what libraries are used within the modules, so the modules should rather use `::`
- `teal` can add all the libraries [according to what is loaded/attached in a session](https://github.com/insightsengineering/teal/blob/60755551478f0f66642399a7b5e78db96dd86471/R/get_rcode_utils.R#L7)

We need to find robust way to do manage this.

averissimo:
# Pull Request

Fixes #950

#### Changes description

- [x] Removes `library` calls if they are already manually specified by the app developer
  - [x] Ignore cases when `library(...)` is after a comment character (`#`)
  - [ ] Ignore cases when `library(...)` is written inside a string `var1 <- "a weird string with library(ggplot2)"`
- [ ] Show error when `loadedNamespaces` package is not installed in the user
  - Being discussed in issue

### DIFF
--- a/R/get_rcode_utils.R
+++ b/R/get_rcode_utils.R
@@ -5,17 +5,13 @@
 #' @return Character object contain code
 #' @keywords internal
 get_rcode_libraries <- function() {
-  vapply(
+  packages <- vapply(
     utils::sessionInfo()$otherPkgs,
-    function(x) {
-      paste0("library(", x$Package, ")")
-    },
+    function(x) paste0("library(", x$Package, ")"),
     character(1)
-  ) %>%
-    # put it into reverse order to correctly simulate executed code
-    rev() %>%
-    paste0(sep = "\n") %>%
-    paste0(collapse = "")
+  )
+  # put it into reverse order to correctly simulate executed code
+  paste(rev(packages), collapse = "\n")
 }
 
 
--- a/R/get_rcode_utils.R
+++ b/R/get_rcode_utils.R
@@ -4,17 +4,33 @@
 #'
 #' @return Character object contain code
 #' @keywords internal
-get_rcode_libraries <- function() {
-  packages <- vapply(
-    utils::sessionInfo()$otherPkgs,
-    function(x) paste0("library(", x$Package, ")"),
-    character(1)
-  )
+get_rcode_libraries <- function(dataset_rcode) {
+  packages <- rev(vapply(utils::sessionInfo()$otherPkgs, base::`[[`, character(1), "Package"))
+
+  parsed_libraries <- c()
+  if (!missing(dataset_rcode)) {
+    # Extract all lines with library()
+    #  TODO: remove strings first as this will pass "this is a string with library(something) in it"
+    user_libraries <- Filter(
+      function(.x) grepl("library\\(.*\\)$", .x),
+      vapply(strsplit(dataset_rcode, "\n")[[1]], trimws, character(1))
+    )
+
+    # Keep only library name
+    parsed_libraries <- gsub(
+      # library(...) must be located at beginning of line, or have a valid character before
+      "(^l|.*<-|.*[ ;=\\({]l)ibrary\\(([a-z][a-zA-Z0-9.]*)\\)$", "\\2",
+      # Strip out comments
+      vapply(user_libraries, function(.x) as.character(str2expression(.x)), character(1L))
+    )
+  }
+
   # put it into reverse order to correctly simulate executed code
-  paste(rev(packages), collapse = "\n")
+  paste(
+    "library(", Filter(Negate(function(.x) .x %in% parsed_libraries), packages), ")",
+    collapse = "\n", sep = ""
+  )
 }
-
-
 
 get_rcode_str_install <- function() {
   code_string <- getOption("teal.load_nest_code")
--- a/R/get_rcode_utils.R
+++ b/R/get_rcode_utils.R
@@ -7,17 +7,17 @@
 get_rcode_libraries <- function(dataset_rcode) {
   packages <- rev(vapply(utils::sessionInfo()$otherPkgs, base::`[[`, character(1), "Package"))
 
-  parsed_libraries <- c()
-  if (!missing(dataset_rcode)) {
+  parsed_libraries <- if (!missing(dataset_rcode)) {
     # Extract all lines with library()
     #  TODO: remove strings first as this will pass "this is a string with library(something) in it"
     user_libraries <- Filter(
-      function(.x) grepl("library\\(.*\\)$", .x),
+      function(.x) grepl("(^l|.*<-|.*[ ;=\\({]l)ibrary\\(([a-z][[:alnum:].]*)\\)$", .x),
+      # function(.x) grepl("library\\(.*\\)$", .x),
       vapply(strsplit(dataset_rcode, "\n")[[1]], trimws, character(1))
     )
 
     # Keep only library name
-    parsed_libraries <- gsub(
+    gsub(
       # library(...) must be located at beginning of line, or have a valid character before
       "(^l|.*<-|.*[ ;=\\({]l)ibrary\\(([a-z][a-zA-Z0-9.]*)\\)$", "\\2",
       # Strip out comments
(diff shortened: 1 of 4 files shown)


chlebowa (changes requested):
So with this PR we use `get_rcode_libraries` to list all loaded namespaces and put them in `library` calls but then we just remove the ones that were attached previously by calling `library` within `teal_data`. Am I correct? The following assumes that I am.

I don't think this is the right way to go. `get_rcode_libraries` was never a good solution to begin with. What is attached by the app dev* should be attached, and what is only loaded should be only loaded. I believe we should _not_ add `library` calls to the code.

(*) Attaching packages in modules is poor practice and should be discouraged but there is no way to prevent it so this extends to the module dev. Problems may ensue if a module attaches a package that masks a function used in prepro.

R/module_nested_tabs.R
```diff
@@ -285,10 +285,15 @@ srv_nested_tabs.teal_module <- function(id, datasets, modules, is_module_specifi
...
+  loaded_libs <- get_rcode_libraries(dataset_rcode)
+
   code <- c(
     get_rcode_str_install(),
-    get_rcode_libraries(),
-    get_datasets_code(datanames, datasets, hashes)
+    loaded_libs,
+    "",
+    dataset_rcode
   )
```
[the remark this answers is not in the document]
averissimo: It's a small UI tweak to create space between automated `library`s and `teal_data` code

I can revert or find a less-tweaky way of achieving this

image

averissimo:
> So with this PR we use `get_rcode_libraries` to list all loaded namespaces and put them in `library` calls, but then we just remove the ones that were attached previously by calling `library` within `teal_data`. Am I correct?

This PR still keeps the previous behavior, it just filters out the library calls. Using `namespaces` will produce too many library calls, which I don't think we want ("Too many" _(easily)_ means more than 50)

> I don't think this is the right way to go. `get_rcode_libraries` was never a good solution to begin with. What is attached by the app dev* should be attached, and what is only loaded should be only loaded. I believe we should _not_ add `library` calls to the code.
>
> (*) Attaching packages in modules is poor practice and should be discouraged but there is no way to prevent it so this extends to the module dev. Problems may ensue if a module attaches a package that masks a function used in prepro.

I agree that it's poor practice. A good solution seems unnatainable.

From my POV, either:
1. We add library calls with attached (as it is currently) or `loadedNamespaces` packages (which will include a lot of unnecessary packages for the code to be reproducible)
2. We relax our standards for the reproducibility code generation and don't include any libraries, leaving to:
    - App developer to add those explicit calls
    - User to install the required packages as the code fails

From my research so far we cannot isolate the `namespaces` in any way. (i.e. prevent the `qenv` environment from using packages attached elsewhere)

R/module_nested_tabs.R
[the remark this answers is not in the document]
chlebowa: The `teal_data` constructor does `code <- paste(code, collapse = "\n")`. Ensuring that the `library` calls string ends with a line break should be enough.

R/get_rcode_utils.R
```diff
@@ -4,21 +4,33 @@
...
+    #  TODO: remove strings first as this will pass "this is a string with library(something) in it"
+    user_libraries <- Filter(
+      function(.x) grepl("library\\(.*\\)$", .x),
+      vapply(strsplit(dataset_rcode, "\n")[[1]], trimws, character(1))
+    )
 
+    # Keep only library name
+    parsed_libraries <- gsub(
+      # library(...) must be located at beginning of line, or have a valid character before
+      "(^l|.*<-|.*[ ;=\\({]l)ibrary\\(([a-z][a-zA-Z0-9.]*)\\)$", "\\2",
```
chlebowa, line 22: Can `[a-z][a-zA-Z0-9.]` be replaced with `[:alnum:]`?

chlebowa (changes requested):
I'm afraid `library("<package>")` is just as valid as `library(<package>)` and parsing fails with the quotemarks. Maybe adding `gsub("\"", "", user_libraries)` will be enough to fix it.

R/module_nested_tabs.R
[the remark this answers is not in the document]
averissimo: reverted in 8b6be4510

R/get_rcode_utils.R
```diff
@@ -4,21 +4,33 @@
...
+      # Strip out comments
+      vapply(user_libraries, function(.x) as.character(str2expression(.x)), character(1L))
+    )
+  }
 
+  # put it into reverse order to correctly simulate executed code
+  paste(
+    "library(", Filter(Negate(function(.x) .x %in% parsed_libraries), packages), ")",
+    collapse = "\n", sep = ""
+  )
```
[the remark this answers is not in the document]
averissimo: Done in 8b6be4510

R/get_rcode_utils.R
```diff
@@ -4,21 +4,33 @@
...
+    )
 
+    # Keep only library name
+    parsed_libraries <- gsub(
+      # library(...) must be located at beginning of line, or have a valid character before
+      "(^l|.*<-|.*[ ;=\\({]l)ibrary\\(([a-z][a-zA-Z0-9.]*)\\)$", "\\2",
+      # Strip out comments
+      vapply(user_libraries, function(.x) as.character(str2expression(.x)), character(1L))
+    )
+  }
```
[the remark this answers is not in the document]
averissimo: Nice: 8b6be4510

R/get_rcode_utils.R
averissimo: Yes, `[[:alnum:].]` due to dot

chlebowa: Right, I missed the full stop when retyping :+1:

chlebowa: Wait, I think I misunderstood your reply to my comment. What I meant to ask is whether you can replace 
```
"(^l|.*<-|.*[ ;=\\({]l)ibrary\\(([a-z][a-zA-Z0-9.]*)\\)$"
```
with 
```
"(^l|.*<-|.*[ ;=\\({]l)ibrary\\(([:alnum:]*)\\)$"
```

I don't quite get why you use `<any lowercase letter><any number of letters/digits/anychar>`.

chlebowa: But I see now that things like `[:alnum:]` are not entirely reliable.

pawelru:
I very much dislike operating on the code as a text using regular expressions and so on. It would be relatively easy to obey the patterns / the amount of work to make it robust is huge due to number of possible options.

Few examples
```
library("foo") # I believe this one is covered
library("foo", character.only = T)
library(foo)
library      (foo)
require(foo)
do.call(library, list("foo"))
pkgs <- c("foo"); for(i in pkgs) library(i, character.only = T)
pkgs <- c("foo"); lapply(pkgs, library, character.only = T)
get("library")(foo)
eval(parse(text = paste0("lib", "rary", "(", "foo", ")")))
foo <- library; foo(bar)
...
```

I haven't gone deep enough to the problem to suggest an alternative - it's just a warning for you that this approach is a rabbit hole.

chlebowa:
The three functions used here _and_ `teal.slice::get_filter_expr` called by `get_datasets_code` (code from `main`) return character.
Would you have them all modified to work on language objects?
```
  code <- c(
    get_rcode_str_install(),
    get_rcode_libraries(),
    get_datasets_code(datanames, datasets, hashes)
  )
```

chlebowa:
> Eventually yes.

Will this do? https://github.com/insightsengineering/teal/issues/1027

pawelru:
> > Eventually yes.
> 
> Will this do? #1027

Yes. Much appreciated!

R/get_rcode_utils.R
averissimo: I misunderstood the documentation about the first letter.

A valid package name needs to be `<letter><any number, letter and dot>`

averissimo:
We can restrict the rabbit hole and only comply with common patterns. I see this feature as a "nice to have" to avoid duplicated code and not as a complete solution.

Otherwise, it might risk becoming a new code parser if we enter the rabbit hole :sweat_smile: (even when dealing with call/expressions)

We might have an alternative where we detect changes in the session during `eval_code` / `within` and store that as metadata along the way. _(a `setdiff` would then be enough to remove duplicates)_

PR #1018, opened 2023-12-22
Add a helper function to add app title with a favicon

### LINKED ISSUE #1017: [Feature Request]: Simpler way to add an app title with a favicon
### Feature description

Currently, the process of adding an app title along with a favicon is hard and we have two problems here:
1. We have a [workaround in teal.gallery](https://github.com/insightsengineering/teal.gallery/blob/9f44972268348b25709e75f6244f9433a1fab899/basic-teal/app.R#L13) by adding the shortcut link and title as raw tags which should not be done by app developers who do not know about `shiny` or `HTML`.
2. We do that in the header argument which is again confusing because we are passing the title over there.

Screenshot 2023-12-22 at 8 33 11 AM

This problem arises because in the [main UI created for teal](https://github.com/insightsengineering/teal/blob/0204dd156d56b901bfa1756d4fdb0d327fbc9dfe/R/module_teal.R#L111-L129) we are passing the title into `fluidPage` which makes use of the [titlePanel](https://shiny.posit.co/r/reference/shiny/1.7.0/titlepanel) which does work well with favicons, They don't appear while interactively running shiny apps and on shinyapps.io deployments.

I propose a fix which is basically the hack solution we have implemented in the `teal.gallery` but implemented with a helper function so it is easier for an app developer to create this.

### Code of Conduct

### Contribution Guidelines

### Security Policy

vedhav:
Closes #1017

Adds a helper function for easy title in teal.

### What titles are created in these scenarios now?

Title - "teal title"
Image - from the remote link
```r
nest_logo <- "https://raw.githubusercontent.com/insightsengineering/hex-stickers/main/PNG/nest.png"
teal::init(
    title = build_app_title(title = "teal title", nest_logo)
)
```
---
Title - "teal title"
Image - from the local path, provided the path is hosted in the static paths of shiny, by default it is `www/`
```r
nest_logo <- "nest.png"
teal::init(
    title = build_app_title(title = "teal title", nest_logo)
)
```
---
Title - "teal title"
Image - NEST logo from GitHub
```r
teal::init(
    title = "teal title"
)
```
---
Title - "Teal app"
Image - NEST logo from GitHub
```r
teal::init()
```
---

After this is merged, we can update the examples in `teal.gallery` with the title in the dev branch.

What do you think about the defaults @lcd2yyz and @donyunardi


_pkgdown.yml
```diff
@@ -97,6 +97,10 @@ reference:
       - srv_teal_with_splash
       - ui_teal_with_splash
       - teal_slices
+  - title: Helper functions
```
chlebowa, line 100: Match case with other section titles.

R/utils.R
```diff
@@ -230,3 +230,25 @@ teal_data_datanames <- function(data) {
     ls(teal.code::get_env(data), all.names = TRUE)
   }
 }
+
+#' Build an app title with a favicon
```
chlebowa, line 234:
```suggestion
#' Build app title with favicon
```

R/utils.R
```diff
@@ -230,3 +230,25 @@ teal_data_datanames <- function(data) {
...
+#' A helper function to create the browser title along with a logo.
+#'
+#' @param title (`character`) The browser title for the teal app
+#' @param favicon (`character`) The path for the icon for the title.
+#' The image/icon path can be remote or the static path accessible by shiny, like the `www/`
+#'
+#' @return A `shiny.tag` containing the element that adds the title and logo to the shiny app
+build_app_title <- function(title = "Teal app", favicon = "https://raw.githubusercontent.com/insightsengineering/hex-stickers/main/PNG/nest.png") { # nolint
+  checkmate::assert_string(title, null.ok = TRUE)
+  checkmate::assert_string(favicon, null.ok = TRUE)
```
chlebowa, line 245: Use `checkmate::assert_file_exists`.

R/utils.R
```diff
@@ -230,3 +230,25 @@ teal_data_datanames <- function(data) {
...
+  tags$head(
+    tags$title(title),
+    tags$link(
+      rel = "icon",
+      href = favicon,
+      sizes = "any"
+    )
```
chlebowa, line 252: This doesn't have to be conditional on `is.null(favicon)`, does it?

R/utils.R
chlebowa, line 245: It would be nice to check if the file is actually there but since this allows a remote resource, `assert_file_exists` will not do the job.

chlebowa (commented):
Can you please fix this while you're here?

`teal_data_modules` or simply a list of a named list of objects

should read

`teal_data_module` or simply a list of a named list of objects

_pkgdown.yml
vedhav:
```suggestion
  - title: Helper Functions
```

R/utils.R
vedhav: It will work fine for a `NULL` value of favicon.

R/init.R
```diff
@@ -16,15 +16,15 @@
...
+#' `teal_data_module` or simply a list of a named list of objects
 #' (`data.frame` or `MultiAssayExperiment`).
 #' @param modules (`list`, `teal_modules` or `teal_module`)\cr
 #'   nested list of `teal_modules` or `teal_module` objects or a single
 #'   `teal_modules` or `teal_module` object. These are the specific output modules which
 #'   will be displayed in the teal application. See [modules()] and [module()] for
 #'   more details.
-#' @param title (`NULL` or `character`)\cr
-#'   The browser window title (defaults to the host URL of the page).
+#' @param title (`shiny.tag` or `character`)\cr
```
chlebowa, line 26: I'm not sure allowing _any_ `shiny.tag` is a good idea. Consider having `build_app_title` add a special class to check. (Emphasis on consider.)

chlebowa (changes requested):
Please set default values of `build_app_title` to `NULL`. We want to allow the user to easily add a title and an icon, not force them to remove our logo every time they build an app.

Maybe add another argument like `use_teal = TRUE` that will fill unset arguments with `teal` defaults? Just a thought.

vedhav:
We could go either way when it comes to the defaults of the `build_app_title`. I would argue that having to see `Teal app` is much better than having to see `127.0.0.1:port`.
Additionally, from my perspective it is still better to add a NEST/teal favicon instead of none at all, This is what is done it both `rhino` and `golem`. It is created with `rhino` or `golem` logo as default. Hence relying on feedback from @lcd2yyz or @donyunardi on the defaults.

R/init.R
vedhav: That's a good catch. Have to restrict to a sensible tag and document it as well. I'll add a change for this.

vedhav: Added a validation when `shiny.tag` is the input and improved the docs. PTAL now.

R/utils.R
```diff
@@ -230,3 +230,43 @@ teal_data_datanames <- function(data) {
...
+#' @keywords internal
+validate_app_title_tag <- function(shiny_tag) {
+  checkmate::assert_class(shiny_tag, "shiny.tag")
+  checkmate::assert_true(shiny_tag$name == "head")
+  child_tags <- sapply(shiny_tag$children, function(child) child$name)
+  checkmate::assert_true(
+    all(
+      c("title", "link") %in% child_tags
+    )
+  )
```
chlebowa, line 247:
```suggestion
child_names <- vapply(shiny_tag$children, `[[`, character(1L), "name")
checkmate::assert_set_equal(child_names, c("title", "link"), .var.name = "child tags")
```

R/utils.R
```diff
@@ -230,3 +230,43 @@ teal_data_datanames <- function(data) {
...
+  link_tag <- shiny_tag$children[which(child_tags == "link")][[1]]
+  checkmate::assert_true(link_tag$attribs$rel %in% c("icon", "shortcut icon"))
```
chlebowa, line 249:
```suggestion
rel_attr <- shiny_tag$children[[which(child_names == "link")]]$attribs$rel
checkmate::assert_subset(rel_attr, c("icon", "shortcut_icon"), .var.name = "Link tag's rel attribute")
```

R/utils.R
chlebowa, line 1: I was thinking of something simpler like adding `class(ans) <- c("title.tag", class(ans))` to `build_app_title`. If you think this is too lax, please consider my suggestions for the checks.

R/utils.R
vedhav: It might be a little restrictive. Technically, only the presence of these two is needed. A user can pass additional elements to this head tag. I can't think of any use case for this. But, I don't see any problem with keeping this open. So I would prefer to allow the users to pass additional tags along with `title` and `link`. We just need to check for the presence of these two. I agree with the checkmate simplification.

chlebowa: I made it restrictive so that _only_ a tag generated by `build_app_title` can get through. Was that not the purpose of the validation function? Again, if you want looser criteria, why not go with just adding a class?

vedhav: Right now, there is flexibility to provide just an HTML tag as input without the need to use the helper function, which is more flexible than creating yet another custom class. We already have a lot of those and I don't feel that we benefit from this.
We indeed want the tags generated from the `build_app_title` to work. But, essentially it has to be an optional helper function. Hence, we should be supporting all the valid title tags generated by the user.

chlebowa: Ok, so if you want `build_app_title` to be optional, you actually open the `title` argument significantly. Previously it accepted only character strings, now it accepts `shiny.tag`s. My original question is: are you sure you want to allow the user to pass _any_ `shiny.tag`, to which you answer seems to yes, as long as it can stick a favicon in the title. Are you sure you want to give the user this much wiggle room? _Any_ tag?

Conversely, if you add a special class and restrict `title` to that class _or character_, you give the user one more, very specific option. And they can still pass a bare string and you function will intercept it.

vedhav: From an app developer's point of view, they are free to pass any shiny tag in the header and footer. A valid shiny tag also includes a character.
So, I thought it's best to follow the same input types for the title too. Think about it like we have a validation of tag in place and a helper function to help create this tag.
Screenshot 2023-12-24 at 10 54 50 AM

lcd2yyz (approved):
Defaults looks great to me, thanks!

R/module_teal.R
```diff
@@ -112,6 +112,7 @@ ui_teal <- function(id,
     title = title,
     theme = get_teal_bs_theme(),
     include_teal_css_js(),
+    title,
```
donyunardi, line 115: I don't understand why we need this line since we already passed the `title` object (result of `build_app_title`) in `title=` argument.

I removed this line locally and still get the same behavior.
Would you please clarify?

donyunardi (commented):
This is a great idea!
Just had a minor comment.

R/module_teal.R
vedhav: Yeah you're right. One of the `title` can be removed. Right now it adds two title tags in the head. I'll keep the title argument as it's more explicit.

PR #1011, opened 2023-12-15
898 save app state version 3

chlebowa:
WIP

PR #1048, opened 2024-01-12
898 save app state version 4

chlebowa:
Extension of #1011 

Five key objects are copied to the app session's `userData`: `slices_global`, `filtered_data_list`, `mapping_matrix`, `snapshot_history`, and `state_history`.
These objects can now be used and modified from any module.

There are multiple implementation within this one draft. Use the SHAs to checkout the one you want to see.

<details>
<summary> first implementation (f987c04d9d555db13cf43ef6730af8308d8cc121)</summary>

Introduced a function `set_intermodule_objects` that sets these objects in the scope of a module:
- checks if an object exists in session
- if YES, assigns a copy in module environment (by reference)
- if NO, creates the object
This function is dirty as all hell, it depends on state and has side effects. Possibly is buggy.

`set_intermodule_objects` is called in filter manager, snapshot manager, and state manager so that the key objects can be created and used. Passing them as arguments is no longer necessary.
Note this is discouraged by `shiny` but perhaps it will allow us to do some things otherwise impossible.

</details>

The problem with the first implementation is that `set_intermodule_objects` contains all routines for setting the key objects and thus the default value of the key objects is separated from the module that uses it. It would be better if they (the routine and the module) can be kept in one file. Therefore, a need arises for a way to send such routines to a place where they can be accessed by all modules.

<details>
<summary> second implementation (3f79bb201667b4f88f15f1faa88178e8ac1d373d)</summary>

Introduced a dedicated setter function for every key object (five in total). These setter functions work much like `set_intermodule_objects` in the first implementation. They are defined in the same file as the modules in which the key object is created, along with all necessary utility functions for better maintainability.

Introduced R6 class `GlobalSetter` as a hub for the setter functions.
The `GlobalSetter` class has two methods:
- `$add_setter`: register function that sets global object
- `$set_global`: call setter function to set global object
A singleton `GlobalSetter`, `GS` is created in the package namespace (assignment is made in `zzz.R`). From there it be accessed by all modules.
Every module can register a setter function for the key objects that it creates in `GS`. Every module can use `GS` to call a setter function.

</details>

#### ISSUES

- `update*Input` is still a problem.
- application opened from bookmark still lacks the most recent bookmark (itself)

### DIFF
--- a/R/module_filter_manager.R
+++ b/R/module_filter_manager.R
@@ -117,39 +117,13 @@ filter_manager_srv <- function(id, filtered_data_list, filter) {
     # Contains all available teal_slice objects available to all modules.
     # Passed whole to instances of FilteredData used for individual modules.
     # Down there a subset that pertains to the data sets used in that module is applied and displayed.
-    slices_global <- reactiveVal(filter)
-
-    filtered_data_list <-
-      if (!is_module_specific) {
-        # Retrieve the first FilteredData from potentially nested list.
-        # List of length one is named "global_filters" because that name is forbidden for a module label.
-        list(global_filters = unlist(filtered_data_list)[[1]])
-      } else {
-        # Flatten potentially nested list of FilteredData objects while maintaining useful names.
-        # Simply using `unlist` would result in concatenated names.
-        flatten_nested <- function(x, name = NULL) {
-          if (inherits(x, "FilteredData")) {
-            setNames(list(x), name)
-          } else {
-            unlist(lapply(names(x), function(name) flatten_nested(x[[name]], name)))
-          }
-        }
-        flatten_nested(filtered_data_list)
-      }
-
-    # Create mapping fo filters to modules in matrix form (presented as data.frame).
-    # Modules get NAs for filters that cannot be set for them.
-    mapping_matrix <- reactive({
-      state_ids_global <- vapply(slices_global(), `[[`, character(1L), "id")
-      mapping_smooth <- lapply(filtered_data_list, function(x) {
-        state_ids_local <- vapply(x$get_filter_state(), `[[`, character(1L), "id")
-        state_ids_allowed <- vapply(x$get_available_teal_slices()(), `[[`, character(1L), "id")
-        states_active <- state_ids_global %in% state_ids_local
-        ifelse(state_ids_global %in% state_ids_allowed, states_active, NA)
-      })
-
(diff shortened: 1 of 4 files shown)


R/module_filter_manager.R
```diff
@@ -252,3 +231,96 @@ filter_manager_module_srv <- function(id, module_fd, slices_global) {
...
+    sesh$userData$filtered_data_list
+  }
+}
+#' @keywords internal
+#' @noRd
+setter_mapping_matrix <- function() {
+  sesh <- getDefaultReactiveDomain()
+  if (is.null(sesh$userData$mapping_matrix)) {
+    filtered_data_list <- dynGet("filtered_data_list")
+    slices_global <- dynGet("slices_global")
```
gogonzo, line 321: Seemingly independent function requires always to be called in the environment where `filtered_data_list` and `slices_global` exists (or will exist in the moment of a evaluation). This is not just unsafe but also too complicated.

R/module_filter_manager.R
```diff
@@ -252,3 +231,96 @@ filter_manager_module_srv <- function(id, module_fd, slices_global) {
...
+    sesh$userData$mapping_matrix <- create_mapping_matrix(filtered_data_list, slices_global)
+  } else {
+    sesh$userData$mapping_matrix
+  }
```
gogonzo, line 325: Existence of `GlobalSetter` complicates this as well. Optimally function `setter_mapping_matrix` should be called once and it shouldn't check whether values are set or not.
```suggestion
```

gogonzo (changes requested):
This class unnecessary complicated the whole process of setting `userData$`. `userData` is a singleton itself (on the session level) and doesn't need extra class to handle the process. 
`GlobalSetter` doesn't make it more robust as `setter_` functions are vulnerable. See the comment in`setter_mapping_matrix`.

I suggest to remove `GlobalSetter` and fix `setter_xyz` to use formal arguments instead of `dynGet`. Apart from this, utilization of `userData` is reasonable. Alternatively I can imagine R6 `TealSingleton` class to control mapping/snapshot/states and whatever exist i a future. However, I would keep it simple now to at least have a POC of bookmarking.

R/module_filter_manager.R
chlebowa: > Seemingly independent function

Where did you get that idea? :laughing: All of this is impure as all hell but If we are to share objects between modules, there is no way to do it in a pure way.

PR #1037, opened 2024-01-10
492 user feedback on documentation

chlebowa:
Closes [this issue](https://github.com/insightsengineering/coredev-tasks/issues/492)

PR #1049, opened 2024-01-15
Remove some@pre release cleanup@main

gogonzo:
Removing section about "empty cycle". I don't agree with reasoning in this section. InsertUI doesn't cause this problem. Inputs specified in UI are initialized perfectly fine and there is no initial NULL value. Initial NULL value is a consequence of initializing empty selection and then updating on the server side. Problem refers only to the data_extract and teal has nothing to do with this. Problem is more generic and pretty normal for `shiny`.


chlebowa:
Would you mind basing this on https://github.com/insightsengineering/teal/pull/1037?

PR #1051, opened 2024-01-17
1031 drop list support

### LINKED ISSUE #1031: init - data arg - Are we planning to keep supporting a "list of objects"? Note that they don't have to be either `data.frame` or `MAE` any more.
Are we planning to keep supporting a "list of objects"? Note that they don't have to be either `data.frame` or `MAE` any more.

_Originally posted by @chlebowa in https://github.com/insightsengineering/teal/pull/1008#discussion_r1444811645_

chlebowa:
Fixes #1031

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,6 @@
 
 ### New features
 
-* `data` argument in `init` now accepts `teal_data` and `teal_data_module`.
 * Added `landing_popup_module` function which creates a module that will display a popup when the app starts. The popup will block access to the app until it is dismissed.
 * Filter state snapshots can now be uploaded from file. See `?snapshot`.
 * Added `as_tdata` function to facilitate migration of modules to the new `teal_data` class.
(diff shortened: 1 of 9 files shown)


pawelru:
Please add a news entry

chlebowa:
I was about to but then I noticed it is already included in the first line: 
```
# teal 0.14.0.9039

### New features

* `data` argument in `init` now accepts `teal_data` and `teal_data_module`.
* Added `landing_popup_module` function which creates a module that will display a popup when the app starts. The popup will block access to the app until it is dismissed.
* Filter state snapshots can now be uploaded from file. See `?snapshot`.
* Added `as_tdata` function to facilitate migration of modules to the new `teal_data` class.
* Added `build_app_title` function to facilitate adding favicons to app title. 
```

Would you like to make the wording stronger, e.g. "only `teal_data` and `teal_data_module`?

pawelru:
☝️ as well as tlg-catalog and teal.gallery

pawelru:
> Would you like to make the wording stronger, e.g. "only teal_data and teal_data_module"?

I think it would be better. Plus also move it to a breaking changes section.

chlebowa:
> ☝️ as well as tlg-catalog and teal.gallery

Those have been using `teal_data` exclusively since we merged the refactor.

m7pr (approved):
excellent simplification of way too many possibilities of a specification of the input data

PR #1053, opened 2024-01-18
1032 fix ellipsis documentation for `teal::modules()`

### LINKED ISSUE #1032: fix documentation for `modules`
Argument descriptions are wrong.

image

m7pr:
Close #1032 

Issue appeared because there is multiple function having the same parameter being documented in one documentation page. The last appearance of `@param` took precedence and overwritten any other text/

The idea is to have one entry for `@param ...` explaining usage for different functions. 

image

The drawback of the solution is that `@param ...` is not not documented near `print` nor `toString` functions. 
We could either 
- leave it as it is (with this PR fixing documentation)
- move all parameters to one place
- leave `@param ...` entries next to `print` and `toString` functions but without `@` tag, so they are treated as comments.


m7pr:
btw, is this a vicious circle

image

chlebowa:
Do you think you could do #942 as well? That way this page could be made much cleaner.

m7pr:
Yeah, made a PR https://github.com/insightsengineering/teal/pull/1054 to this branch for fix #942 
I didn't find it to be more sophisticated than replacing names from `toString` to `format` as `toString` was a custom function.

m7pr:
You think we are ready to close this one @chlebowa ?

chlebowa (changes requested):
Please break line after lifecycle badges.

I don't think `x` needs to be documented for methods of base generics.

`submodule` looks like an entity but it is not. Please rephrase, e.g. "each nested `teal_modules` or `teal_module` is indented..."

m7pr:
Hey @chlebowa when I remove documentation of `x` argument, we do get this WARNING during R CMD CHECK

```r
checking Rd \usage sections ... WARNING
  Undocumented arguments in documentation object 'modules'
    'x'
  
  Functions with \usage entries need to have the appropriate \alias
  entries, and all their arguments documented.
  The \usage entries must correspond to syntactically valid R code.
  See chapter 'Writing R documentation files' in the 'Writing R
  Extensions' manual.
```

So I guess we should keep `@param x` for `format.teal_module*` and `print.teal_module*`

m7pr:
> Please break line after lifecycle badges.

Good catch! Fixed

> I don't think x needs to be documented for methods of base generics.

Won't do. It throws WARNING when we build R CMD CHECK.

> submodule looks like an entity but it is not. Please rephrase, e.g. "each nested teal_modules or teal_module is indented..."

Valid point! Also changed as requested. Thanks!

chlebowa:
How about rebasing this PR to `pre-release_cleanup@main`?

R/modules.R
```diff
@@ -405,44 +408,40 @@ module_labels <- function(modules) {
...
-  cat(s)
-  return(invisible(s))
+  cat(format(x, ...))
+  invisible(x)
 }
 
 #' Prints `teal_module`
 #' @param x `teal_module`
-#' @param ... parameters passed to `toString`
+#' @param ... arguments passed to `format`
```
chlebowa, line 444: This leaves `...` in `format` undocumented.

chlebowa (commented):
Please add a line break after the badge in `module` as well.

R/modules.R
m7pr: aaah right. changed in here so they have generic message b01b57207c2b538e56ecbf8dfca2cac3d3e11fa6
`arguments passed to other methods.`

m7pr:
> Please add a line break after the badge in module as well.

8c0dcb97462bc3d364ea8c961183cde3a121288f

m7pr:
> How about rebasing this PR to pre-release_cleanup@main? 

Changed the base of this PR and will resolve conflicts in a second

chlebowa:
> an automated merge

What in blazes is that? :monocle_face:

m7pr:
auto-merge on an approval that change into an auto-merge without the approval since feature branches do not require reviews and approvals : P

PR #1054, opened 2024-01-18
942 rename toString to format for modules methods

### LINKED ISSUE #942: printing `teal_module(s)` objects

m7pr:
Close #942

Renamed `toString.modules` to `format.modules`


R/modules.R
```diff
@@ -436,14 +434,14 @@ toString.teal_module <- function(x, indent = 0, ...) { # nolint
 #' @export
 #' @rdname modules
 print.teal_modules <- function(x, ...) {
-  s <- toString(x, ...)
+  s <- format(x, ...)
   cat(s)
   return(invisible(s))
```
chlebowa, line 439: `print` should return `x`, not `format(x)`.
Are you defining a `print` method is necessary? Won't the default one do?

R/modules.R
```diff
@@ -436,14 +434,14 @@ toString.teal_module <- function(x, indent = 0, ...) { # nolint
...
 }
 
 #' Prints `teal_module`
 #' @param x `teal_module`
-#' @param ... parameters passed to `toString`
+#' @param ... parameters passed to `format`
```
chlebowa, line 444:
```suggestion
#' @param ... arguments passed to `format`
```

R/modules.R
```diff
@@ -9,9 +9,7 @@
 #'
 #' @param ...
 #' - For `modules()`: (`teal_module` or `teal_modules`) see [module()] and [modules()] for more details.
-#' - For `toString()`: (optional) additional parameters to pass to recursive calls of `toString`.
-#' - For `print()`: parameters passed to `toString`.
-#'
+#' - For `format()` and `print()`: (optional) additional parameters to pass to recursive calls of `format`.
```
chlebowa, line 12: A common phrase in R documentation is 
`... arguments passed to other methods`

chlebowa (changes requested):
Could you add some unit tests for `format` methods? Just a few will do.

R/modules.R
m7pr: I am maintaining consistency of what we had before this PR. So before this PR `print.teal_modules` `cat` and returned `toString(x)`. I do see `print` being weird in `cat` being called and then returned silently so I guess there is specific reason to do so. It does cat into the console and it returns an object in the same time so you can assign the result of the print while seeing the console output at the same time. Not sure why (and if) it's needed.

```r
> print(example_module()) -> x
+ example teal module
> x
[1] "+ example teal module"
```

I don't think default `print` is enough, since `teal_module` is a list (with an extra class `teal_module`). Probably the `teal_module` class is added to this list so that we can have assertions on this object in functions usage and we can have a personalized `print` method. In this case previous `toString` (currently `format`) went through the list recursively and appended indentation in the final output of the `print`.

So, customized `print` prints labels of modules in a nested setup

```r
> example_module()
+ example teal module
```

where default `print` just prints the list structure

```r
print(unclass(example_module()))
$label
[1] "example teal module"

$server
function(id, data) {
        ns <- session$ns
        updateSelectInput(session, "dataname", choices = isolate(teal.data::datanames(data())))
        output$text <- renderPrint({
          req(input$dataname)
          data()[[input$dataname]]
        })
          id = "rcode",
          verbatim_content = reactive(teal.code::get_code(data())),
          title = "Association Plot"
        )
      })
    }
<bytecode: 0x0000017c830e3130>
<environment: 0x0000017c82c20f18>

$ui
function(id) {
      ns <- NS(id)
        encoding = div(
        )
      )
    }
<bytecode: 0x0000017c83386a78>
<environment: 0x0000017c82c20f18>

$datanames
[1] "all"

$server_args
NULL

$ui_args
NULL

```

R/modules.R
m7pr: Yeah I know, but for `format` it actually passes an argument (recursively) to `format` and not anything else : ) For `print` it only passes `...` to `format`, So I think the phrasing is accurate.

R/modules.R
chlebowa: > What should be returned?

You're right, it does return the object. My mistake.

I do think, however that we can dispense with the assignment and just do 
```
cat(format(x, ...))
invisible(x)
```
wouldn't you agree?


> Is default print enought?

You're absolutely right, again. Sorry, I was distracted by `R6` :unamused:

R/modules.R
chlebowa: Doesn't `format.modules` also pass to `format.module`?

m7pr: changing to `... arguments passed to other methods
`

m7pr: bf7fd6b16a2fdb4d7f15084fe7017ebdf3ca148b

R/modules.R
m7pr: Changed to 

```r
cat(format(x, ...))
invisible(x)
```

53438604963421d01d6b9a1d9b6713a927085c7d

chlebowa (approved):
:+1:

PR #1067, opened 2024-01-24
Amend error message when `teal_data` is empty

### LINKED ISSUE #1042: bad unit tests
```
testthat::test_that("init throws when data has no datanames", {
  testthat::expect_error(
    init(data = teal_data(), modules = list(example_module())),
    "`data` object has no datanames and its environment is empty"
  )
})
```

This runs
```
td <- within(teal_data(), i <- iris)
init(data = td, modules = example_module()) %>% runApp
```

The test passes because there is no data.

Defaulting `datanames` to `teal_data@env` contents has made things somewhat confusing.

Amend error message and unit test.

vedhav:
Closes #1042

### DIFF
--- a/R/init.R
+++ b/R/init.R
@@ -217,7 +217,12 @@ init <- function(data,
   ## `data` - `modules`
   if (inherits(data, "teal_data")) {
     if (length(teal_data_datanames(data)) == 0) {
-      stop("`data` object has no datanames and its environment is empty. Specify `datanames(data)` and try again.")
+      stop(
+        paste0(
+          "`data` object has no datanames and its environment is empty.\n",
+          "Specify `datanames(data)` or make sure that the `data` object is not empty."
+        )
+      )
     }
     # in case of teal_data_module this check is postponed to the srv_teal_with_splash
     is_modules_ok <- check_modules_datanames(modules, teal_data_datanames(data))
--- a/tests/testthat/test-init.R
+++ b/tests/testthat/test-init.R
@@ -81,7 +81,7 @@ testthat::test_that("init filter accepts `teal_slices`", {
   )
 })
 
-testthat::test_that("init throws when data has no datanames", {
+testthat::test_that("init throws when an empty teal_data is used", {
   testthat::expect_error(
     init(data = teal_data(), modules = list(example_module())),
     "`data` object has no datanames and its environment is empty"


kartikeyakirar (approved):
LGTM

R/init.R
```diff
@@ -208,7 +208,12 @@ init <- function(data,
   ## `data` - `modules`
   if (inherits(data, "teal_data")) {
     if (length(teal_data_datanames(data)) == 0) {
-      stop("`data` object has no datanames and its environment is empty. Specify `datanames(data)` and try again.")
+      stop(
+        paste0(
+          "`data` object has no datanames and its environment is empty.\n",
+          "Specify `datanames(data)` or make sure that the `data` object is not empty."
+        )
+      )
```
chlebowa, line 216: This change changes nothing.
Also, don't use `paste` in condition messages unless you have to.

chlebowa (changes requested):

R/init.R
vedhav: I will remove the paste.
I thought the main issue was that we're computing the datanames so nudging the user when it's empty was the way to do it. What did you have in mind?

chlebowa: The line you appended repeats the previous one and adds nothing.

vedhav:
```suggestion
      stop(
          "`data` object has no datanames and its environment is empty.\n",
          "Specify `datanames(data)` or make sure that the `data` object is not empty."
      )
```

vedhav: Then I don't get the crux of the issue yet.
Was it not to let the user know that the `data` slot is empty which they should fix, which was missing earlier?

chlebowa:
I may have phrased #1042 insufficiently clearly.

Consider an app run with a `teal_data` that is populated by evaluating code and thus having no `datanames`.
```
> td <- teal_data() %>% within(i <- iris)
> init(data = td, modules = example_module()) %>% runApp
```

This app runs but raises a warning:
```
Listening on http://127.0.0.1:3633
Warning in `<reactive:teal_data_rv_validate>`(...) :
  `data` object has no datanames. Default datanames are set using `teal_data`'s environment.
```

The presence of `datanames` is now irrelevant for running the app. The error message may suggest otherwise and the test description states the exact opposite.

vedhav:
> The presence of datanames is now irrelevant for running the app. The error message may suggest otherwise and the test description states the exact opposite.

True, But we still would like to nudge the users to provide datanames themselves and not rely on automatic datanames assignment which literally takes all the names from the `@env`
The realistic cause of this error only arise when an empty `teal_data()` is passed which is more likely caused when the user forgets to assign the result of the `within()` evaluation. Don't you think it's best to remind and make sure the user knows to set the datanames?

chlebowa:
> True, But we still would like to nudge the users to provide datanames themselves and not rely on automatic datanames assignment which literally takes all the names from the `@env`

This is already done by the aforementioned warning message.

The error will now only arise when you pass an empty container. Setting `datanames` would do no good then, even if you could set them.
```
> td <- teal_data()
> datanames(td) <- "a"
Error in `datanames<-`(`*tmp*`, value = "a") : 
  Assertion on 'value' failed: Must be a subset of the empty set, i.e. also empty.
```

vedhav:
> This is already done by the aforementioned warning message.

I don't think so. This warning only arises when app is run. While the error arises during the creation of the app object. So, the error proceeds the warning.

chlebowa:
But the error says that the problem is a) no datanames and b) empty environment and only the latter is true because the former is irrelevant for raising the error.

vedhav:
Okay, technically yes this can only be caused by the later. And since we have a warning to show I can see why completely removing it here makes sense. I just did not want two steps for the user to figure this out. But, you're right. This is redundant.

chlebowa:
I agree one step would be better than one but we have decoupled the logic and messages must follow logic to the letter. Any deviation is the first step on the road to madness.

chlebowa (approved):
:+1:

Issue #1069, opened 2024-01-25
[Bug]: The verification status of the `teal_data` object is lost making all data to be unverified


vedhav:
### What happened?

As pointed out over here https://github.com/insightsengineering/teal.modules.clinical/issues/919#issuecomment-1909141154
even when a verified `teal_data` object is passed into the init it becomes unverified when it reaches the modules.

This happens because we always create a teal_data object without verifying in the [.datasets_to_data](https://github.com/insightsengineering/teal/blob/b1bbe12b2e6655db7daac1fc4fc9433a45ee33f5/R/module_nested_tabs.R#L294-L297) function.

Changes to be made:
1. Update the `teal.data::teal_data` so that it can handle `check = TRUE` case.
2. Update the `teal` module server functions to properly pass the check into this function.

Example apps for testing:
```r
app_without_warning <- function() {
  data <- within(teal_data(), {
    IRIS <- iris
  })
  datanames(data) <- "IRIS"
  app <- init(data = data, modules = example_module())
  shinyApp(app$ui, app$server)
}

app_with_warning <- function() {
  data <- teal_data(IRIS = iris)
  app <- init(data = data, modules = example_module())
  shinyApp(app$ui, app$server)
}

# Expectation: This should create an app without a warning
app_without_warning()

# Expectation: This should create an app with a warning
app_with_warning()
```

### sessionInfo()

```R
R version 4.3.2 (2023-10-31)
Platform: aarch64-apple-darwin20 (64-bit)
Running under: macOS Sonoma 14.2.1

Matrix products: default
BLAS:   /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib 
LAPACK: /Library/Frameworks/R.framework/Versions/4.3-arm64/Resources/lib/libRlapack.dylib;  LAPACK version 3.11.0

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

time zone: Asia/Kolkata
tzcode source: internal

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] teal_0.14.0.9040      testthat_3.2.1        teal.slice_0.4.0.9034 teal.data_0.3.0.9041 
[5] teal.code_0.5.0.9001  shiny_1.8.0          

loaded via a namespace (and not attached):
[...]

PR #1021, opened 2024-01-02
Pre-release activities

kartikeyakirar:
part of https://github.com/insightsengineering/nestdevs-tasks/issues/48

- [x] Review and update:
      - README.md (check example code)
      - NEWS.md
- [x] Run urlchecker::url_check() to identify broken links and fix
- [x] Review functions:
       - @example tag, make sure it runs, fix if otherwise
       - Make sure functions has @return tag to document the return value
       - no \dontrun tag, replace with if(interactive()) if needed
- [x] Sanity check of all vignettes, make sure there is no typo, no wrong format, etc.
- [x] Run R CMD check --as-cran make sure everything pass
- [x] Package Title is not duplicated in Package Description in DESCRIPTION file (e.g. this happens in teal.slice currently)
- [x] All package names in Title and Description fields of DESCRIPTION file are quoted with '
- [x] You have checked the Package Release Template https://github.com/insightsengineering/teal.reporter/pull/205/files
- [x] Make sure there are no ::: in examples
- [x] Make sure all teal.* mentions are lower-cased and quoted
- [x] Make Sure inst/WORDLIST is minimalized
- [x] Make sure non-exported functions do not have examples
- [x] Make sure each link to our documentation hosted with pkgdown on github pages do not have /main/ in the address but it has /latest/ instead, so we always expose the documentation of the latest release and not what's currently on main branch but not yet released
- [x] remove old rd syntax
- [x] switch from title case into sentence case for title and descriotion of functions.

PR #1070, opened 2024-01-25
Pass the check status to the main teal_data function call

vedhav:
Part of #1069 

Please make sure that you also use the `teal.data` branch `1069-fix-code-warnings-when-data-is-verified@main` for testing the example from the issue.

### DIFF
--- a/R/module_nested_tabs.R
+++ b/R/module_nested_tabs.R
@@ -140,7 +140,7 @@ ui_nested_tabs.teal_module <- function(id, modules, datasets, depth = 0L, is_mod
 
 #' @rdname module_nested_tabs
 srv_nested_tabs <- function(id, datasets, modules, is_module_specific = FALSE,
-                            reporter = teal.reporter::Reporter$new()) {
+                            reporter = teal.reporter::Reporter$new(), check = FALSE) {
   checkmate::assert_multi_class(modules, c("teal_modules", "teal_module"))
   checkmate::assert_class(reporter, "Reporter")
   UseMethod("srv_nested_tabs", modules)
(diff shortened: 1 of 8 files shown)


R/module_nested_tabs.R
```diff
@@ -291,9 +297,10 @@ srv_nested_tabs.teal_module <- function(id, datasets, modules, is_module_specifi
     get_datasets_code(datanames, datasets, hashes)
   )
 
+
   do.call(
     teal.data::teal_data,
-    args = c(data, code = list(code), join_keys = list(datasets$get_join_keys()[datanames]))
+    args = c(data, code = list(code), join_keys = list(datasets$get_join_keys()[datanames]), check = check)
```
gogonzo, line 303: This implementation causes double evaluation - each time when filter is changed.

If the object is verified it shouldn't be re-verified. Instead, `@verified` should be forced to the value of `@verified` slot from the object which has been passed to the application. How to pass this value from `srv_teal` to here?
- not through the modules argument. We have some hacky way to do this https://github.com/insightsengineering/teal/blob/b1bbe12b2e6655db7daac1fc4fc9433a45ee33f5/R/utils.R#L68 
Instead of passing 

```suggestion
    args = c(data, code = list(code), join_keys = list(datasets$get_join_keys()[datanames]))
```

gogonzo (changes requested):
Please see the comment below

R/module_nested_tabs.R
vedhav: Makes sense

gogonzo (commented):
👍

gogonzo (approved):

R/module_nested_tabs.R
```diff
@@ -291,10 +293,14 @@ srv_nested_tabs.teal_module <- function(id, datasets, modules, is_module_specifi
...
 
-  do.call(
+
+  data <- do.call(
     teal.data::teal_data,
     args = c(data, code = list(code), join_keys = list(datasets$get_join_keys()[datanames]))
   )
+
+  data@verified <- attr(datasets, "verification_status")
+  return(data)
```
chlebowa, line 303: Per NEST convention we don't use explicit returns at the end of functions.

vedhav: I'll squeeze this in the pre-release activities PR which looks for checks and conventions like these. Thanks!

vedhav: Fixed here https://github.com/insightsengineering/teal/pull/1021/commits/d11315ac1aec9d8703ec88b5c8ba447d926028c4

Issue #1019, opened 2023-12-22
update `init` documentation


chlebowa:
Some argument descriptions should be rephrased to be more clear.
The `id` parameter is completely opaque to me and it refers to a vignette which I believe no longer exists.

kartikeyakirar:
**Investigation:**

The` id` parameter was initially introduced during the early stages of development when the `init` function was a straightforward wrapper mainly for end-user. This function essentially delegates directly to` module_teal_with_splash.R`.

The primary purpose of the `id` parameter was to facilitate the embedding of two separate teal applications side-by-side. This concept is also applicable for integrating these applications into other `Shiny modules.`

To individually access elements within init, we utilize the `id `argument of the` init `function.

However, for above case it is also recommanded to use `ui_teal_with_splash` and `srv_teal_with_splash` directly for shiny modules.

now init with default `id = character(0)`. Instead of separately calling `ui_teal_with_splash` and `srv_teal_with_splash`, we use init with id = character(0).

An earlier example demonstrating how to display two teal apps side-by-side, or embed them in a larger application, is as follows: I am not able to run complete application as tabs and encoding are not visible

```
# Code to have two teal apps side-by-side or embed it in a larger application
create_teal_app <- function(title, id) {
  init(
    data = teal_data(
      new_iris = transform(iris, id = seq_len(nrow(iris))),
      new_mtcars = transform(mtcars, id = seq_len(nrow(mtcars))),
      code = "
      new_iris <- transform(iris, id = seq_len(nrow(iris)))
      new_mtcars <- transform(mtcars, id = seq_len(nrow(mtcars)))
    "
    ),
    modules = modules(
      module(
        label = "data source",
        server = function(input, output, session, data) {},
        ui = function(id, ...) div(p("information about data source")),
        datanames = "all"
      ),
      example_module(label = "example teal module"),
      module(
        "Iris Sepal.Length histogram",
        server = function(input, output, session, data) {
          output$hist <- renderPlot(
            hist(data()[["new_iris"]]$Sepal.Length)
          )
        },
        ui = function(id, ...) {
          ns <- NS(id)
          plotOutput(ns("hist"))
        },
        datanames = "new_iris"
      )
    ),
    title = title,
    filter = teal_slices(
      teal_slice(dataname = "new_iris", varname = "Species"),
      teal_slice(dataname = "new_iris", varname = "Sepal.Length"),
      teal_slice(dataname = "new_mtcars", varname = "cyl"),
[...]

PR #1071, opened 2024-01-25
updated  arg `id` description and `ui_teal_with_splash` description.

kartikeyakirar:
fixes https://github.com/insightsengineering/teal/issues/1019

I have revised the description of the id parameter to reflect its current usage, based on how it was previously utilized. Additionally, I have rephrased the description of ui_teal_with_splash for better clarity and understanding.

### DIFF
--- a/R/init.R
+++ b/R/init.R
@@ -34,11 +34,11 @@
 #'   The header of the app.
 #' @param footer (`shiny.tag` or `character(1)`)
 #'   The footer of the app.
-#' @param id (`character`)
-#'   module id to embed it, if provided,
-#'   the server function must be called with [shiny::moduleServer()];
-#'   See the vignette for an example. However, [ui_teal_with_splash()]
-#'   is then preferred to this function.
+#' @param id optional (`character`)
+#'   specifying the module id to be used when embedding the teal app within another `shiny` app as a module.
+#'   - If an id is provided, the server function must be called using [shiny::moduleServer()] with the specified id.
+#'   However, for better integration use [ui_teal_with_splash()] and [srv_teal_with_splash()].
+#'   - If no id is given (i.e., `id = character(0)`), the teal app runs as a standalone application without submodule integration.
 #'
 #' @return named list with server and UI function
 #'
@@ -227,9 +227,8 @@ init <- function(data,
   }
 
   # Note regarding case `id = character(0)`:
-  # rather than using `callModule` and creating a submodule of this module, we directly modify
+  # rather than using `shiny::moduleServer` and creating a submodule of this module, we directly modify
   # the UI and server with `id = character(0)` and calling the server function directly
-  # rather than through `callModule`
   res <- list(
     ui = ui_teal_with_splash(id = id, data = data, title = title, header = header, footer = footer),
     server = function(input, output, session) {
--- a/R/module_teal_with_splash.R
+++ b/R/module_teal_with_splash.R
@@ -1,16 +1,17 @@
 # This file adds a splash screen for delayed data loading on top of teal
 
-#' UI to show a splash screen in the beginning, then delegate to [srv_teal()]
+#' UI to show a splash screen in the beginning, then delegate to [ui_teal()]
 #'
 #' @description `r lifecycle::badge("stable")`
 #'
-#' The splash screen could be used to query for a password to fetch the data.
-#' [init()] is a very thin wrapper around this module useful for end-users which
+#' UI function offers the dynamic display that switches between a custom splash screen
+#' for delayed data loading and a default splash screen.This helps ensure smooth
+#' user interaction, particularly during the initial heavy data processing or
+#' when querying for a password to fetch data.
+#'
+#' [init()] is a very wrapper around this module useful for end-users which
 #' assumes that it is a top-level module and cannot be embedded.
 #' This function instead adheres to the `shiny` module conventions.
-#'
-#' If data is obtained through delayed loading, its splash screen is used. Otherwise,
-#' a default splash screen is shown.
 #'
 #' Please also refer to the doc of [init()].
 #'
--- a/man/init.Rd
+++ b/man/init.Rd
@@ -40,11 +40,13 @@ The header of the app.}
 \item{footer}{(\code{shiny.tag} or \code{character(1)})
 The footer of the app.}
 
-\item{id}{(\code{character})
-module id to embed it, if provided,
-the server function must be called with \code{\link[shiny:moduleServer]{shiny::moduleServer()}};
-See the vignette for an example. However, \code{\link[=ui_teal_with_splash]{ui_teal_with_splash()}}
-is then preferred to this function.}
+\item{id}{optional (\code{character})
+specifying the module id to be used when embedding the teal app within another \code{shiny} app as a module.
+\itemize{
+\item If an id is provided, the server function must be called using \code{\link[shiny:moduleServer]{shiny::moduleServer()}} with the specified id.
+However, for better integration use \code{\link[=ui_teal_with_splash]{ui_teal_with_splash()}} and \code{\link[=srv_teal_with_splash]{srv_teal_with_splash()}}.
+\item If no id is given (i.e., \code{id = character(0)}), the teal app runs as a standalone application without submodule integration.
+}}
 }
 \value{
 named list with server and UI function
--- a/man/srv_teal_with_splash.Rd
+++ b/man/srv_teal_with_splash.Rd
@@ -8,11 +8,13 @@ to \code{\link[=srv_teal]{srv_teal()}}.}
 srv_teal_with_splash(id, data, modules, filter = teal_slices())
 }
 \arguments{
-\item{id}{(\code{character})
-module id to embed it, if provided,
-the server function must be called with \code{\link[shiny:moduleServer]{shiny::moduleServer()}};
-See the vignette for an example. However, \code{\link[=ui_teal_with_splash]{ui_teal_with_splash()}}
-is then preferred to this function.}
+\item{id}{optional (\code{character})
+specifying the module id to be used when embedding the teal app within another \code{shiny} app as a module.
+\itemize{
+\item If an id is provided, the server function must be called using \code{\link[shiny:moduleServer]{shiny::moduleServer()}} with the specified id.
+However, for better integration use \code{\link[=ui_teal_with_splash]{ui_teal_with_splash()}} and \code{\link[=srv_teal_with_splash]{srv_teal_with_splash()}}.
+\item If no id is given (i.e., \code{id = character(0)}), the teal app runs as a standalone application without submodule integration.
+}}
 
 \item{data}{(\code{teal_data} or \code{teal_data_module})
 \code{teal_data} object as returned by \code{\link[teal.data:teal_data]{teal.data::teal_data()}} or \code{teal_data_module}.}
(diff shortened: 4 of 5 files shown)


R/init.R
```diff
@@ -34,11 +34,11 @@
...
 #' @param footer (`shiny.tag` or `character(1)`)
 #'   The footer of the app.
-#' @param id (`character`)
-#'   module id to embed it, if provided,
-#'   the server function must be called with [shiny::moduleServer()];
-#'   See the vignette for an example. However, [ui_teal_with_splash()]
-#'   is then preferred to this function.
+#' @param id optional (`character`)
+#'   specifying the module id to be used when embedding the teal app within another `shiny` app as a module.
+#'   - If an id is provided, the server function must be called using [`shiny::moduleServer()`] with the specified id.
```
gogonzo, line 39: I'm not sure if it is possible to call teal through `moduleServer`. Probably `callModule` would work instead. Have anyone tried to wrap teal output in other app?
 @averissimo 
 
```suggestion
#'   - If an id is provided, the server function must be called using [`shiny::moduleServer()`] with the specified id.
```

averissimo: It is, check out this "small" example

```r
teal_modules <- teal::modules(teal::example_module())

ui <- function(id) {
  ns <- shiny::NS(id)
  shiny::div(
    style = "background-color: #ffc6c4;",
    shiny::h1("This is a shiny app wrapper around teal"),
    shiny::hr(),
    shiny::div(
      style = "margin: 5em; border: 1px solid black; background-color: white",
      shiny::div(
        teal::ui_teal_with_splash(id = ns("teal_wrapper"), data = teal.data::teal_data())
      )
    )
  )
}

server <- function(id) {
    teal::srv_teal_with_splash(
      id = "teal_wrapper", data = teal.data::teal_data(iris = iris), modules = teal_modules
    )
  })
}

shiny::shinyApp(ui("app"), function(input, output, session) server("app"))
```

image

R/init.R
```diff
@@ -34,11 +34,11 @@
...
+#'   However, for better integration use [`ui_teal_with_splash()`] and [`srv_teal_with_splash()`].
+#'   - If no id is given (i.e., `id = character(0)`), the teal app runs as a standalone application without submodule integration.
```
chlebowa, line 41: This part it too much for argument description, please add "see `Details` and move it there.

vignettes/getting-started-with-teal.Rmd
chlebowa, line 1: The vignette becomes hugely inflated with no benefit. This is not a topic for an introduction. In fact, I don't think it is a worthy topic at all.

Do we even want to support `teal`-as-module in the long term? @pawelru

R/module_teal_with_splash.R
```diff
@@ -1,17 +1,18 @@
 # This file adds a splash screen for delayed data loading on top of teal
 
-#' UI to show a splash screen in the beginning, then delegate to [srv_teal()]
```
pawelru, line 3: is ticks required?
(there are more examples like this)

kartikeyakirar: As part of our pre-release, we have added backticks in all places where they were inconsistent in the documentation.

vignettes/getting-started-with-teal.Rmd
pawelru: > Do we even want to support teal-as-module in the long term? @pawelru

I'm not super convinced tbh. Even though I saw this already I don't think it's a good thing to do. Therefore I would discourage / not promote it.

chlebowa: Would you then agree that it should not be elaborated upon in the documentation of the coming release.
Maybe we should add a "legacy" note of some sort?

R/module_teal_with_splash.R
pawelru: Usually I saw just `[foo()]` and not ``[`foo()`]``. That's also in line with roxygen2 manyal - see example [here](https://roxygen2.r-lib.org/articles/index-crossref.html#see-also). Did you check if roxygen does not complain?

vignettes/getting-started-with-teal.Rmd
pawelru: Yes - I think it would be sensible for us not to advertise it further. `id` param description should be enough.

R/module_teal_with_splash.R
chlebowa: We did, it interprets both the same.

pawelru: OK so then let's leave it. It doesn't follow the standards but that's fine as long as roxygen does not complain.

vignettes/getting-started-with-teal.Rmd
averissimo: Only experienced developers will look for this functionality and from experience, the man pages were enough to find it :grin: 

I would just suggest that the "recommended" example on the vignette could move to the `srv_teal_with_splash` man page

kartikeyakirar: So this means, this PR doesn't require additional @details or newly added vignette examples ?

kartikeyakirar: > I would just suggest that the "recommended" example on the vignette could move to the `srv_teal_with_splash `man page

should we move below example under the `@examples` tag? WDYT @pawelru?

```
teal_modules <- modules(example_module())
# Shiny app with modular integration of teal
ui <- fluidPage(
  ui_teal_with_splash(id = "app1", data = teal_data())
)
  srv_teal_with_splash(
    id = "app1", 
    modules = teal_modules
  )
}
if (interactive()) {
  shinyApp(ui, server)
}
```

kartikeyakirar: removed vignetted example  here [2a25cf6](https://github.com/insightsengineering/teal/pull/1071/commits/2a25cf6d458a666692550e2cc232657df086ea0b) and added  example in `module_teal_with_splash`

R/module_teal_with_splash.R
```diff
@@ -1,23 +1,56 @@
...
+#'
+#' @details
+#' [`init()`] is a wrapper around this module useful for end-users which
 #' assumes that it is a top-level module and cannot be embedded.
 #' This function instead adheres to the `shiny` module conventions.
 #'
-#' If data is obtained through delayed loading, its splash screen is used. Otherwise,
-#' a default splash screen is shown.
-#'
 #' Please also refer to the doc of [init()].
```
chlebowa, line 17: This should be a `@seealso`.

kartikeyakirar: done [5ddf206](https://github.com/insightsengineering/teal/pull/1071/commits/5ddf2069424e87d30ea48644572862196c261eab)

R/init.R
```diff
@@ -36,11 +36,8 @@
 #'   The header of the app.
 #' @param footer (`shiny.tag` or `character(1)`)
 #'   The footer of the app.
-#' @param id (`character`)
-#'   module id to embed it, if provided,
-#'   the server function must be called with [shiny::moduleServer()];
-#'   See the vignette for an example. However, [ui_teal_with_splash()]
-#'   is then preferred to this function.
+#' @param id optional (`character`)
+#'   specifying the module id to be used when embedding the teal app within another `shiny` app as a module.
```
chlebowa, line 40:
```suggestion
#' @param id (`character`) 
#'   Optional string specifying the `shiny` for `teal` in cases it is used as a module 
#'   rather than a standalone `shiny` app. This is a legacy feature.
```

R/module_teal_with_splash.R
```diff
@@ -1,23 +1,56 @@
...
-#' [init()] is a very thin wrapper around this module useful for end-users which
+#' Module offers the dynamic display that switches between a custom splash screen
+#' for delayed data loading and a default splash screen.This helps ensure smooth
+#' user interaction, particularly during the initial heavy data processing or
+#' when querying for a password to fetch data.
+#'
+#' @details
+#' [`init()`] is a wrapper around this module useful for end-users which
 #' assumes that it is a top-level module and cannot be embedded.
 #' This function instead adheres to the `shiny` module conventions.
```
chlebowa, line 15:
```suggestion
#' Add splash screen to `teal` application.
#'
#' @description `r lifecycle::badge("stable")`
#'
#' Displays custom splash screen during initial delayed data loading.
#'
#' @details
#' This module pauses app initialization pending delayed data loading.
#' This is necessary because the filter panel and modules depend on the data to initialize.
#'
#' `teal_with_splash` follows the `shiny` module convention.
#' [`init()`] is a wrapper around this that assumes that `teal` it is
#' the top-level module and cannot be embedded.
#'
#' Note it is no longer recommended to embed `teal` in `shiny` apps as a module
#' but rather use `init` to create standalone application.
```

R/module_teal_with_splash.R
```diff
@@ -1,23 +1,56 @@
...
-#'
-#' Please also refer to the doc of [init()].
+#' @seealso [init()]
 #'
 #' @param id (`character(1)`)
 #'   module id
 #' @inheritParams init
+#' @param modules (`teal_modules`) object containing the output modules which
+#'   will be displayed in the `teal` application. See [modules()] and [module()] for
+#'   more details.
```
chlebowa, line 24: Isn't `modules` inherited from `init`? :raised_eyebrow:

R/module_teal_with_splash.R
```diff
@@ -1,23 +1,56 @@
...
+#' @inheritParams shiny::moduleServer
+#' @return
+#' - `ui_teal_with_splash` returns `HTML` for `shiny` UI module.
+#' - `srv_teal_with_splash` returns `reactive` containing `teal_data` object when data is loaded.
+#' If data is not loaded yet, `reactive` returns `NULL`.
```
chlebowa, line 29: I think describing the return value of the UI function is redundant, especially after mentioning `shiny` module convention. It's enough to say that the module's server functions returns whatever it returns.

chlebowa (changes requested):

R/module_teal_with_splash.R
kartikeyakirar: `module_teal_with_splash` accepts the object of class `teal_modules` only where as` init` accepts nested list of `teal_modules` or `teal_module` objects or a single `teal_modules` or `teal_module` object.

chlebowa: Right, I missed the difference.

R/module_teal_with_splash.R
```diff
@@ -1,23 +1,58 @@
...
 #' @param id (`character(1)`)
 #'   module id
 #' @inheritParams init
+#' @param modules (`teal_modules`) object containing the output modules which
+#'   will be displayed in the `teal` application. See [modules()] and [module()] for
+#'   more details.
+#' @inheritParams shiny::moduleServer
+#' @return
+#'   Returns `reactive` containing `teal_data` object when data is loaded.
+#'   If data is not loaded yet, `reactive` returns `NULL`.
```
chlebowa, line 31:
```suggestion
#' Returns `reactive` containing a `teal_data` object when data is loaded or `NULL` when it is not.
```

chlebowa (changes requested):
List specification for filter is no longer possible. Please modify these two sections accordingly:

```
#' Note: the older method of filter specification [...]
```
```
  ## `filter`
  checkmate::assert( [...]
  ```

chlebowa (approved):

PR #1082, opened 2024-02-01
Reduces the size of sub-directories

### LINKED ISSUE #1079: Reduce the size of sub-directories
As we have extensive docs for teal. The docs and help directory is populated with a lot of files.
We need to reduce the file size as we get a NOTE from R CMD check.

Possible solutions are to investigate unwanted files. Convert/Shrink them. Or remove and replace them with assets from a CDN.

averissimo:
# Pull Request

Fixes #1079

note: `base branch` is `1076_....` as that has a fix for a test that is failing. `main` branch doesn't have one of the images that is being optimized

### Changes description

- [x] Use a smaller logo file
- [x] Remove embedded PNG on SVG
- [x] Reduce gif


averissimo:
It already removes the NOTE, nevertheless I'll try to reduce the size of the gif

vignettes/images/teal-app-components-hover.svg
chlebowa, line 1: The printout in the image has extra space between rows as seen in the thumbnails here.
On my machine, on the other hand, it has no line breaks:
Screenshot from 2024-02-01 19-08-25


I only now noticed that the actual app image only appears on hover. This is not mentioned in the text and not obvious. I would expect app code to be followed by an app image.

averissimo:
`.webp` is a image standard that is supported by all browsers (Chrome, Firefox, Safari) since ~2020

### New `.webp` 0.787 MB (better quality)



### Old `.gif` 1.3 MB

vignettes/images/teal-app-components-hover.svg
averissimo: I'm not seeing that empty space locally nor when building the site

Here is how it looks with a caption

[zz.webm](https://github.com/insightsengineering/teal/assets/211358/26e1f501-3f06-4ef4-8ff8-ba9a1353ad99)

averissimo: Also added "Snapshot manager" placeholder and "hide filter" (edited video on comment)

vedhav: I think the snapshot manager would be too much here. The labels in the image are 1:1 representation of what we show below. We used to show "Encoding Panel" & "Module Output Panel" in place of "Active Module Content". But, we simplified it. Doing this would increase the complexity.
Like the info about the hover 👍🏽

vedhav:
`.webp` is 💙

vignettes/images/teal-app-components-hover.svg
chlebowa: Agreed, let's keep it simple.
Maybe the hamburger and cog should be mentioned in the text. It was pointed out that there are _no_ user-facing instructions on how to use the snapshot manager.

chlebowa: > I'm not seeing that empty space locally nor when building the site

It's fixed now. The image is different, it now has a white background, where it previously had a grey one, but the formatting is proper.

averissimo: You guys are right, I'll remove the placeholders from the wireframe.

However, I think it will look weird not to have the cog and burger (when hovering) as they are an integral part of the UI

vedhav: Yeah, The cog and hamburgers should stay in the app example snapshot.

vedhav: Good call on that one 👍🏽 Sorry I missed it when creating it.

chlebowa: The diagram shows the header bar, let the reader focus on that.

averissimo:
@vedhav I couldn't optimize the `drawio` files. What do you think about changing the source to `svg` as they are an open format and vectorial.

Inkscape can be used to edit.

There are also some online tools to edit on the browser

vignettes/images/teal-app-components-hover.svg
chlebowa: Pretty but not _entirely _ accurate :wink: 
The two icons are not elements of the filter panel. The filter panel UI is created by `FilteredData` in its entirety. The two icons are defined and place in `teal` if there is a filter panel present.

averissimo: Hahahah I know, I know :grin: 

I think we could let this slight inaccuracy pass as this is user-facing documentation and it makes sense to group with the filter panel

vedhav: Yes, I think this will be the closest to the functional understanding of the layouts. I like it @averissimo

vedhav: What do you think about adding the text between the screenshot and the code?
Along the lines of: you can hover to see the output generated by this code.
Now it's between the screenshot and area label definition but hovering should be referencing the code above.
I get that it's common to place these legends at the bottom of the image, But this ain't any common figure with a label so I prefer it at the top. Not pushing too hard. Okay with either, just slightly leaning towards upwards placement.

vignettes/getting-started-with-teal.Rmd
```diff
@@ -63,27 +63,30 @@ if (interactive()) {
   }
 </style>
 
-<!-- These images are generated using the diagrams inside `inst/design`  -->
+_Hovering_ the image shows the `teal` application generated by this code.
```
averissimo, line 66: @vedhav done here.

Also see section below that I reflowed by using sub-bullets

vignettes/images/teal-app-components-hover.svg
averissimo: Resolving this conversation as I think it is solved.

I've added the "pre-caption" and tagged Vedha there so we can move conversation to that context

vedhav:
I reduced the `drawio` size as much as I could.

vedhav (approved):
👌🏽

PR #1092, opened 2024-02-07
Update the options vignette

### LINKED ISSUE #1066: Investigate the exhaustiveness of the documentation about `options` in our vignettes
Part of #1013 check
The [vignettes/teal-options.Rmd](https://insightsengineering.github.io/teal/main/articles/teal-options.html) lists the different `options` one can use to modify the behavior of teal. Check if all of them are still valid or does it miss some options for example `options(teal.reporter.global_knitr)`

vedhav:
Closes #1066 

Changes:
1. Verified the current options we support.
    - ✅ teal (`teal.load_nest_code`, `teal.show_js_log`, `teal.bs_theme`)
    - ✅ teal.logger (`teal.log_level`)
    - ✅ teal.widgets (`teal.basic_table_args`, `teal.ggplot2_args`, `teal.plot_dpi`)
    - ✅ teal.slice (`teal.threshold_slider_vs_checkboxgroup`)
    - ❌ teal.modules.general (`ggplot2.discrete.colour`)
    - ❌ teal.modules.clinical (`ggplot2.discrete.colour`, `tmc_default_total_label`)
2. Removed the context of where the options were implemented, it seems unnecessary.
3. Reordered the options in the way that important and most used are at the top and are grouped when possible. And, moved the depreciated options to the bottom.

### DIFF
--- a/vignettes/teal-options.Rmd
+++ b/vignettes/teal-options.Rmd
@@ -1,6 +1,6 @@
 ---
 title: "Modifying a teal Application With R Options"
-author: "Paweł Rucki"
+author: "NEST CoreDev"
 output: rmarkdown::html_vignette
 vignette: >
   %\VignetteIndexEntry{Modifying a teal Application With R Options}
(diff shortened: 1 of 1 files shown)


m7pr (approved):
👍 looks good.
I would only think that `teal.log_level` is the most popular option... but probably for a teal core dev team member and not the app developer :)

vedhav:
> I would only think that teal.log_level is the most popular option... but probably for a teal core dev team member and not the app developer :)

Yeah, I had to prioritize app developer > module developer > framework developer.

PR #1094, opened 2024-02-07
update docs of module() server and ui args

pawelru:


R/modules.R
```diff
@@ -23,15 +23,16 @@
...
+#'     data specified in the `filters` argument.
+#'  - `datasets` (optional) If present, module will receive `FilteredData`. (See [`teal.slice::FilteredData`]).
+#'  - `reporter` (optional) If present, module will receive `Reporter`. (See [`teal.reporter::Reporter`]).
+#'  - `filter_panel_api` (optional) If present, module will receive `FilterPanelAPI`.
+#'     (See [`teal.slice::FilterPanelAPI`]).
+#'  - `...` (optional) If present, `server_args` elements will be passed to the module named argument or to the `...`.
 #' @param ui (`function`) `shiny` UI module function with following arguments:
 #'  - `id` - `teal` will set proper `shiny` namespace for this module.
-#'  - `...` (optional) `ui_args` elements will be passed to the module named argument or to the `...`.
+#'  - `...` (optional) If present, `ui_args` elements will be passed to the module named argument or to the `...`.
```
m7pr, line 35: @pawelru did you add `If present` just after the word `(optional)`? for every `optional` parameter?

pawelru: Yes. It's because the english doesn't sound well.
After a short break from this PR I think that the current version is not perfect either :P Currently it's more describing a _flag_ behaviour. Let me enhance it more.

R/module_teal_with_splash.R
```diff
@@ -1,6 +1,6 @@
 # This file adds a splash screen for delayed data loading on top of teal
 
-#' Add splash screen to `teal` application.
```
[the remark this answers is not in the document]
pawelru: I tracked these by looking at the reference manual -> https://insightsengineering.github.io/teal/latest-tag/reference/index.html

Shall I remove these dots as well?

R/modules.R
m7pr: `when provided`?

R/module_teal_with_splash.R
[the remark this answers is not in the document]
m7pr: yes. that would be awesome to have dots be removed everywhere. thanks

PR #1090, opened 2024-02-06
Use `teal.data::get_code` instead of `teal.data:::get_code_dependency`

m7pr:
After introduction of `...` in `teal.data::get_code()` during https://github.com/insightsengineering/teal.data/pull/290,
I reckon there is no point using non-exported `teal.data:::get_code_dependency` 
and we can just use exported `teal.data::get_code`.

Merge only when this is discussed and updated version of `teal.data` lands on CRAN https://github.com/insightsengineering/teal.data/issues/291

PR #1117, opened 2024-02-28
move `label` of `teal_module(s)` from list element to attribute

chlebowa:
In short: the `label` property of `teal_module` and `teal_modules` objects are moved from a list element to an attribute.
This will make it less of a headache to write recursive functions. Unfortunately, we will not be able to use `rapply` as it requires that the "bottom" class not be list-like, which `teal_module` is.

`teal_modules` is no longer a list of length 2 with `label` and `children` elements, the latter being a list of `teal_module` or `teal_modules`, but rather a list of `teal_module`/`teal_modules` objects with a `label` attribute. In essence, `teal_modules` is now what `teal_modules$children` was.
As a consequence, the `label` of a `teal_module` is moved to attribute accordingly.

Added function `module_label` for convenient extraction of the `label` attribute.

#### STATUS
Introduced core changes.
Tests pass.
There are probably places downstream that will be modified.

Conflicts with https://github.com/insightsengineering/teal/pull/1090 will be substantial.


m7pr:
kudos for starting on that!

PR #1118, opened 2024-03-01
name specific lintr in utils.R

m7pr:
Follow-up after https://github.com/insightsengineering/teal/pull/1090#discussion_r1507466823


chlebowa (approved):
:+1:

PR #1127, opened 2024-03-05
Introduce shinytest2

vedhav:


tests/testthat/test-shinytest2-teal.R
```diff
@@ -0,0 +1,368 @@
...
+    )
+  )
+  expect_true(
+    grepl(
+      app_footer,
+      app$get_html("footer") |>
+        rvest::read_html() |>
+        rvest::html_text()
+    )
+  )
```
gogonzo, line 83: Given the title "initializes with no errors" tests should stop on `expect_no_shiny_error(app)`
```suggestion
```

tests/testthat/test-shinytest2-teal.R
```diff
@@ -0,0 +1,368 @@
...
+  app$stop()
+})
+
+
+test_that("teal filters are initialized as expected", {
```
gogonzo, line 88: Few comments regarding this whole test_that  

1. Title is not precise (doesn't say anything about expectation). 
2. there are two apps testing the same thing - that filter panel initialized with values selected specified in teal_slice objects.

I suggest to test if:
     - there are cards which datanames are the same as module datanames (three variants module$datanames specified, module$datanames is NULL, identical(module$datanames, "all") 
     - if module receives filtered data

tests/testthat/test-shinytest2-teal.R
```diff
@@ -0,0 +1,368 @@
...
+    c(3, 4)
+  )
+  expect_identical(
+    get_active_selection_value(app, "mtcars", "gear"),
+    c("3", "4", "5")
+  )
+  app$stop()
+})
+
+test_that("module with a reporter creates the reporter tab", {
```
gogonzo, line 190: two apps in a single test_that. Please split into separate test_that and adjust the description to the expectation

tests/testthat/test-shinytest2-teal.R
```diff
@@ -0,0 +1,368 @@
...
+  app$click(selector = ".btn.action-button.filter_hamburger")
+  app$wait_for_idle(timeout = default_idle_timeout)
+  primary_attrs <- get_class_attributes(app, ".teal_primary_col")
+  secondary_attrs <- get_class_attributes(app, ".teal_secondary_col")
+
+  expect_true(grepl("col-sm-12", primary_attrs$class))
+  expect_true(grepl("display: none;", secondary_attrs$style))
+})
+
+test_that("check the teal app", {
```
gogonzo, line 266: Please name it precisely and write tests accordingly.
```suggestion
test_that("check the teal app", {
```

gogonzo (commented):
Imprecise tests and too long `test_that`. Please revisit tests and make as short test_that as possible. What is the overhead of starting headless browser here? Is it seconds or fractions of seconds?

gogonzo:
> It takes about 2 seconds to initialize theAppDriver object ...  I'm hesitant about starting them outside the test_that scope as we will add a skip_on_cran in every shinytest2 tests. I think ~2 seconds of app start is not that bad.

Let's do this inside test_that - I think it is more readable.

>  I see your point in splitting the tests into smaller and more precise test desc.

The point is to minimize time to detect failure and understand what is tested and what not. This file will soon have thousands of lines and imprecise test_that description (which doesn't describe expectation) will make it very hard for us developers to track the cause of the bug and monitor true-coverage. I suggest to be as precise as possible even if it affects performance (but in the same time shinytest2 can't take 10 minutes - or can?)

tests/testthat/test-shinytest2-teal.R
vedhav: The reason for having two test apps here was to test if the reporter appears only in a teal app with the reporter and it does not appear in a teal app without the reporter. I'll change the description so this expectation is clear.

tests/testthat/test-shinytest2-teal.R
vedhav: This was a local test spec. It should be completely removed.

tests/testthat/test-shinytest2-teal.R
```diff
@@ -0,0 +1,274 @@
...
+          teal.reporter::simple_reporter_ui(ns("reporter")),
+          selectInput(ns("dataname"), "Choose a dataset", choices = NULL)
+        ),
+        mainPanel(verbatimTextOutput(ns("dataset")))
+      )
+    }
+  )
+}
+
+test_that("teal app initializes with no errors", {
```
averissimo, line 37: prefix all tests with "shinytest2" or "e2e"?

tests/testthat/test-shinytest2-teal.R
```diff
@@ -0,0 +1,274 @@
+simple_teal_data <- within(teal_data(), {
+  iris <- iris
+  mtcars <- mtcars
+})
+datanames(simple_teal_data) <- c("iris", "mtcars")
+
+report_module <- function(label = "example teal module") {
```
averissimo, line 7: This function is only used once.

Do you want to re-use it in the future (move it to the helper file) or can it be included in the body of the test?

tests/testthat/test-shinytest2-teal.R
```diff
@@ -0,0 +1,274 @@
+simple_teal_data <- within(teal_data(), {
+  iris <- iris
+  mtcars <- mtcars
+})
+datanames(simple_teal_data) <- c("iris", "mtcars")
```
averissimo, line 5: I'm not a big fan of having objects being reused on multiple tests as they can be modified.

What do you think about changing this to a generator function and moved to the helper?

DESCRIPTION
```diff
@@ -60,6 +60,8 @@ Suggests:
     MultiAssayExperiment,
     R6,
     rmarkdown (>= 2.19),
+    rvest,
```
averissimo, line 63: `{glue}` needs to be added here, alternatively, we could use `{stringr}` and use `stringr::str_glue()` instead of `glue::glue` _(as you mentioned that `{stringr}` could be useful)_

They also need to be added below to the "Config/Need/verdepcheck"

`tidyverse/rvest, rstudio/shinytest2, tidyverse/glue`

```suggestion
    rvest,
    glue,
```

tests/testthat/helper-shinytest2.R
```diff
@@ -0,0 +1,176 @@
+library(shinytest2)
+library(glue)
+library(htmltools)
```
averissimo, line 3: Could we use `shiny::tags` instead of this library?

vedhav: `htmltools` is not used at all. I will remove it.

DESCRIPTION
gogonzo: We can also use `sprintf` which you probably noticed I prefer to use. This is why:

```r
library(glue)
val1 <- "elo1"
val2 <- "elo2"
val3 <- "elo3"

elo <- microbenchmark::microbenchmark(
  sprintf = sprintf("%s %s %s", val1, val2, val3),
  paste = paste(val1, val2, val3),
  glue = glue("{val1} {val2} {val3}")
)

ggplot2::autoplot(elo)
``` 

image

m7pr: wow

tests/testthat/helper-shinytest2.R
```diff
@@ -0,0 +1,223 @@
...
+  suppressWarnings(
+    AppDriver$new(
+      shinyApp(app$ui, app$server),
+      name = "teal",
+      variant = platform_variant(),
+      load_timeout = 300000,
+      seed = 123
+    )
+  )
+}
```
averissimo, line 49: I just had an issue when running the same test repeatedly in a short period.

What do you think about stopping the app automatically just like `{withr}` (maybe the name could also change to follow convention.

```suggestion
local_test_app_object <- function(data,
                                  modules,
                                  filter = teal_slices(),
                                  title = build_app_title(),
                                  header = tags$p(),
                                  footer = tags$p(),
                                  id = character(0),
                                  envir = parent.frame()) {
  app <- init(
    data = data,
    modules = modules,
    filter = filter,
    title = title,
    header = header,
    footer = footer,
    id = id
  )
  test_app <- suppressWarnings(
    shinytest2::AppDriver$new(
      shiny::shinyApp(app$ui, app$server),
      name = "teal",
      variant = shinytest2::platform_variant(),
      load_timeout = 300000,
      seed = 123
    )
  )

  if (requireNamespace("withr", quietly = TRUE)) {
    withr::defer(test_app$stop(), envir = envir)
  }

  test_app
}
```

DESCRIPTION
averissimo: `paste` on my laptop is almost on par with `sprintf`, but I'm switching teams to `sprintf` anyway :grin:

``` r
elo <- microbenchmark::microbenchmark(
  sprintf = sprintf("%s %s %s", val1, val2, val3), 
  paste = paste(val1, val2, val3), 
  glue = glue("{val1} {val2} {val3}"),
  strc_c = stringr::str_c(val1, val2, val3, sep = " "),
  glue2 = glue(val1, val2, val3, .sep = " "),
  times = 100L
) 
```

image

tests/testthat/helper-shinytest2.R
```diff
@@ -0,0 +1,223 @@
...
+      el <- rvest::read_html(x)
+      root_id <- el |>
+        html_node("ul") |>
+        html_attr("id") |>
+        gsub(pattern = "(^teal-main_ui-)|(-active_tab$)", replacement = "")
+      tab_id <- el |>
+        html_nodes("li a") |>
+        html_attr("data-value")
+      tab_name <- el |>
+        html_nodes("li a") |>
```
gogonzo, line 208: usage of a `|>` require to change `Depends: R (>=4.1)`

tests/testthat/test-shinytest2-teal.R
```diff
@@ -0,0 +1,357 @@
...
+  app <- get_test_app_object(
+    data = simple_teal_data,
+    modules = example_module(label = "Example Module")
+  )
+  app$wait_for_idle(timeout = default_idle_timeout)
+  expect_no_shiny_error(app)
+  app$stop()
+})
+
+test_that("UI parameters of init creates the expected UI", {
```
gogonzo, line 47:
```suggestion
test_that("init creates UI containing specified title, favicon, header and footer", {
```

tests/testthat/helper-shinytest2.R
```diff
@@ -0,0 +1,223 @@
...
+    NS(glue("teal-main_ui-root-{module_id}-module_filter_panel"))
+  )
+  app$click(
+    fns(
+      glue("active-{data_name}-filter-{data_name}_{var_name}-remove")
+    )
+  )
+}
+
+get_active_selection_value <- function(app, data_name, filter_name, is_numeric = FALSE, module_id = "") {
```
gogonzo, line 101: I think `get_filter_selection_value` is more accurate

```suggestion
get_filter_selection_value <- function(app, data_name, filter_name, is_numeric = FALSE, module_id = "") {
```

tests/testthat/test-shinytest2-teal.R
```diff
@@ -0,0 +1,357 @@
...
+      app_footer,
+      app$get_html("footer") |>
+        rvest::read_html() |>
+        rvest::html_text()
+    )
+  )
+  app$stop()
+})
+
+test_that("expected teal filters are initialized when module specific filters are created", {
```
gogonzo, line 96: I'm not sure if we should test filter values here. We should first confirm that `module_specific/mapping` distribute appropriate filters to the modules. So you just need to check if `#<id>` of the filter cards match `id` of the `teal_slice` element.

```suggestion
test_that("filters are distributed to modules according to mapping argument", {
```

gogonzo (commented):
Please see another round of comments. I think at least one test of the filter panel is missing. We need to test if `<teal_module>$server` receives filtered data. It means you need to test `teal_data` object in the `<teal_module>$server` namespace and see if for example `mtcars` is filtered properly.

```r
testthat::test_that("teal_module server receives reactive teal_data object with data filtered by active filters", { 
  ...
})
```

averissimo:
I've found it helpful to create a custom helper to define namespaces with an arbitrary number of arguments. Instead of building it manually every time.

This could be integrated with `get_active_ns` so that it returns a function instead of a string.

WDYT? @vedhav 

``` r
helper_ns <- function(id) {
  function(..., .css_prefix = "") {
    args <- rlang::list2(...)
    checkmate::assert_list(args, types = "character")
    (shiny::NS(paste0(.css_prefix, id)))(paste(args, collapse = "-"))
  }
}

ns <- helper_ns("teal-main_ui-filter_manager")

a_name <- "snapshot_manager"
output_id <- ns("filter_manager", a_name, "snapshot_add")
output_id
#> [1] "teal-main_ui-filter_manager-filter_manager-snapshot_manager-snapshot_add"

selector <- ns("filter_manager", "snapshot_manager", "snapshot_add", .css_prefix = "#")
selector
#> [1] "#teal-main_ui-filter_manager-filter_manager-snapshot_manager-snapshot_add"
```

DESCRIPTION
m7pr, line 63: Nowadays I think `rvest` is just a wrapper around `xml2`. Maybe we can work directly with `xml2`? See `namespace:xml2`

```r
> rvest::read_html
function (x, encoding = "", ..., options = c("RECOVER", "NOERROR", 
    "NOBLANKS")) 
{
    UseMethod("read_html")
}
<bytecode: 0x000002a3d09722e0>
<environment: namespace:xml2>
```

image

tests/testthat/helper-shinytest2.R
```diff
@@ -0,0 +1,254 @@
...
+}
+
+# add_filter_var(app, "mtcars", "wt")
+add_filter_var <- function(app, data_name, var_name) {
+  app$set_inputs(
+    !!sprintf(
+      "%s-add-%s-filter-var_to_add",
+      get_active_ns(app, "filter"),
+      data_name
+    ) := var_name
```
m7pr, line 224: If there is any way to omit `!!` and `:=` then I would appreciate it : D

m7pr: I guess it's unavoidable

vedhav:
> This could be integrated with get_active_ns so that it returns a function instead of a string.

Thanks @averissimo it can be a new `component` to that function. Do you want to do the honors or shall I add it?

tests/testthat/helper-shinytest2.R
vedhav: I tried to eliminate it for a while and was unsuccessful.

DESCRIPTION
vedhav: Nice suggestion. Let's do this before merging. `rvest` is easy to use so good for productivity.
The `rvest`/`xml2` functions are mostly going to live inside our helper functions so it won't affect the productivity of the test creation.

averissimo:
@vedhav I can take the honors :laughing:

DESCRIPTION
vedhav: Now we are using `sprintf` throughout 💪🏽

tests/testthat/helper-shinytest2.R
vedhav: Interesting. For me, the port crash never occurred. Ideally, I would term this as a bug in `shinytest2`. Let me try to recreate this.

vedhav:
@averissimo Thanks for branching off of this and providing the `helper_NS` and the namespace for the `filter_manager`
I've merged your changes here inside the R6 class. Please have a look.

tests/testthat/test-shinytest2-teal.R
```diff
@@ -0,0 +1,317 @@
...
+        example_module(label = "Nested 2"),
+        modules(
+          label = "Sub Nested Modules",
+          example_module(label = "Nested 1"),
+          example_module(label = "Nested 1")
+        )
+      )
+    )
+  )
+  app_modules <- get_app_module_tabs(app)
```
gogonzo, line 308: Can we move it to the object?
```suggestion
  app_modules <- app$get_app_module_tabs()
```

tests/testthat/test-shinytest2-teal.R
```diff
@@ -0,0 +1,317 @@
...
+  app$set_filter_selection_value("iris", "Species", "setosa")
+  app$navigate_teal_tab("Module_1")
+  app$wait_for_idle(timeout = default_idle_timeout)
+
+  expect_identical(
+    app$get_filter_selection_value("iris", "Species"),
+    "setosa"
+  )
+  app$stop()
+})
```
gogonzo, line 123:
```suggestion
test_that("teal initializes module specific filters according to mapping and respective teal_slice values", {
  # App with module specific filters
  app <- TealAppDriver$new(
    data = simple_teal_data(),
    modules = modules(
      example_module(label = "Module_1"),
      example_module(label = "Module_2")
    ),
    filter = teal_slices(
      teal_slice(id = "iris_species", dataname = "iris", varname = "Species", multiple = TRUE),
      teal_slice(id = "mtcars_cyl", dataname = "mtcars", varname = "cyl", selected = c(4, 6)),
      teal_slice(id = "mtcars_drat", dataname = "mtcars", varname = "drat", selected = c(3, 4)),
      teal_slice(id = "mtcars_gear", dataname = "mtcars", varname = "gear"),
      module_specific = TRUE,
      mapping = list(
        "Module_1" = c("iris_species", "mtcars_cyl"),
        "Module_2" = c("iris_species", "mtcars_drat", "mtcars_gear")
      )
    )
  )
  app$wait_for_idle(timeout = default_idle_timeout)

  expect_identical(app$get_active_data_filters("iris"), "Species")
  expect_identical(app$get_active_data_filters("mtcars"), "cyl")
  expect_identical(
    app$get_filter_selection_value("iris", "Species"),
    c("setosa", "versicolor", "virginica")
  )
  expect_identical(
    app$get_filter_selection_value("mtcars", "cyl"),
    c("4", "6")
  )
  expect_null(app$get_filter_selection_value("mtcars", "drat", is_numeric = TRUE))
  expect_null(app$get_filter_selection_value("mtcars", "gear"))

  app$navigate_teal_tab("Module_2")
  app$wait_for_idle(timeout = default_idle_timeout)

  expect_identical(app$get_active_data_filters("iris"), "Species")
  expect_identical(app$get_active_data_filters("mtcars"), c("drat", "gear"))
  expect_identical(
    app$get_filter_selection_value("iris", "Species"),
    c("setosa", "versicolor", "virginica")
  )
  expect_identical(
    app$get_filter_selection_value("mtcars", "drat", is_numeric = TRUE),
    c(3, 4)
  )
  expect_identical(
    app$get_filter_selection_value("mtcars", "gear"),
    c("3", "4", "5")
  )
  expect_null(app$get_filter_selection_value("mtcars", "cyl"))
  app$stop()
})
```

tests/testthat/test-shinytest2-teal.R
```diff
@@ -0,0 +1,317 @@
...
+  app$wait_for_idle(timeout = default_idle_timeout)
+
+  expect_identical(
+    app$get_active_filter_vars(),
+    c("iris", "mtcars")
+  )
+  app$stop()
+})
+
+test_that("filter panel is not created when datanames is NULL", {
```
gogonzo, line 271:
```suggestion
test_that("filter panel is not displayed when datanames is NULL", {
```

R/TealAppDriver.R
```diff
@@ -0,0 +1,393 @@
...
+        read_html() %>%
+        html_nodes(".filter_panel_dataname") %>%
+        html_text()
+      available_datasets[displayed_datasets_index]
+    },
+    #' @description
+    #' Get the active filter variables from a dataset in the `teal` app.
+    #'
+    #' @param dataset_name (character) The name of the dataset to get the filter variables from.
+    get_active_data_filters = function(dataset_name) {
```
gogonzo, line 233: maybe get_active_filters should return a list:
- dataname
- varname
- selected

Then it will be easier in tests to compare expectations (less code to write)

gogonzo (commented):
Round 3 👍

R/TealAppDriver.R
vedhav: Nice idea, I agree.

R/TealAppDriver.R
```diff
@@ -0,0 +1,427 @@
...
+    #' @description
+    #' Get the active shiny name space for the Module content and the Filter panel.
+    #' Note that in the case of the filter panel, the name space is constant when it is not moudle specific.
+    #' However, module specific filter panel will have the name space linked with the module name space.
+    #'
+    #' @param component (character) The component to get the active name space for.
+    #' Currently supported components are "module", "filter_panel", and "filter_manager".
+    #'
+    #' @return The active shiny name space of the component.
+    get_active_ns = function(component) {
```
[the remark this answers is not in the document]
vedhav: I like these getters 👍🏽

R/TealAppDriver.R
```diff
@@ -0,0 +1,427 @@
...
+    #'
+    #' @return The `TealAppDriver` object invisibly.
+    add_filter_var = function(dataset_name, var_name) {
+      self$app$set_inputs(
+        !!sprintf(
+          "%s-add-%s-filter-var_to_add",
+          self$get_active_ns("filter_panel"),
+          dataset_name
+        ) := var_name
+      )
```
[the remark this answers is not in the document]
m7pr: doesn't do.call require to put parameters in a list, which does not resolve the issue of the name of the parameter being a pasted string?

R/TealAppDriver.R
```diff
@@ -0,0 +1,427 @@
...
+      self$app$get_value(input = sprintf("%s-%s", self$get_active_ns("module"), input_id))
+    },
+    #' @description
+    #' Get the output from the module in the `teal` app.
+    #' This function will only access outputs from the name space of the current active teal module.
+    #'
+    #' @param output_id (character) The shiny output id to get the value from.
+    #'
+    #' @return The value of the shiny output.
+    get_module_output = function(output_id) {
```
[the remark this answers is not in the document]
vedhav: It depends, When it's a `textOutput` it returns a string, when it's a `uiOutput` it returns a list with html and it's deps.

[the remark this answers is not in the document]
vedhav: I think it's worth providing a valid doc for what will be returned in different cases.

R/TealAppDriver.R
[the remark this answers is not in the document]
gogonzo: @m7pr yes:
```r
do.call(
  self$app$set_inputs, 
  setNames(
    list(varname), 
    sprintf(
       "%s-add-%s-filter-var_to_add",
       self$get_active_ns("filter_panel"),
       dataset_name
      )
   )
)
```

R/TealAppDriver.R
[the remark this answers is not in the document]
gogonzo: got it, cool 👍

R/TealAppDriver.R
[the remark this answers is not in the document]
vedhav: Do you mean providing the ability to add multiple vars together?

I'm not fully convinced if `do.call` is the way to do this. Primarily because this encourages people to store their vars in a variable and use it and this can be hard to debug when the test will fail. So, I would like an implementation that involves allowing people to explicitly set the inputs that they wish to change [like how it's done in `AppDriver$set_inputs()`](https://rstudio.github.io/shinytest2/reference/AppDriver.html#method-set-inputs-)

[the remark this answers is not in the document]
vedhav: Looks like I was thinking of something else entirely. Thanks for the code example. This looks good.

tests/testthat/test-shinytest2-teal.R
vedhav: `get_app_module_tabs` is a function from the test helper file. There is no real use for people to get this value once we test this inside teal. So, I did not move this inside the R6 class.
I think we can even remove this function completely and just implement this logic inside one test. WDYT?

R/TealAppDriver.R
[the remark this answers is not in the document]
vedhav: Added a public [`set_input`](https://github.com/insightsengineering/teal/pull/1127/commits/2d13df0b80320f21551c964094c072e2867f8004) method to help set inputs based on the suggestion. Will replace all the instances of `AppDriver$set_inputs` with this wrapper.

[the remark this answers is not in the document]
m7pr: thanks, this removes the `!!` and `:=`. `setNames` now looks so obvious : P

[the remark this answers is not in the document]
gogonzo: @vedhav @m7pr it is more about operators which I don't like. We avoided any non-base operators whenever it is possible on production (except magrittr `%>%` which is used in reproducible modules code).

However, it is 2024 and I think I should move on ;)

```r
library(rlang)
foo <- function(...) return(list2(...))
arg_name <- "a"

microbenchmark::microbenchmark(
  rlang_bang2 = foo(!!arg_name := TRUE),
  do.call = do.call(foo, setNames(list(TRUE), arg_name)),
  times = 10000L
)
# Unit: nanoseconds
#         expr  min   lq     mean median   uq   max neval cld
#  rlang_bang2  902  984 1106.857    984 1025 38991 10000  a 
#      do.call 1599 1681 1953.859   1722 1886 27265 10000   b
```

👍

tests/testthat/helper-shinytest2.R
vedhav: Well now I know :)

tests/testthat/test-shinytest2-teal.R
gogonzo: Ok, makes sense to keep it outside of the class 👍

tests/testthat/helper-shinytest2.R
```diff
@@ -0,0 +1,115 @@
+library(shinytest2)
+library(rvest)
+
+default_idle_timeout <- 20000
+
+
+default_idle_timeout <- 20000
```
chlebowa, line 7:
```suggestion
default_idle_timeout <- 20000
```

tests/testthat/test-shinytest2-teal.R
```diff
@@ -0,0 +1,317 @@
...
+  app$navigate_teal_tab("Module_1")
+  app$wait_for_idle(timeout = default_idle_timeout)
+
+  expect_identical(
+    app$get_filter_selection_value("iris", "Species"),
+    "setosa"
+  )
+  app$stop()
+})
+
```
gogonzo, line 124: There are following missing tests but descriptions might be slightly different. Generally items in module-specific filter panel are independent, but some filters can be shared. If filter is mapped into multiple modules then it is the same filter (same object, same pointer). This means that changing values in one is reflected in another. 
It is also possible to add a new filter with the same variable/values and map it to single module - then the other (theoretically duplicated) filter is not affected. 

testthat::test_that("setting values of shared module-specific items are reflected in all modules", {})
testthat::test_that("setting values of independent module-specific items are limit only to this module", {})

R/TealAppDriver.R
```diff
@@ -0,0 +1,410 @@
...
+            "#%s-active-filter_active_vars_contents > span",
+            self$get_active_ns("filter_panel")
+          )
+        ),
+        function(x) {
+          style <- x %>%
+            rvest::read_html() %>%
+            rvest::html_node("span") %>%
+            rvest::html_attr("style")
+          style <- ifelse(is.na(style), "", style)
```
chlebowa, line 229: Please don't use `ifelse` like this unless it is absolutely impossible for `style` to be of `length != 1`.

R/TealAppDriver.R
```diff
@@ -0,0 +1,410 @@
...
+    set_module_input = function(input_id, value) {
+      self$app$set_inputs(
+        !!sprintf("%s-%s", self$get_active_ns("module"), input_id) := value
+      )
+      invisible(self$app)
+    },
+    #' @description
+    #' Get the active datasets that can be accessed via the filter panel of the current active teal module.
+    get_active_filter_vars = function() {
+      displayed_datasets_index <- sapply(
```
chlebowa, line 217: `vapply` > `sapply`

R/TealAppDriver.R
```diff
@@ -0,0 +1,410 @@
...
+        function(x) {
+          style <- x %>%
+            rvest::read_html() %>%
+            rvest::html_node("span") %>%
+            rvest::html_attr("style")
+          style <- ifelse(is.na(style), "", style)
+          style != "display: none;"
+        }
+      ) %>%
+        unname()
```
chlebowa, line 233: `sapply(USE.NAMES = FALSE)`

chlebowa (commented):
round one

tests/testthat/helper-shinytest2.R
```diff
@@ -0,0 +1,115 @@
...
+      )
+    }
+  )
+}
+
+#' Get all the module tabs in the `AppDriver` object along with their
+#' root_ids and module_ids used for name spacing in the shiny ui.
+get_app_module_tabs <- function(app) {
+  lapply(
+    app$get_html(selector = "ul.shiny-bound-input"),
```
averissimo, line 52: I'm concerned that this is too generic. Should make the selector more specific

```suggestion
    app$get_html(selector = "ul.nav.shiny-tab-input.shiny-bound-input"),
```

Or search for `ul` with id that ends on `active_tab`

```suggestion
    app$get_html(selector = "ul.shiny-bound-input[id$='active_tab']")
```

R/utils.R
```diff
@@ -367,3 +367,16 @@ defunction <- function(x) {
...
+#' Get unique labels
+#'
+#' Get unique labels for the modules to avoid namespace conflicts.
+#'
+#' @param labels (`character`) vector of labels
+#'
+#' @return (`character`) vector of unique labels
+#'
+#' @keywords internal
+get_unique_labels <- function(labels) {
```
chlebowa, line 380: I prefer `unique_labels` myself.

R/TealAppDriver.R
```diff
@@ -0,0 +1,410 @@
...
+    #' @param filter (`teal_slices`)
+    #'   Specifies the initial filter using [teal_slices()].
+    #' @param title (`shiny.tag` or `character(1)`)
+    #'   The browser window title. Defaults to a title "teal app" with the icon of NEST.
+    #'   Can be created using the `build_app_title()` or
+    #'   by passing a valid `shiny.tag` which is a head tag with title and link tag.
+    #' @param header (`shiny.tag` or `character(1)`)
+    #'   The header of the app.
+    #' @param footer (`shiny.tag` or `character(1)`)
+    #'   The footer of the app.
```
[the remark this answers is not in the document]
vedhav: I like it! I was today days old when I found we cannot inherit params into R6 method docs.

R/TealAppDriver.R
```diff
@@ -0,0 +1,410 @@
...
+    #' Set the input in the module in the `teal` app.
+    #' This function will only set inputs in the name space of the current active teal module.
+    #'
+    #' @param input_id (character) The shiny input id to get the value from.
+    #' @param value The value to set the input to.
+    #'
+    #' @return The `TealAppDriver` object invisibly.
+    set_module_input = function(input_id, value) {
+      self$app$set_inputs(
+        !!sprintf("%s-%s", self$get_active_ns("module"), input_id) := value
```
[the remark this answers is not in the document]
vedhav: Yeah, Dawid suggested it and I'll replace in all the other places. We have a method `set_input` that will do this.

R/TealAppDriver.R
vedhav: 💯

R/TealAppDriver.R
vedhav: Thanks! I'm so rusty haha. Haven't used `apply` in years after I started to use `purrr`

R/TealAppDriver.R
```diff
@@ -0,0 +1,410 @@
...
+    #' Navigate the teal tabs in the `teal` app.
+    #'
+    #' @param tabs (character) The tabs to navigate to. The order of the tabs is important,
+    #' and it should start with the most parent level tab.
+    #' Note: In case the teal tab group has duplicate names, the first tab will be selected,
+    #' If you wish to select the second tab with the same name, use the suffix "_1".
+    #' If you wish to select the third tab with the same name, use the siffix "_2" and so on.
+    #'
+    #' @return The `TealAppDriver` object invisibly.
+    navigate_teal_tab = function(tabs) {
```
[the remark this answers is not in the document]
vedhav: I tested this for triple nested tabs and it worked as expected:
```r
app <- TealAppDriver$new(
  teal_data(iris = iris, mtcars = mtcars),
  modules(
    example_module("Module 1"),
    example_module("Module 2"),
    modules(
      label = "Nested modules",
      example_module("Module 3"),
      example_module("Module 4"),
      modules(
        label = "Sub nested modules",
        example_module("Module 5"),
        example_module("Module 6")
      )
    )
  )
)
app$view()

app$navigate_teal_tab("Module 2")
app$navigate_teal_tab(c("Nested modules", "Module 4"))
app$navigate_teal_tab(c("Nested modules", "Sub nested modules", "Module 6"))
```

R/TealAppDriver.R
```diff
@@ -0,0 +1,410 @@
...
+      root <- "root"
+      for (tab in tabs) {
+        self$set_input(
+          sprintf("teal-main_ui-%s-active_tab", root),
+          get_unique_labels(tab),
+          wait_ = FALSE
+        )
+        root <- sprintf("%s-%s", root, get_unique_labels(tab))
```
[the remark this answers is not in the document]
vedhav: Thanks, We store the `modules` and `filters` to get the modules metadata and teal_slices object. Will use them effectively. Also, I think the unique labels are created in the server logic which could be avoided and created during the `modules()` creation.

R/TealAppDriver.R
```diff
@@ -0,0 +1,410 @@
...
+            self$get_active_ns("filter_panel")
+          )
+        ),
+        function(x) {
+          style <- x %>%
+            rvest::read_html() %>%
+            rvest::html_element("span") %>%
+            rvest::html_attr("style")
+          style <- ifelse(is.na(style), "", style)
+          style != "display: none;"
```
averissimo, line 230: Is this the only possible style that it can have? I don't think that this is a good solution, but my only alternative is to run javascript code

```r
      css_ns <- self$helper_NS(
        self$get_active_ns("filter_panel"), "active", "filter_active_vars_contents"
      )
      displayed_datasets_index <- app$get_js(
        sprintf(
          "Array.from(document.querySelectorAll(\"%s\")).map((el) => window.getComputedStyle(el).display == 'none')",
          ns(.css_prefix = "#", .css_suffix = " > span")
        )
      ) |> unlist()
```

R/TealAppDriver.R
```diff
@@ -0,0 +1,410 @@
...
+      expect_null(
+        self$app$get_html(".shiny-output-error-validation"),
+        info = "No validation error is observed"
+      )
+    },
+    #' @description
+    #' Check if the app has validation errors. This checks for global shiny validation errors.
+    expect_validation_error = function() {
+      expect_true(
+        !is.null(self$app$get_html(".shiny-output-error-validation")),
```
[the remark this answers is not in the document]
m7pr: 👍

R/TealAppDriver.R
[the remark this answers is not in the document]
m7pr: > Also add a backstop for requireNamespace("testthat") somewhere.

Isn't it gonna be run only during `devtools::test()`  and R CMD CHECK which first tries to `library(testthat)` in `tests/testthat.R`? So we're covered for `testthat` as we are for other regular tests.

R/TealAppDriver.R
```diff
@@ -0,0 +1,410 @@
...
+          ...
+        )
+      )
+      private$set_active_ns()
+      self$app
+    },
+    #' @description
+    #' Check if the app has shiny errors. This checks for global shiny errors.
+    #' Note that any shiny errors dependant on shiny server render will only be captured after the teal module tab
+    #' is visited because shiny will not trigger server computations when the tab is invisible.
```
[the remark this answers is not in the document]
vedhav: [suspendWhenHidden](https://shiny.posit.co/r/reference/shiny/0.12.0/outputoptions) is an inbuilt shiny optimization which does not run server logic dependent on hidden input elements. I was talking about this feature.

R/TealAppDriver.R
vedhav: Yeah, not proud of this test. Changing style will definitely break it. I'll think about other ways too. Thanks for the JS code.

vedhav:

averissimo:

averissimo:
> Apparently I can't say mean stuff while agreeing to the CLA :(

@cicdguy that's no fun!!

Before I signed the list of contributors was badly formatted with you and Vedha's user. 

image

m7pr:

.github/workflows/check.yaml
```diff
@@ -10,9 +10,11 @@ on:
       - ready_for_review
     branches:
       - main
+      - 503-introduce-shinytest2@main
   push:
     branches:
       - main
+      - 503-introduce-shinytest2@main
```
averissimo, line 17: This must be reverted before merging to main

averissimo (changes requested):
Blocking with reminder to remove CI changes before merging to main

R/TealAppDriver.R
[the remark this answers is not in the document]
chlebowa: Oh yeah, I would assume one would just pass `c("Module 1", "Module 6")`.

R/TealAppDriver.R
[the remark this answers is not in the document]
chlebowa: Note we do prefix `testthat` functions in all packages, `teal` included.

R/TealAppDriver.R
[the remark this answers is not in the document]
chlebowa: No, `names(submodules) <- get_unique_labels(labels)` is done in `modules` and `append_module`.

R/TealAppDriver.R
vedhav: Extended `get_active_data_filters` to handle the case when `dataset_name` is `NULL`. When it is NULL it will return a list of all the active filter vars for all the datanames.

tests/testthat/helper-shinytest2.R
vedhav: Thanks. Must have been some bad merge.

R/TealAppDriver.R
```diff
@@ -0,0 +1,425 @@
...
+          style <- x %>%
+            rvest::read_html() %>%
+            rvest::html_element("span") %>%
+            rvest::html_attr("style")
+          style <- ifelse(is.na(style), "", style)
+          style != "display: none;"
+        },
+        logical(1)
+      ) %>%
+        unname()
```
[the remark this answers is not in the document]
vedhav: I just made this change ;)

R/TealAppDriver.R
```diff
@@ -0,0 +1,425 @@
...
+    #' Get the active shiny name space for the Module content and the Filter panel.
+    #' Note that in the case of the filter panel, the name space is constant when it is not module specific.
+    #' However, module specific filter panel will have the name space linked with the module name space.
+    #'
+    #' @param component (character) The component to get the active name space for.
+    #' Currently supported components are "module", "filter_panel", and "filter_manager".
+    #'
+    #' @return The active shiny name space of the component.
+    get_active_ns = function(component) {
+      switch(component,
```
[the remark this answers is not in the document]
averissimo: Or let's discuss alternative to trigger this call :-)

vedhav:
@averissimo What do you think about writing a wrapper over `TealAppDriver$helper_NS` method as we're constantly using it for:
```r
ns <- app$helper_NS(
    app$filter_manager_ns(),
    "filter_manager",
    "snapshot_manager"
  )
```
Would be nice to do something simple like this:
```r
ns <- NS(app$snapshot_manager_ns())
```

Also, it feels weird to have two namespaces with the same id `filter_manager`. Perhaps there is something we can change in teal? Is there a good reason to have these two namespaces?

R/TealAppDriver.R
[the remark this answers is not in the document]
vedhav: I made changes to the `get_active_ns` as per the previous comments in the commit https://github.com/insightsengineering/teal/pull/1127/commits/22df0a2cccb1832c1abb32e8b8022eae31a6eb10

tests/testthat/helper-shinytest2.R
averissimo: I'm resolving this as tests now have a `app$stop()` call.

Maybe we could still use this strategy or add a `withr::defer(app$stop())` to make sure that the app stops cleanly even if there are errors.

R/TealAppDriver.R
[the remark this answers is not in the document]
averissimo: The problem still persists with teal_data_modules

What do you think about changing `active_ns` to an active binding and using it instead of `private$ns` directly?

```r
    #' @description
    #' Get the active shiny name space for interacting with the module content.
    #'
    #' @return (`string`) The active shiny name space of the component.
    active_module_ns = function() {
      self$active_ns$module
    },
    #' @description
    #' Get the active shiny name space for interacting with the filter panel.
    #'
    #' @return (`string`) The active shiny name space of the component.
    active_filters_ns = function() {
      self$active_ns$filter_panel
    },
    #' @description
    #' Get the active shiny name space for interacting with the filter panel.
    #'
    #' @return (`string`) The active shiny name space of the component.
    filter_manager_ns = function() {
      self$active_ns$filter_manager
    },
   # ...
  ),
  active = list(
    active_ns = function() {
      if (all(vapply(private$ns, length, integer(1)) == 0)) {
        private$set_active_ns()
      }
      private$ns
    }
  ),
```

averissimo:
> @averissimo What do you think about writing a wrapper over `TealAppDriver$helper_NS` method as we're constantly using it for:

@vedhav sounds good! Right now all it's being used the same as `NS` so we can scrap it :+1:

I'm creating a PR

R/TealAppDriver.R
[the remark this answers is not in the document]
vedhav: Can you try this. Making the `set_active_ns` public and calling it once the teal app is initialised after the DDL.

Here's an example app after you make it public
```r
tdm <- teal_data_module(
  ui = function(id) {
    ns <- NS(id)
    actionButton(ns("submit"), label = "Load data")
  },
  server = function(id) {
      eventReactive(input$submit, {
        data <- within(
          teal_data(),
          {
            dataset1 <- iris
            dataset2 <- mtcars
          }
        )
        datanames(data) <- c("dataset1", "dataset2")
        data
      })
    })
  }
)

app <- TealAppDriver$new(
  data = tdm,
  modules = example_module(label = "Example Module")
)
app$view()

app$active_ns()
app$click("teal_data_module-submit")
app$set_active_ns()
app$active_ns()

app$stop()
```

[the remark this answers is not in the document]
averissimo: Do you think `set_active_ns` should be public? If so then why call it explicitly here only and not while initialization or after navigation?

This might protect the `active_ns` from other (future) cases when it's not initialized.

R/TealAppDriver.R
```diff
@@ -0,0 +1,436 @@
...
+        super$initialize(
+          shinyApp(app$ui, app$server),
+          name = "teal",
+          variant = platform_variant(),
+          ...
+        )
+      )
+
+      private$set_active_ns()
+    },
```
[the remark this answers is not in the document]
averissimo:
```diff
diff --git a/R/TealAppDriver.R b/R/TealAppDriver.R
index 5902e04ad..a5e9ff7ad 100644
--- a/R/TealAppDriver.R
+++ b/R/TealAppDriver.R
@@ -26,6 +26,7 @@ TealAppDriver <- R6::R6Class( # nolint
                           title = build_app_title(),
                           header = tags$p(),
                           footer = tags$p(),
+                          timeout = 20000,
                           ...) {
       private$data <- data
       private$modules <- modules
@@ -43,11 +44,13 @@ TealAppDriver <- R6::R6Class( # nolint
           shinyApp(app$ui, app$server),
           name = "teal",
           variant = platform_variant(),
+          timeout = timeout,
           ...
         )
       )
 
       private$set_active_ns()
+      self$wait_for_idle()
     },
     #' @description
     #' Check if the app has shiny errors. This checks for global shiny errors.
```

tests/testthat/test-shinytest2-init.R
```diff
@@ -0,0 +1,58 @@
...
+    ),
+    header = app_header,
+    footer = app_footer
+  )
+  app$wait_for_idle(timeout = default_idle_timeout)
+
+  testthat::expect_equal(
+    app$get_html("head > title")[1] %>%
+      rvest::read_html() %>%
+      rvest::html_text(),
```
averissimo, line 31:
```suggestion
    app$get_text("head > title")[1],
```

tests/testthat/test-shinytest2-init.R
```diff
@@ -0,0 +1,58 @@
...
+    )
+  )
+  testthat::expect_true(
+    grepl(
+      app_footer,
+      app$get_html("footer") %>%
+        rvest::read_html() %>%
+        rvest::html_text()
+    )
+  )
```
averissimo, line 56:
```suggestion
  testthat::expect_match(
    app$get_text("footer"),
    app_footer
  )
```

tests/testthat/test-shinytest2-init.R
```diff
@@ -0,0 +1,58 @@
...
+    app_favicon
+  )
+  testthat::expect_true(
+    grepl(
+      app_header,
+      app$get_html("header") %>%
+        rvest::read_html() %>%
+        rvest::html_text()
+    )
+  )
```
averissimo, line 48:
```suggestion
  testthat::expect_match(
    app$get_text("header"),
    app_header
  )
```

tests/testthat/test-shinytest2-modules.R
```diff
@@ -0,0 +1,127 @@
...
+    app$get_html(selector = "ul.shiny-bound-input"),
+    function(x) {
+      el <- rvest::read_html(x)
+      el %>%
+        rvest::html_elements("li a") %>%
+        rvest::html_text()
+    },
+    USE.NAMES = FALSE
+  ) %>%
+    unlist()
```
averissimo, line 118:
```suggestion
  app_modules <- app$get_text(selector = "ul.shiny-bound-input li a")
```

tests/testthat/test-shinytest2-reporter.R
```diff
@@ -0,0 +1,41 @@
...
+      rvest::html_text()
+  )
+  teal_tabs <- app_without_reporter$get_html(selector = "#teal-main_ui-root-active_tab") %>%
+    rvest::read_html() %>%
+    rvest::html_elements("a")
+  non_reporter_tabs <- setNames(
+    teal_tabs %>%
+      rvest::html_attr("data-value"),
+    teal_tabs %>%
+      rvest::html_text()
```
averissimo, line 27: I don't think we should use pipe on such simple 1 line calls

```suggestion
    rvest::html_attr(teal_tabs, "data-value"),
    rvest::html_text(teal_tabs)
```

tests/testthat/test-shinytest2-reporter.R
```diff
@@ -0,0 +1,41 @@
...
+  )
+
+  teal_tabs <- app_with_reporter$get_html(selector = "#teal-main_ui-root-active_tab") %>%
+    rvest::read_html() %>%
+    rvest::html_elements("a")
+  reporter_tabs <- setNames(
+    teal_tabs %>%
+      rvest::html_attr("data-value"),
+    teal_tabs %>%
+      rvest::html_text()
```
averissimo, line 18:
```suggestion
    rvest::html_attr(teal_tabs, "data-value"),
    rvest::html_text(teal_tabs)
```

tests/testthat/test-shinytest2-teal_slices.R
```diff
@@ -0,0 +1,101 @@
...
+    c("4", "6")
+  )
+  testthat::expect_identical(
+    app$get_active_filter_selection("mtcars", "drat", is_numeric = TRUE),
+    c(3, 4)
+  )
+  testthat::expect_identical(
+    app$get_active_filter_selection("mtcars", "gear"),
+    c("3", "4", "5")
+  )
```
averissimo, line 35: (minor comment) Consider using `testthat::expect_setequal` instead of `expect_identical` just in case the order changes in the future.

tests/testthat/test-shinytest2-utils.R
```diff
@@ -0,0 +1,33 @@
...
+
+  get_class_attributes <- function(app, selector) {
+    element <- app$get_html(selector = selector) %>%
+      rvest::read_html() %>%
+      rvest::html_elements(selector)
+    list(
+      class = element %>%
+        rvest::html_attr("class"),
+      style = element %>%
+        rvest::html_attr("style")
```
averissimo, line 15:
```suggestion
      class = rvest::html_attr(element, "class"),
      style = rvest::html_attr(element, "style")
```

R/TealAppDriver.R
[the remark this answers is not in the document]
vedhav: Setting the name space when not initially set 29cbca7b0f4ab904b872f8e0313758c40f08b7d5. The only reason why this failed was because the app was not initiated.

tests/testthat/test-shinytest2-teal_slices.R
vedhav: I thought it would be nice to catch the order change too.

averissimo: Sounds good... as I've said, minor

tests/testthat/test-shinytest2-filter_panel.R
pawelru, line 1: It would be good to introduce skip mechanism for these tests. Some time ago we used to have `TESTING_DEPTH` feature but I am happy with any other alternative. The goal is to keep test (overall r cmd check) timings at reasonable value for the CRAN submission (and execute everything in CI). I have seen somewhere a threshold of 10 mins but I cannot find a confirmation now. It might happen that we are pretty far from this right now but still I think it's good to introduce it now so this is available as we go forward with more testing. It would be much easier to do it now compared to mass modify every tests later on

m7pr:
Hey @vedhav hey @averissimo, do you already have a function that merges the name of the element with the name of the current active namespace?

```r
>   sprintf2 <- function(x, y = app$active_module_ns()){
+     sprintf("#%s-%s", y, x)
+   }
> sprintf2("button")
[1] "#teal-main_ui-root-module_with_verbatim_popup-module-button"
> app$active_module_ns()
[1] "teal-main_ui-root-module_with_verbatim_popup-module"
```

m7pr:
Is this `helper_ns` any of help?

R/TealAppDriver.R
pawelru, line 1: A thought of mine. Feel free not to apply it.

I'm currently seeing a 3 groups of methods: (i) expect, (ii) filter, (iii) ns (and presumably all other). You might want to organize them in a groups - either my naming convention or a list (`<instance>`$filter$foo()`). This might be better to organize the definition of a class (and potentially its future methods that we anticipate currently).

tests/testthat/test-shinytest2-modules.R
```diff
@@ -0,0 +1,117 @@
...
+      },
+      ui = function(id) {
+        ns <- NS(id)
+        h1("Module that exports a random value for testing")
+      }
+    )
+  }
+
+  app <- TealAppDriver$new(
+    data = simple_teal_data(),
```
pawelru, line 20: Consider this (with additional changes in helper file):
```suggestion
    data = simple_teal_data,
```

In `helper-xyz.R` we can create objects to be used across test files - not only functions.

This should be a little bit faster.

tests/testthat/test-shinytest2-filter_panel.R
```diff
@@ -0,0 +1,27 @@
...
+
+  app$wait_for_idle(timeout = default_idle_timeout)
+
+  old_output <- app$get_active_module_output("text")
+
+  app$set_active_filter_selection("iris", "Species", c("setosa", "versicolor"))
+
+  testthat::expect_false(
+    identical(old_output, app$get_active_module_output("text"))
+  )
```
gogonzo, line 24: This is close but not enough IMO. Is there any way to get into module's data?
```
testthat::expect_identical(
  app$get_active_module_data()[["iris"]],
  subset(iris, Species %in% c("setosa", "versicolor")),
)
```

tests/testthat/test-shinytest2-modules.R
averissimo: Using an object instead of a function could be a problem if the `simple_teal_data` object is modified. 

I know that none of these tests or future ones should change this, but it seems reasonable to have a function. 

Performance-wise it has a very low overhead compared to `shinytest2` runtimes

``` r
microbenchmark::microbenchmark(build = simple_teal_data(), unit = "ms")
#> Unit: milliseconds
#>   expr      min       lq      mean    median        uq    max neval
#>  build 0.506708 0.588549 0.8417622 0.6442315 0.7887155 14.569   100
```

pawelru: Yes - you are definitely right about modification. OK, this is definitely fine to leave it as function. Closing

tests/testthat/test-shinytest2-snapshot.R
```diff
@@ -0,0 +1,188 @@
+testthat::test_that("e2e: Create empty snapshot", {
```
gogonzo, line 1: Please remove this file and let's come back to testing snapshot manager when we merge [that PR](https://github.com/insightsengineering/teal/pull/1011). There will be some changes so the tests will have to be rewritten anyway

tests/testthat/test-shinytest2-reporter.R
```diff
@@ -0,0 +1,81 @@
...
+})
+
+testthat::test_that("e2e: adding a report card in a module adds it in the report previewer tab", {
+  app <- TealAppDriver$new(
+    data = simple_teal_data(),
+    modules = report_module(label = "Module with Reporter")
+  )
+  app$wait_for_idle(timeout = default_idle_timeout)
+
+  app$click(NS(app$active_module_ns(), "reporter-add_report_card_simple-add_report_card_button"))
```
pawelru, line 46: This introduces a coupling to the ids defined in `teal.reporter`. If we ever change it for some reason - `teal` would have to be re-released. The same goes with `teal.slice` and other packages. I would like to avoid this and keep releases as much independent as possible. 
This actually happened in the past - on `teal.slice` refactor we broke tests in `teal.modules.hermes` just because they were written like this.
Please invest in finding a good way to avoid such coupling.

averissimo: This is a difficult problem with E2E tests, in particular with `shinytest2` which has a workflow that is geared towards using IDs and CSS selectors.

We could mitigate some of this coupling by adding helpers that identify elements from labels/content, but there will always be some dependency on IDs / selectors / labels.

Maybe we could use a CSS selector pattern that looks for "ends with" `parent-context [id$=-add_report_card_button]` instead of relying on the full namespace.

m7pr: What if we test that only in `{teal.reporter}`, even though we have a class in `{teal}`?

averissimo: `{teal.reporter}` would need to have `{teal}` on Suggests or we would need to implement a non-standard way of testing (making use of staged dependencies environment).

It may be possible, but the question becomes: "should we?"

averissimo: The idea of a testing E2E package is growing on me. One that would contain:
- TealAppDriver
- E2E tests for `teal.{code,data,slices,reporter,widgets} packages.

All packages that already depend on `teal` could have the e2e tests living on them.

m7pr: yeah, but it's hard to decide what to do with packages on which `{teal}` relies. At least half of the problem is solved :P

R/TealAppDriver.R
```diff
@@ -0,0 +1,427 @@
...
+      function(..., .css_prefix = "", .css_suffix = "") {
+        dots <- rlang::list2(...)
+        checkmate::assert_list(dots, types = "character")
+        base_string <- sprintf("%s%s%s", .css_prefix, base_id, .css_suffix)
+        if (length(dots) == 0) {
+          return(base_string)
+        }
+        (shiny::NS(base_string))(paste(dots, collapse = shiny::ns.sep))
+      }
+    },
```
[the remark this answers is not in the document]
averissimo: This is going to be removed

tests/testthat/test-shinytest2-reporter.R
donyunardi: Thanks guys for the review.
I created a new issue #1160 so we can move forward with this PR.

DESCRIPTION
vedhav: Being tracked here #1162

R/TealAppDriver.R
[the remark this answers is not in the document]
vedhav: We can add think about adding this feature at the end of the e2e tests bonanza. Resolving this for now. Created a separate issue to track this.

R/TealAppDriver.R
averissimo: I took a jab at this on #1161 , it still holds on to `get_active_filter_selection` as a private method.

R/TealAppDriver.R
```diff
@@ -0,0 +1,410 @@
...
+    #' @param ... (`character`) The additional ids to be appended to `namespace`.
+    #'
+    #' @return A function similar to [shiny::NS()] that is used to create a `character`
+    #' namespace for the shiny UI.
+    #'
+    helper_NS = function(namespace, ...) { # nolint: object_name_linter.
+      dots <- rlang::list2(...)
+      checkmate::assert_list(dots, types = "character")
+      base_id <- namespace
+      if (length(dots) > 0) base_id <- paste(c(namespace, dots), collapse = shiny::ns.sep)
```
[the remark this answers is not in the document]
vedhav: PR #1147 removes the `helper_NS`. So, I'm resolving this.

tests/testthat/test-shinytest2-teal.R
vedhav: Thanks, this is done in 66dfc495e5c24df7ed39202944cf80faab80affd

R/TealAppDriver.R
vedhav: Thanks! I implemented your solution in 7bacc4aad614b52530202e5849f7828208846f29

tests/testthat/test-shinytest2-filter_panel.R
vedhav: The fiter panel fetch API can be discussed in the PR #1161 which is trying to make changes to it.

R/TealAppDriver.R
```diff
@@ -0,0 +1,443 @@
...
+        filter = filter,
+        title = title,
+        header = header,
+        footer = footer
+      )
+      suppressWarnings(
+        super$initialize(
+          shinyApp(app$ui, app$server),
+          name = "teal",
+          variant = platform_variant(),
```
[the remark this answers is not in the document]
vedhav: There are three things that help during the snapshot testing: `variant`, `seed`, and dimensions (`height` and `width`). So far we do not want to add them. But, the constructor can get these arguments as part of the `...` to the [AppDriver's constructor](https://rstudio.github.io/shinytest2/reference/AppDriver.html#method-AppDriver-new).

R/TealAppDriver.R
vedhav: Thanks for the suggestion, I like this way of organizing methods, it's certainly easier to use. I think we're not fully done with adding app-driving functionalities to manipulate the teal app yet. Once we map all the possible (useful) actions we will think about naming and keep this in mind.

.github/workflows/check.yaml
vedhav: reverted

tests/testthat/test-shinytest2-filter_panel.R
vedhav: Will apply this suggestion in #1161

R/TealAppDriver.R
vedhav: Tracking this on that PR. So, resolving.

averissimo (approved):
Nice work everyone!

Looks good!

m7pr:
Whoa. Great job

PR #1138, opened 2024-03-08
503 shinytest2 for `landing_popup_module`

m7pr:
Just a warm-up. Trying to understand `shinytest2` tests in action. Took `landing_popup_module` to the battlefield.


m7pr:
Hey @averissimo @gogonzo @vedhav 

Took a stab at `shinytest2` tests. Any chance you can review the proposed tests for `landing_popup_module` and review comments? I think the last test does not execute the `onclick/click`.

tests/testthat/test-shinytest2-landing_popup.R
```diff
@@ -0,0 +1,184 @@
...
+    data = simple_teal_data(),
+    modules = modules(
+      landing_popup_module(
+        title = "Welcome",
+        content = tags$b("A welcome message!", style = "color: red;")
+      ),
+      example_module()
+    )
+  )
+  app$view()
```
averissimo, line 12: This is only for debugging.

```suggestion
```

tests/testthat/test-shinytest2-landing_popup.R
```diff
@@ -0,0 +1,184 @@
...
+
+  app$wait_for_idle(timeout = default_idle_timeout)
+  app$expect_no_shiny_error()
```
averissimo, line 15: What do you think about looking for the text instead?

```suggestion
  testthat::expect_equal(
    app$get_text("#landingpopup b"),
    "A welcome message!"
  )
```

tests/testthat/test-shinytest2-landing_popup.R
```diff
@@ -0,0 +1,184 @@
...
+  )
+  app$wait_for_idle(timeout = default_idle_timeout)
+
+  expect_equal(
+    app$get_html(".btn-default") %>%
+      grep("Accept", x = ., value = TRUE) %>%
+      rvest::read_html() %>%
+      rvest::html_text(),
+    "Accept"
+  )
```
averissimo, line 35: Alternative suggestion (only buttons that have "Accept", exist inside popup):

```suggestion
  testthat::expect_setequal(
    app$get_text("#shiny-modal-wrapper .btn-default"),
    "Accept"
  )
```

Or make sure there exists only 1

```suggestion
  testthat::expect_equal(
    app$get_text("#shiny-modal-wrapper .btn-default"),
    "Accept"
  )
```

tests/testthat/test-shinytest2-landing_popup.R
```diff
@@ -0,0 +1,184 @@
...
+    modules = modules(
+      landing_popup_module(),
+      example_module()
+    )
+  )
+  app$wait_for_idle(timeout = default_idle_timeout)
+
+  app$click(selector = ".btn-default:nth-child(1)")
+
+  # QUESTION: Is there a better way of checking the name of current module?
```
averissimo, line 52: `app$get_active_ns("module")` gets the full namespace, not sure about the current name. @vedhav?

tests/testthat/test-shinytest2-landing_popup.R
```diff
@@ -0,0 +1,184 @@
...
+  expect_equal(
```
averissimo, line 53: Consider making sure the modal is gone from DOM. As the test passes even if button is not clicked.

```r
testthat::expect_null(app$get_html("#shiny-modal-wrapper"))
```

tests/testthat/test-shinytest2-landing_popup.R
```diff
@@ -0,0 +1,184 @@
...
+  app <- TealAppDriver$new(
+    data = simple_teal_data(),
+    modules = modules(
+      landing_popup_module(),
+      example_module()
+    )
+  )
+  app$wait_for_idle(timeout = default_idle_timeout)
+
+  app$click(selector = ".btn-default:nth-child(1)")
```
averissimo, line 50: How about clicking on the button that has a `data-dismiss` attribute?

```suggestion
  app$click(selector = "#shiny-modal-wrapper button[data-dismiss='modal']")
```

Or the first one, but inside the wrapper

```suggestion
  app$click(selector = "#shiny-modal-wrapper button:nth-child(1)")
```

tests/testthat/test-shinytest2-landing_popup.R
```diff
@@ -0,0 +1,184 @@
...
+extract_onclick <- function(id) {
+  app$get_html(id) %>%
+    rvest::read_html() %>%
+    rvest::html_nodes("button") %>%
+    rvest::html_attr("onclick")
+}
+extract_text <- function(id) {
+  app$get_html(id) %>%
+    rvest::read_html() %>%
+    rvest::html_text() %>%
```
averissimo, line 77: I believe this is equivalent, but the tests below are failing and can't confirm :sweat_smile: 

```suggestion
  app$get_text(id) %>%
```

tests/testthat/test-shinytest2-landing_popup.R
vedhav: We had a good suggestion in the parent PR where we are planning to rename it `active_ns` and use individual getters `active_module_ns`, `active_filters_ns`, `filter_manager_ns`

m7pr: Thanks!

tests/testthat/test-shinytest2-landing_popup.R
vedhav: Is there a real need to know which module is active?

The Teal app initializes with the first module as active. Then the navigation made in the test logic is the only way to change the module so it should be quite clear which is the active module because the test code should have this information.

tests/testthat/test-shinytest2-landing_popup.R
m7pr: Looks like ` app$get_text` is a gamechanger and a huge improvement in comparison to manual parsing of the text.

@averissimo your both suggestions are the same?

m7pr: Ah, there is `testthat::expect_setequal(` and `testthat::expect_equal(`

tests/testthat/test-shinytest2-landing_popup.R
m7pr: thanks. sure. saw it in some other tests at `test-shinytest2-teal.R` and probably carried over.

tests/testthat/test-shinytest2-landing_popup.R
vedhav: I always prefer selecting using data attributes over other ways, especially over getting the nth element. It will easily break once a new element is inserted/removed.

m7pr:
Thanks @averissimo for letting me know that `app$get_text` exists : D
And also for showing the better selector statements

tests/testthat/test-shinytest2-landing_popup.R
vedhav: Yeah, that was my bad. I was focused on helpers that I was committing the test file less cautiously.

tests/testthat/test-shinytest2-landing_popup.R
m7pr: Will double check!

tests/testthat/test-shinytest2-landing_popup.R
m7pr: Yeah, just wanted to double check if the modal is closed.

tests/testthat/test-shinytest2-landing_popup.R
averissimo: Yup, one would technically accept multiple buttons as long as all have "Accept"

tests/testthat/test-shinytest2-landing_popup.R
m7pr: one of the tests needed `trimws()` after applying `app$get_text()`

tests/testthat/test-shinytest2-landing_popup.R
m7pr: d7a99d55837db5d07eabdd34ca33e28ffaa00edf

m7pr:
Hey @averissimo maybe you have an idea why the buttons on

> e2e: when customized button in landing_popup_module is clicked, it redirects to a certain page

test do not redirect after clicking. Trying to test the `onclick` functionality but ended up in a dead end.

averissimo:
I tried a bunch of JS ideas and can't get the event to be triggered in `shinytest2`. This may be a limitation of `{shinytest2}`

- Running the `onclick` function from the property itself
- Running open directly (`app$run_js("window.open('http://google.com', '_blank')")`)
- ....

The best I can get is the function definition from the element itself

```r
app$get_js("document.getElementById(\"read\").onclick.toString()")
#> [1] "function onclick(event) {\nwindow.open('http://google.com', '_blank')\n}"
```

image

m7pr:
So @averissimo you say we can't click the button to achieve the `onlick` action? We can only extract the information that is included in the button and then compare if this contains what it should contain? This is also fine!

averissimo:
I don't think the problem is with the `onclick` per se, but maybe the `window` JS API is restricted. Consider the example below:

On click works for the first one, but not the second.

```r
library(shiny)

app <- shinyApp(
  fluidPage(
    div(
      id = "clicky",
      "click me",
      onclick = "Shiny.setInputValue('test', { 'key1' : 'val1', 'key2' : 'val2' }, {priority: 'event'})"
    ),
    div(
      id = "clicky2",
      "click me",
      onclick = "window.open('http://google.com', '_blank')"
    ),
    verbatimTextOutput("test")
  ),
  
  function(input, output, session) {
    output$test <- renderPrint({
      str(input$test)
    }) 
  }
)

driver <- shinytest2::AppDriver$new(app)

driver$view()
driver$click(selector = "#clicky")
driver$click(selector = "#clicky2")
```

m7pr:
Thanks. But what matters the most that you can get the JS code from under the button info

```r
app$get_js("document.getElementById(\"read\").onclick.toString()")
```

And this way we can actually see what was put during the app creation, and this is what I wanted to test.
So I will update the test and we can call it a day!

tests/testthat/test-shinytest2-landing_popup.R
```diff
@@ -0,0 +1,172 @@
...
+    modules = modules(
+      landing_popup_module(
+        buttons = actionButton("read", "Read more", onclick = onclick_text)
+      ),
+      example_module()
+    )
+  )
+  app$wait_for_idle(timeout = default_idle_timeout)
+
+  onclick_text_app <- app$get_js("document.getElementById(\"read\").onclick.toString()")
```
averissimo, line 165: Could use `extract_onclick("#read")` just like the previous test?

tests/testthat/test-shinytest2-landing_popup.R
```diff
@@ -0,0 +1,172 @@
...
+
+  testthat::expect_true(
+    grepl(onclick_text, onclick_text_app, fixed = TRUE)
+  )
```
averissimo, line 169:
```suggestion
  testthat::expect_match(onclick_text, onclick_text_app, fixed = TRUE)
```

tests/testthat/test-shinytest2-landing_popup.R
averissimo: And move that function to the helper or as method of `TealAppDriver`

averissimo (commented):
Small changes:

- I think we can avoid the `:nth-child(1)` as only 1 element will have `data-dismiss` attribute.
- Please unify how buttons are specified on all PRs with `button` or `.btn` or `.btn-default`
  - I would prefer one of the first 2, but I think any of them would be fine

---

It would be nice to detect onclick behavior via without resorting to string matching.
I've come up with this: (after failing to use a `r shiny::textInput` with `js Shiny::setInputValue` call to change it)

```r
onclick_text <- "document.getElementById('unique_input_id').value = 'Lorem ipsum';"

  app <- TealAppDriver$new(
    data = simple_teal_data(),
    modules = modules(
      landing_popup_module(
        buttons = actionButton("read", "Read more", onclick = onclick_text),
        content = tags$input(id = "unique_input_id")
      ),
      example_module()
    )
  )
  app$wait_for_idle(timeout = default_idle_timeout)

  onclick_text_app <- app$click(selector = "#read")

  # Looks for side-effects of button click (i.e. input value change).
  testthat::expect_equal(
    app$get_js("document.getElementById('unique_input_id').value"),
    "Lorem ipsum"
  )
```

tests/testthat/test-shinytest2-landing_popup.R
```diff
@@ -0,0 +1,172 @@
...
+    data = simple_teal_data(),
+    modules = modules(
+      landing_popup_module(),
+      example_module()
+    )
+  )
+  app$wait_for_idle(timeout = default_idle_timeout)
+
+  testthat::expect_equal(
+    app$get_text("#shiny-modal-wrapper .btn-default"),
```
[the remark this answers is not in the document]
m7pr: thanks, unified to `button`!

tests/testthat/test-shinytest2-landing_popup.R
```diff
@@ -0,0 +1,172 @@
...
+# customized landing_popup_module ---------------------------------------------------------------------------------
+
+
+extract_onclick <- function(app, id) {
+  app$get_html(id) %>%
+    rvest::read_html() %>%
+    rvest::html_nodes("button") %>%
+    rvest::html_attr("onclick")
+}
+phash <- function(text) paste0("#", text)
```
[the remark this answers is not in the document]
m7pr: totally, will try to rethink

tests/testthat/test-shinytest2-landing_popup.R
```diff
@@ -0,0 +1,172 @@
...
+      modal_title
+    )
+
+    testthat::expect_equal(
+      trimws(app$get_text(".modal-body")),
+      modal_content_message
+    )
+
+    testthat::expect_equal(
+      app$get_text(".btn-default:nth-child(1)"),
```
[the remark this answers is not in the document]
m7pr: Thanks, did that and unified to `button[data-dismiss]` ac18af56a93016b92be9159f1b0ea97f76ab17c6

tests/testthat/test-shinytest2-landing_popup.R
```diff
@@ -0,0 +1,172 @@
...
+  "e2e: app with customized landing_popup_module creates modal containing specified title, content and buttons",
+  {
+    modal_title <- "Custom Landing Popup Module Title"
+    modal_content_message <- "A welcome message!"
+    modal_content <- tags$b(modal_content_message, style = "color: red;")
+
+    modal_btns <- list(
+      list(text = "Proceed"),
+      list(text = "Read more", onclick = "window.open('http://google.com', '_blank')", id = "read"),
+      list(text = "Reject", onclick = "window.close()", id = "close")
```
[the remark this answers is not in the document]
m7pr: hell yeah e05ba8ac52a7a81f770fc1abb1d077f2878ae571

tests/testthat/test-shinytest2-landing_popup.R
m7pr: obviously! 0879bd0e510efbf5025686e5a343a3a1f09dd5b2
moved to a `TealAppDriver` method f3385b44355900bd65e30e6a8490423ef7a7362c

tests/testthat/test-shinytest2-landing_popup.R
m7pr: with `app$get_onclick("#read")` no `expect_match` is needed

tests/testthat/test-shinytest2-landing_popup.R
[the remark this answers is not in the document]
m7pr: moved `phash` directly to a test 88a3bc53e8582875c67be4a54caf1fb90749acc6

m7pr:
Per your last commit @averissimo , I think we now have a very concise way of doing this test with `app$get_onclick("#read")`

m7pr:
Ah, I see you are referring to `value` which gets changed by JS when a button is clicked. Hmmm

averissimo:
Yes, I'm not sure about it though as this is a Shiny functionality, not a landing module one.

Maybe we should only test that the landing popup appears. Everything else is Shiny-related

m7pr:
got it @averissimo  and totally agree. I would stop the develpoment and keep the state as it's currently is.

R/TealAppDriver.R
```diff
@@ -356,6 +356,30 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
     #'
+    #' @return The `TealAppDriver` object invisibly.
+    open_filter_manager = function() {
+      active_ns <- self$filter_manager_ns()
+      ns <- self$helper_NS(active_ns)
+
+      self$click(ns("show"))
+      self$wait_for_idle(500)
+      invisible(self)
+    },
```
m7pr, line 367: @vedhav just double checking this shouldn't be here right : P ?

R/TealAppDriver.R
```diff
@@ -356,6 +356,30 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
+    #' @description
+    #' Extract `onlick` attribute from a `button` (found by `selector`).
+    #'
+    #' @param selector (`character(1)`) specifying the selector to be used to get the content of a `button`.
+    #'
+    #' @return The `character` vector.
+    get_onclick = function(selector) {
+      self$get_html(selector) %>%
+        rvest::read_html() %>%
```
m7pr, line 376:
```suggestion
      self$get_html_rvest(selector) %>%
```

R/TealAppDriver.R
m7pr: removed

m7pr:
ugh, there is test failing for reporter part. but this PR does not involve testing reporter

m7pr:
it was failing on the `landing_popup` test. some commits were missing. fixed the issue

R/TealAppDriver.R
```diff
@@ -382,6 +382,17 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
+    #' Extract `onlick` attribute from a `button` (found by `selector`).
+    #'
+    #' @param selector (`character(1)`) specifying the selector to be used to get the content of a `button`.
+    #'
+    #' @return The `character` vector.
+    get_onclick = function(selector) {
+      self$get_html_rvest(selector) %>%
+        rvest::html_nodes("button") %>%
+        rvest::html_attr("onclick")
+    },
```
averissimo, line 394: This is a very specific method for a single attribute, what do you think about converting it to a generic attribute getter? (that works for all HTML nodes besides `button`)

then replace all `app$get_onclick(selector)` with `app$get_attr(selector, "onclick"))`

```suggestion
    #' @description
    #' Extract `html` attribute from (found by a `selector`).
    #'
    #' @param selector (`character(1)`) specifying the selector to be used to get the content of a `button`.
    #'
    #' @return The `character` vector.
    get_attr = function(selector, attribute_name) {
      rvest::read_html(self$get_html("html")) %>%
        rvest::html_nodes(selector) %>%
        rvest::html_attr(attribute_name)
    }
```

averissimo (commented):
Thanks for the changes, I have one more comment, let's discuss it :smile: 

It boils down to the specificity of `get_onclick`.

It would be nice to:
- have access to other attributes in the future tests
- getting `onclick` from `<a>` elements or another element
    - _(`<div>` on screenshot below also has `onclick` attr)_

image

R/TealAppDriver.R
averissimo: Note that `rvest::read_html` after piping from `app$get_html()` will add a `<body>` wrapper around the element.

Hence it might be better to use `rvest::read_html(self$get_html("html"))`

Ideally, we'd use `xml2::read_xml` after `app$get_html(selector)`, but that would add a new dependency :-\

averissimo:
Other occurrences of getting attributes in the code could use this new method.

I can create a follow-up PR to this one and change those if you accept the changes :relaxed:

vedhav:
Please add `skip_if_too_deep(5)` too when you get a chance.

R/TealAppDriver.R
m7pr: `rvest::read_html(self$get_html("html"))` is implemented as `self$get_html_rvest("html")`

m7pr: I like `get_attr` to be customized for a wider approach! Will include, and rewrite in few places where applicable.

m7pr:
@averissimo thanks for the proposition of `$get_attr`. I think this will be super useful

R/TealAppDriver.R
averissimo: :facepalm: I've updated the suggestion to use `self$get_html_rvest("html")`

> I think we will replace `rvest` with `xml2` #1162 so no new dependencies. `rvest` under the hood is just a `xml2` all over the place : )

Cool!

m7pr:
cleaned up some documentation and just pushed it for final review

averissimo (approved):
Looks good! Nice work :100:

PR #1139, opened 2024-03-08
Removes redundant `app` from R6 class

averissimo:
# Pull Request

Part of [#503](https://github.com/insightsengineering/coredev-tasks/issues/503)

There's no need to store "app" from `super$initialize`

image


vedhav:
Thanks @averissimo!

PR #1152, opened 2024-03-13
503 shinytest2 fixes bug with setting numeric range

### LINKED ISSUE #1151: [Bug]: `TealAppDriver` setting numeric filter state
`TealAppDriver$set_active_filter_selection` only works for categorical variables. It should also work for numeric ranges.

Example below of test that should work

```R
testthat::test_that("e2e: Set numeric filter state", {
  app <- TealAppDriver$new(
    data = simple_teal_data(),
    modules = example_module(label = "Example Module"),
    timeout = default_idle_timeout
  )
  withr::defer(app$stop())

  app$wait_for_idle()

  app$add_filter_var("iris", "Sepal.Length")
  app$view()
  app$set_active_filter_selection("iris", "Sepal.Length", c(4, 5), is_numeric = TRUE)
  app$wait_for_idle()
  testthat::expect_equal(
    app$get_active_filter_selection("iris", "Sepal.Length", is_numeric = TRUE), 
    c(4, 5)
  )
})
```

averissimo:
# Pull Request

Fixes #1151 

#### Changes description

- `shinyWidgets::numericRangeInput` uses a custom handler and seems to require a `js: Shiny.setInputValue` call
- ~Change of explicit arguments in `{s,g}et_active_filter_selection` to `type` to reflect this logic and allow for further extensions.~
- Removed `is_numeric` argument in favor of auto-detection of slice type (categorical / numerical range)
  - from `{s,g}et_active_filter_selection`


tests/testthat/test-shinytest2-teal_slices.R
```diff
@@ -99,3 +99,25 @@ testthat::test_that("e2e: teal_slices filters are initialized when module specif
   )
   app$stop()
 })
+
+testthat::test_that("e2e: Set numeric filter state", {
```
vedhav, line 103: I think this test is mainly testing the `TealAppDriver$set_active_filter_selection` method. In this case I would think that it's better placed inside a new file dedicated to test the public methods from the class.
We can still test some business logic that uses `set_active_filter_selection`. For example, setting the filter state changes the filter state in another module (when module_specific is FALSE) and shared module (when module_specific is TRUE)

Also, the test `e2e: teal_slices filters are initialized when module specific filters are created` can remove the last part where we use `set_active_filter_selection`. It can stop after expecting the initialized values as suggested by the test desc.

averissimo: The idea was to test the filter panel using the `shinytest2` way. But you're right, it only tests the function.

I'd say remove the test completely as well as that last part of the other test

averissimo: Re-include once we start to test `TealAppDriver`

averissimo: Done with 3346cc7

vedhav (approved):
Thanks! LGTM!

PR #1156, opened 2024-03-15
503 `$active_module_element` method for `TealAppDriver`

m7pr:
I found it handy to have a function like this in here https://github.com/insightsengineering/teal/pull/1127#issuecomment-1994283810

Proposing a method to exist at `TealAppDriver` directly.

### DIFF
--- a/R/TealAppDriver.R
+++ b/R/TealAppDriver.R
@@ -137,6 +137,16 @@ TealAppDriver <- R6::R6Class( # nolint
       private$ns$module
     },
     #' @description
+    #' Get the active shiny name space bound with a custom `element` name.
+    #'
+    #' @param element `character(1)` custom element name.
+    #'
+    #' @return (`string`) The active shiny name space of the component bound with the input `element`.
+    active_module_element = function(element) {
+      checkmate::assert_string(element)
+      sprintf("#%s-%s", self$active_module_ns(), element)
+    },
+    #' @description
     #' Get the active shiny name space for interacting with the filter panel.
     #'
     #' @return (`string`) The active shiny name space of the component.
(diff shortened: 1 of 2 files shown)


vedhav (approved):
👍🏽 Agreed, It would help when referencing modules inside the teal_module.

m7pr:
@vedhav what about one more method

```r
active_module_element_text <- function(selector) {
    self$active_module_element(selector) %>% get_text()
}
```

So instead of
```r
app$active_module_element("copy_button1") %>%
    app$get_text()
```    

You just type
```r
app$active_module_element_text("copy_button1")
```

m7pr:
CC @vedhav

vedhav:
Thanks for the ping. [After Paweł bought it up](https://github.com/insightsengineering/teal/pull/1127/files#r1523206384). It makes me wonder if we have too many of these methods which makes it hard for people to know what they do. Let's rethink about trying to simplify the usage of these methods once the dust settles and we narrow down all the methods we need.

m7pr:
I don't think Pawel mentioned we have too many methods. What he mentioned was we should just organize them so it's easy to use and seek trhough in an alphabetically ordered list. If we think a method is useful and makes the code shorter I think we should go for whathever the amount of methods that makes our life easier

m7pr:
We can have `$active_module_element_name` or `$active_module_element_id` and `$active_module_element_text`. So one method gets the ID of the element, and the other one get's the text : )

vedhav:
Yeah, let’s extend it to your suggestions. They’re definitely improvement.
In the future we can think about some nomenclature like `app$active$element_name`

PR #1157, opened 2024-03-15
503 add `{checkmate}` assertions to `TealAppDriver`

m7pr:
We often just use our methods to pass parameters directly to other methods of `TealAppDriver` or `AppDriver` directly. Those already can have `{checkmate}`-like assertions on parameters' classes. However we tend to do operations on our parameters (mostly pasting) before we use them further. Now a big teal, but I would suggest checking if the input has a proper value before we paste it and pass forward. The length and the class.


averissimo (approved):
Awesome!

PR #1161, opened 2024-03-15
503 Applying suggestions

averissimo:
# Pull Request

### Changes description:

- [x] Implement `TESTING_DEPTH
- [x] Refactor `get_active_data_filters` to return active selection value
- [x] Remove pipe operator on `TealAppDriver`


vedhav:
Let's also introduce `TESTING_DEPTH` functionality in the testthat setup to help avoid these tests on CRAN and when wanting to run only unit tests [as suggested here](https://github.com/insightsengineering/teal/pull/1127/files#r1523053484).
[The teal.modules.helios has an example of how it was done before.](https://github.com/insightsengineering/teal.modules.helios/blob/main/tests/testthat/setup-skip_if_too_deep.R)

averissimo:
I think we should also remove the sole pipe operator on `TealAppDriver`

vedhav:
>  I think we should also remove the sole pipe operator on TealAppDriver

Thank you. I always forget that :(

vedhav:
To run the e2e tests make sure to set the option for `TESTING_DEPTH` above or equal to 5. Alternatively, a env variable with the same name can also be set. But, the precedence would be given to the option in case both are set.

```r
options(TESTING_DEPTH = 5)
```

tests/testthat/setup-testing_depth.R
```diff
@@ -0,0 +1,56 @@
...
+#'
+#' @details Looks for the option `TESTING_DEPTH` first, if not set,
+#' takes the system environmental variable `TESTING_DEPTH`. If neither
+#' is set, then returns 3 by default. If the value of `TESTING_DEPTH`
+#' is not a scalar numeric, then returns 3.
+#'
+#' @return `numeric(1)` the testing depth.
+#'
+testing_depth <- function() {
+  testing_depth <- getOption("TESTING_DEPTH")
```
m7pr, line 11: what about exposing the depth to be a parameter? so that you can set it during the function call OR with a session option.
```suggestion
testing_depth <- function(testing_depth = getOption("TESTING_DEPTH")) {

```

tests/testthat/setup-testing_depth.R
```diff
@@ -0,0 +1,56 @@
...
+  if (is.null(testing_depth)) {
```
m7pr, line 12:
```suggestion
  if (is.null(testing_depth) & Sys.getenv("TESTING_DEPTH") != "") {
```

tests/testthat/setup-testing_depth.R
vedhav: I think having no parameter sends a stronger signal that the only way to influence the TEST DEPTH is to set it using options or env variable. Else we give the reader an additional task to figure out what is happening in the function call. So, I prefer not having a parameter to this function. Essentially it is like a read function to get the depth.

tests/testthat/setup-testing_depth.R
```diff
@@ -0,0 +1,56 @@
...
+
+  testing_depth <- tryCatch(
+    as.numeric(testing_depth),
+    error = function(error) 3,
+    warning = function(warning) 3
+  )
+
+  if (length(testing_depth) != 1 || is.na(testing_depth)) {
+    testing_depth <- 3
+  }
```
m7pr, line 24: If you do not allow `testing_depth` to be `""` (when `Sys.getenv()` returns `""`) https://github.com/insightsengineering/teal/pull/1161/files#r1531798495 then `as.numeric("")` doesn't make it `NA`
```suggestion
  testing_depth <- testing_depth[1]
```

tests/testthat/setup-testing_depth.R
```diff
@@ -0,0 +1,56 @@
+#' Returns testing depth set by an environmental variable.
+#'
+#' @details Looks for the option `TESTING_DEPTH` first, if not set,
+#' takes the system environmental variable `TESTING_DEPTH`. If neither
+#' is set, then returns 3 by default. If the value of `TESTING_DEPTH`
+#' is not a scalar numeric, then returns 3.
+#'
+#' @return `numeric(1)` the testing depth.
+#'
+testing_depth <- function() {
```
m7pr, line 10: Can we get the function name to be called `testing_depth` but the variable inside be called `depth`?

tests/testthat/setup-testing_depth.R
m7pr: Or is it just for getting the depth, and not setting? Then maybe rename to `get_testing_depth()`

tests/testthat/setup-testing_depth.R
```diff
@@ -0,0 +1,56 @@
...
+#'   has opposite interpretation to environment variable `TESTING_DEPTH`.
+#'   So e.g. `0` means run it always and `5` means a heavy test which should be run rarely.
+#'   If the `depth` argument is larger than `TESTING_DEPTH` then the test is skipped.
+#' @importFrom testthat skip
+#' @return `NULL` or invoke an error produced by `testthat::skip`
+#' @note By default `TESTING_DEPTH` is equal to 3 if there is no environment variable for it.
+#' By default `depth` argument lower or equal to 3 will not be skipped because by default `TESTING_DEPTH`
+#' is equal to 3. To skip <= 3 depth tests then the environment variable has to be lower than 3 respectively.
+skip_if_too_deep <- function(depth) { # nolintr
+  stopifnot(length(depth) == 1 && is.numeric(depth) && depth >= 0 && depth <= 5)
```
m7pr, line 51:
```suggestion
  checkmate::assert_numeric(depth, len = 1, lower = 0, upper = 5)
```

tests/testthat/setup-testing_depth.R
m7pr: Yeah, let's have no parameter but let's change the name to `get_` as I spent some time figuring out if the purpose is to set (based on an ENVIRONEMNT VARIABLE / option) or to get the depth.

tests/testthat/setup-testing_depth.R
m7pr: or we make upper to be INF?

tests/testthat/setup-testing_depth.R
vedhav: Sure! Nice consistancy.

vedhav: Done in 75fa708792334ef552b35417c3fdb1fe14587536

tests/testthat/setup-testing_depth.R
vedhav: Thanks! changed the name.

tests/testthat/setup-testing_depth.R
vedhav: I think the current logic is straightforward. We set depth from options, then set options from env.

Perhaps setting the `default` and `unset` to `3` for the `getOption` and `Sys.getenv` respectively?

tests/testthat/setup-testing_depth.R
```diff
@@ -0,0 +1,56 @@
...
+#' @importFrom testthat skip
+#' @return `NULL` or invoke an error produced by `testthat::skip`
+#' @note By default `TESTING_DEPTH` is equal to 3 if there is no environment variable for it.
+#' By default `depth` argument lower or equal to 3 will not be skipped because by default `TESTING_DEPTH`
+#' is equal to 3. To skip <= 3 depth tests then the environment variable has to be lower than 3 respectively.
+skip_if_too_deep <- function(depth) { # nolintr
+  checkmate::assert_numeric(depth, len = 1, lower = 0, upper = 5)
+  test_to_depth <- get_testing_depth() # by default 3 if there are no env variable
+  if (test_to_depth < depth) {
+    testthat::skip(paste("testing depth", test_to_depth, "is below current testing specification", depth))
```
m7pr, line 54:
```suggestion
  testing_depth <- get_testing_depth() # by default 3 if there are no env variable
  if (testing_depth < depth) {
    testthat::skip(paste("testing depth", testing_depth, "is below current testing specification", depth))
```

tests/testthat/setup-testing_depth.R
vedhav: And keep this 3 as a default variable in the function as it's used in many places/

tests/testthat/setup-testing_depth.R
vedhav: Done!

tests/testthat/setup-testing_depth.R
m7pr: What is not straightforward is that empty `getOption` return NULL and empty `Sys.getenv` returns `""` and then you need to do `tryCatch` on `as.numeric` and figure out if you got `NA`

tests/testthat/setup-testing_depth.R
vedhav: an unset value solves this.

tests/testthat/setup-testing_depth.R
vedhav: Now the empty `getOption` will return 3 and empty `Sys.getenv` will return "3"
84568ec81c453e3cb3a37b161661e3a3135f92ab

tests/testthat/setup-testing_depth.R
vedhav: Reference commit 84568ec81c453e3cb3a37b161661e3a3135f92ab

tests/testthat/setup-testing_depth.R
```diff
@@ -0,0 +1,57 @@
...
+#' is not a scalar numeric, then returns 3.
+#'
+#' @return `numeric(1)` the testing depth.
+#'
+get_testing_depth <- function() {
+  default_depth <- 3
+  depth <- getOption("TESTING_DEPTH", default_depth)
+  if (is.null(depth)) {
+    depth <- Sys.getenv("TESTING_DEPTH", default_depth)
+  }
```
m7pr, line 15: this way you will never have `is.null(depth)` hence you will never use `Sys.getenv`

```r
getOption("XSSASASACASX", 2)
#> 2
```

m7pr: What about 

```r
depth <- getOption("TESTING_DEPTH", Sys.getenv("TESTING_DEPTH"))
```

vedhav: We can, if the user sets the depth as "bazinga!"

vedhav: Yes, this is perfect.

```r
default_depth <- 3
depth <- getOption("TESTING_DEPTH", Sys.getenv("TESTING_DEPTH", default_depth))
```

m7pr: nice!

Just remember `Sys.getenv` returns characters

```r
> Sys.getenv("TESTING_DEPTH", default_depth)
[1] "3"
> default_depth
[1] 3
```

m7pr: but later `as.numeric` should be fine

vedhav: Yes sir

tests/testthat/setup-testing_depth.R
```diff
@@ -0,0 +1,55 @@
...
+#' is set, then returns 3 by default. If the value of `TESTING_DEPTH`
+#' is not a scalar numeric, then returns 3.
+#'
+#' @return `numeric(1)` the testing depth.
+#'
+get_testing_depth <- function() {
+  default_depth <- 3
+  depth <- getOption("TESTING_DEPTH", Sys.getenv("TESTING_DEPTH", default_depth))
+
+
```
m7pr, line 14:
```suggestion

```

tests/testthat/setup-testing_depth.R
```diff
@@ -0,0 +1,55 @@
...
+  depth <- tryCatch(
+    as.numeric(depth),
+    error = function(error) default_depth,
+    warning = function(warning) default_depth
+  )
+
+  if (length(depth) != 1 || is.na(depth)) {
```
m7pr, line 21: Now I think there is no way this is `NA`
```suggestion
  if (length(depth) != 1) {
```

tests/testthat/setup-testing_depth.R
vedhav: I had so much faith in the linter.

tests/testthat/setup-testing_depth.R
```diff
@@ -0,0 +1,55 @@
...
+    depth <- default_depth
+  }
+
```
m7pr, line 22:
```suggestion
```

tests/testthat/setup-testing_depth.R
vedhav: True, it will never be `NA` anymore.

tests/testthat/setup-testing_depth.R
m7pr: I could be faster then the lintr GHA check : P

tests/testthat/setup-testing_depth.R
```diff
@@ -0,0 +1,52 @@
+#' Returns testing depth set by an environmental variable.
```
m7pr, line 1:
```suggestion
#' Returns testing depth set by session option or by environmental variable.
```

tests/testthat/setup-testing_depth.R
vedhav: lol

tests/testthat/setup-testing_depth.R
```diff
@@ -0,0 +1,52 @@
...
+#'
+#' @details Looks for the option `TESTING_DEPTH` first, if not set,
```
m7pr, line 3:
```suggestion
#' @details Looks for the session option `TESTING_DEPTH` first. If not set,
```

tests/testthat/setup-testing_depth.R
```diff
@@ -0,0 +1,52 @@
...
+#' takes the system environmental variable `TESTING_DEPTH`. If neither
+#' is set, then returns 3 by default. If the value of `TESTING_DEPTH`
+#' is not a scalar numeric, then returns 3.
```
m7pr, line 6:
```suggestion
#' is not a numeric of length 1, then returns 3.
```

tests/testthat/setup-testing_depth.R
```diff
@@ -0,0 +1,52 @@
...
+  depth <- getOption("TESTING_DEPTH", Sys.getenv("TESTING_DEPTH", default_depth))
+  depth <- tryCatch(
+    as.numeric(depth),
+    error = function(error) default_depth,
+    warning = function(warning) default_depth
+  )
+
+  if (length(depth) != 1) {
+    depth <- default_depth
+  }
```
m7pr, line 21:
```suggestion
  if (length(depth) != 1) depth <- default_depth
```

tests/testthat/setup-testing_depth.R
```diff
@@ -0,0 +1,52 @@
...
+#'
+#' @return `numeric(1)` the testing depth.
+#'
+get_testing_depth <- function() {
+  default_depth <- 3
+  depth <- getOption("TESTING_DEPTH", Sys.getenv("TESTING_DEPTH", default_depth))
+  depth <- tryCatch(
+    as.numeric(depth),
+    error = function(error) default_depth,
+    warning = function(warning) default_depth
```
m7pr, line 16: @vedhav when this can be a `warning`? Is this needed?

tests/testthat/setup-testing_depth.R
chlebowa: styler will revert this anyway

tests/testthat/setup-testing_depth.R
vedhav:
```r
as.numeric('User setting testing depth as a character that cannot coerce into number')
```

tests/testthat/setup-testing_depth.R
vedhav: I always prefer `{}`. Even when it's one liner. But, I'll accept it. I've seen our repos have this.

tests/testthat/setup-testing_depth.R
m7pr: coolio, thanks

tests/testthat/setup-testing_depth.R
```diff
@@ -0,0 +1,49 @@
...
+  )
+  if (length(depth) != 1) depth <- default_depth
```
m7pr, line 18: @chlebowa I think styler
https://github.com/insightsengineering/teal/blob/main/.github/workflows/check.yaml#L86C11-L86C45
runs simple `styler::style_file` without any `transformations` https://github.com/insightsengineering/r.pkg.template/blob/main/.github/workflows/style.yaml#L129
hence this will not be changed
image

tests/testthat/setup-testing_depth.R
m7pr: https://github.com/insightsengineering/teal/pull/1161#discussion_r1531884490

m7pr: thanks @vedhav

tests/testthat/setup-testing_depth.R
```diff
@@ -0,0 +1,49 @@
+#' Returns testing depth set by session option or by environmental variable.
+#'
+#' @details Looks for the session option `TESTING_DEPTH` first. If not set,
+#' takes the system environmental variable `TESTING_DEPTH`. If neither
+#' is set, then returns 3 by default. If the value of `TESTING_DEPTH`
+#' is not a numeric of length 1, then returns 3.
```
m7pr, line 6:
```suggestion
#' @details Looks for the session option `TESTING_DEPTH` first. 
#' If not set, takes the system environmental variable `TESTING_DEPTH`. 
#' If neither is set, then returns 3 by default. 
#' If the value of `TESTING_DEPTH` is not a numeric of length 1, then returns 3.
```

m7pr (approved):
@vedhav looks great. thanks for incorporating that! 
Left one last suggestion, feel free to ignore.

averissimo:
Great job with `TESTING_DEPTH`. Is it worth to propagate these changes to other repos that use it?

vedhav:
Eventually, we will have to do it once we setup the shinytest2 for them.

vedhav:

PR #1167, opened 2024-03-19
`$active_module_element_text` method `TealAppDriver`

m7pr:
Followu-up after https://github.com/insightsengineering/teal/pull/1156


vedhav (approved):
Thanks @m7pr LGTM

PR #1171, opened 2024-03-21
1163 extend `TealAppDriver$new()` and `TealAppDriver$click()` with `$wait_for_idle()` calls

### LINKED ISSUE #1163: [Feature] Adding `wait_for_idle` in the constructor of `TealAppDriver`
How about adding the `wait_for_idle` here instead of using it after `TealAppDriver$new` on each single test?

I also suggest re-implementing `wait_for_idle` to increase the timeout of the `TealAppDriver`, wdyt?

```suggestion
      private$set_active_ns()
      self$wait_for_idle()
    },
    wait_for_idle = function(duration = 500, timeout = 20000) {
      super$wait_for_idle(duration = duration, timeout = timeout)
    },
```

_Originally posted by @averissimo in https://github.com/insightsengineering/teal/pull/1127#discussion_r1521365360_

m7pr:
Close #1163 

- removed all `$wait_for_idle()` calls after `TealAppDrvier$new()` as I extended the method to call `$wait_for_idle()`
- removed all `$wait_for_idle()` calls after `app$click()` as I extended the method to call `$wait_for_idle()`
- removed all `$wait_for_idle()` calls after `app$navigate_teal_tab()` as it is called inside the method already

Questions: 
- Should we add `timeout` parameter to initialize that will be passed to `self$wait_for_idle()` or add `...` in `self$wait_for_idle(...)` at the end of the call. So that during `TealAppDriver$new()` you can pass `timeout = `.
- `app$navigate_teal_tab` uses `private$idle_timeout`  (`self$wait_for_idle(timeout = private$idle_timeout)`). Should we use `private$idle_timeout` in `TealAppDrvier$new` where we call `$wait_for_idle()`. Also, should we set `private$idle_timeout` in `TealAppDrvier$new`


averissimo:
> * Should we add `timeout` parameter to initialize that will be passed to `self$wait_for_idle()` or add `...` in `self$wait_for_idle(...)` at the end of the call. So that during `TealAppDriver$new()` you can pass `timeout = `.
> 
> * `app$navigate_teal_tab` uses `private$idle_timeout`  (`self$wait_for_idle(timeout = private$idle_timeout)`). Should we use `private$idle_timeout` in `TealAppDrvier$new` where we call `$wait_for_idle()`. Also, should we set `private$idle_timeout` in `TealAppDrvier$new`

I think we should change the default value of the private element `private$timeout` to 20s instead of this non-default field and reimplementing `wait_for_idle`.

@vedhav was there a specific rationale to create `private$idle_timeout` instead of changing the default value of `private$timeout`?

averissimo:
Ps. I know it was me that suggested the reimplemented `wait_for_idle`, but changing the default private field might be enough to achieve the same purpose 😅

m7pr:
@averissimo no worries, but the part of adding `$wait_for_idle()` calls to the `$new()` and `$click()` methods is still valid : )

averissimo:
> @averissimo no worries, but the part of adding `$wait_for_idle()` calls to the `$new()` and `$click()` methods is still valid : )

For sure! This will help us avoid a bunch of statements on tests :+1:

R/TealAppDriver.R
```diff
@@ -47,6 +47,20 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
+    #' @param duration,timeout (`numeric(1)`) passed to parent [`AppDriver`] `waif_for_idle()` method.
+    wait_for_idle = function(duration = 500, timeout = 20000) {
+      super$wait_for_idle(duration = duration, timeout = timeout)
+    },
+    #' @description
+    #' Append parent [`AppDriver`] `click` method with a call to `waif_for_idle()` method.
+    #' @param ... arguments passed to parent [`AppDriver`] `click()` method.
+    click = function(...) {
+      super$click(...)
+      wait_for_idle()
```
averissimo, line 63:
```suggestion
      self$wait_for_idle()
```

R/TealAppDriver.R
```diff
@@ -37,16 +37,36 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
+              app_dir = shinyApp(app$ui, app$server),
+              name = "teal",
+              variant = platform_variant()
+            )
+          )
         )
       )
 
       private$set_active_ns()
+      self$wait_for_idle()
```
vedhav, line 62: 👍🏽 for waiting by default during initialization.

R/TealAppDriver.R
```diff
@@ -37,16 +37,36 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
         header = header,
         footer = footer
       )
+
+      dots <- rlang::list2(...)
+      # Default timeout is hardcoded to 4s in shinytest2:::resolve_timeout
+      # It must be set as parameter to the AppDriver
+      if (is.null(dots$timeout)) dots$timeout <- 20 * 1000
+      if (is.null(dots$load_timeout)) dots$load_timeout <- 100 * 1000
```
vedhav, line 45: Nice implementation to overwrite the timeouts. I think it's worth mentioning the timeouts in the function doc.

vedhav (approved):
Thanks both of you for the changes! It's a nice improvement for the class. Just felt it will be nice to add this to the doc. Apart from this LGTM.

R/TealAppDriver.R
averissimo: image

Does this look good?

vedhav: Thanks for the doc changes. LGTM now :)

m7pr:
Thanks guys for taking care of this!

PR #1172, opened 2024-03-21
add wrapper for wait_for_value for output value.

kartikeyakirar:
This PR introduces a wrapper for the wait_for_value method, specifically designed for output values.


m7pr:
Lol, I just did the same in here https://github.com/insightsengineering/teal/pull/1171 but also edited tests and extended one more method :P @kartikeyakirar I think you can close this one

m7pr:
ignore my previous comment!

R/TealAppDriver.R
```diff
@@ -407,6 +407,14 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
     open_url = function() {
       browseURL(self$get_url())
+    },
+    #' @description
+    #' Wait until an output value is not one of `ignored` values `timeout` is reached.
+    #' This function will only access output value and its wrapper around `wait_for_value` method.
+    #' @param output_id (character) The shiny output id to get the value from.
+    wait_for_ouput_value = function(output_id, ...) {
+      checkmate::check_string(input_id)
+      self$wait_for_value(output = sprintf("%s-%s", self$active_module_ns(), output_id), ...)
```
averissimo, line 417: - What you do you think about renaming the method to explicitly say that this is for the active module output value?
- Should we check if `...` contains `export`, `output` or `input` parameter and throw an error if it does?

R/TealAppDriver.R
```diff
@@ -407,6 +407,14 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
     #' @return Nothing. Opens the underlying teal app in the browser.
     open_url = function() {
       browseURL(self$get_url())
+    },
+    #' @description
+    #' Wait until an output value is not one of `ignored` values `timeout` is reached.
+    #' This function will only access output value and its wrapper around `wait_for_value` method.
+    #' @param output_id (character) The shiny output id to get the value from.
```
averissimo, line 414: `...` is not documented

R/TealAppDriver.R
kartikeyakirar: OR lets create a wrapper for `wait_for_value` function that checks if the function contains export, output or input parameter and assigns it accordingly.

WDYT?

averissimo: So it would have `id` and then prepend to the correct? Sounds good.

`wait_for_active_module_value` as name?

R/TealAppDriver.R
```diff
@@ -407,6 +407,30 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
+    #' Waits until a specified input, output, or export value.
+    #' This function serves as a wrapper around the `wait_for_value` method,
+    #' providing a more flexible interface for waiting on different types of values within the active module namespace.
+    #' @param ... Dynamic parameters allowing the specification of either an `input`, `output`, or `export`
+    #' parameter with additional value to passed in `wait_for_value`.
+    wait_for_active_module_value = function(...) {
+      dots <- list(...)
+      param_types <- intersect(names(dots), c("input", "output", "export"))
+      checkmate::assert(
+        checkmate::assert_choice(parm_type, c("input", "output", "export")),
```
averissimo, line 421:
```suggestion
        checkmate::assert_choice(param_type, c("input", "output", "export")),
```

R/TealAppDriver.R
```diff
@@ -407,6 +407,30 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
+        combine = "and"
+      )
+
+
+      modified_id <- sprintf("%s-%s", self$active_module_ns(), dots[[param_type]])
+      dots <- dots[!names(dots) %in% param_type]
+
+      do.call(
+        what = self$wait_for_value,
+        args = c(list(param_type = modified_id, timeout = private$load_timeout), dots)
```
averissimo, line 432: Load timeout is not used in this context

> How long to wait for the app to load, in ms. This includes the time to start R. Defaults to 15s

Our own `private$idle_timeout` may be used. Or something different once I get a PR up.

```suggestion
        args = c(list(param_type = modified_id, timeout = private$idle_timeout), dots)
```

averissimo (changes requested):
I think this still requires some improvements.

- typo
- better `*_timeout` to be used
- character vector is hardcoded twice

Consider reducing the logic on this wrapper and just passing modified `input/output/export` and leave the deep checks to the parent method

```r
    wait_for_active_module_value = function(input = rlang::missing_arg(),
                                            output = rlang::missing_arg(),
                                            export = rlang::missing_arg(),
                                            ...) {
      ns <- shiny::NS(self$active_module_ns())
      
      if (!rlang::is_missing(input) && checkmate::test_string(input, min.chars = 1)) input <- ns(input)
      if (!rlang::is_missing(output) && checkmate::test_string(output, min.chars = 1)) output <- ns(output)
      if (!rlang::is_missing(export) && checkmate::test_string(export, min.chars = 1)) export <- ns(export)

      
      self$wait_for_value(input = input, output = output, export = export, ...)
    }
```

kartikeyakirar:
> Consider reducing the logic on this wrapper and just passing modified input/output/export and leave the deep checks to the parent method

I like this idea, thanks @averissimo! Do you think we should use `self$get_active_module_input(input)` instead of `ns(input)?` Also, we need to introduce the `self$get_active_module_export` method. WDYT?

averissimo:
Yes, we should! Great catch

kartikeyakirar:
Following our discussion with @averissimo, it appears that self$get_active_module_... methods are not useful as they return a value that is not being used. However, I have added the rest of the great suggestions here: [a2f4392](https://github.com/insightsengineering/teal/pull/1172/commits/a2f439220ef09ce1f1d684f5e64f74365575d5fc).

R/TealAppDriver.R
```diff
@@ -407,6 +407,32 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
+
+      if (!rlang::is_missing(input) && checkmate::test_string(input, min.chars = 1)) input <- ns(input)
+      if (!rlang::is_missing(output) && checkmate::test_string(output, min.chars = 1)) output <- ns(output)
+      if (!rlang::is_missing(export) && checkmate::test_string(export, min.chars = 1)) export <- ns(export)
+
+      self$wait_for_value(
+        input = input,
+        output = output,
+        export = export,
+        timeout = private$idle_timeout,
```
averissimo, line 433: This might clash with `...` if `timeout = 1000` is passed when called.

I recommend you remove `timeout` as #1171 will use TealAppDriver's default "idle_timeout"

averissimo (approved):
Looks good.

One small comment about `timeout`. After that it can be merged :-)

PR #1182, opened 2024-03-25
Expose method in `TealAppDriver` that checks if elements are visible

### LINKED ISSUE #1181: [Feature Request]: Add function that checks if elements on page are visible (`TealAppDriver`)
Between current `main` and #1150 there are 2 custom implementations of this feature.

I suggest we add a `test_visibility` method based on `js: checkVisibility()` [browser API](https://developer.mozilla.org/en-US/docs/Web/API/Element/checkVisibility)

```r
    #' @description
    #' Test if CSS elements are visible on the page via Javascript call.
    #' @param selector (`character`) CSS selector to test visibility.
    #' A `CSS` id will return only one element if the UI is well formed.
    #' @param content_visibility_auto, opacity_property, visibility_property (`logical`) See more information
    #' on <https://developer.mozilla.org/en-US/docs/Web/API/Element/checkVisibility>.
    #'
    #' @return Logical vector with all occurences of the selector.
    test_visibility = function(selector,
                               content_visibility_auto = FALSE,
                               opacity_property = FALSE,
                               visibility_property = FALSE) {
      checkmate::assert_string(selector)
      checkmate::assert_flag(content_visibility_auto)
      checkmate::assert_flag(opacity_property)
      checkmate::assert_flag(visibility_property)
      unlist(
        self$get_js(
          sprintf(
            "Array.from(document.querySelectorAll('%s')).map(el => el.checkVisibility({%s, %s, %s}))",
            selector,
            # Extra parameters
            sprintf("contentVisibilityAuto: %s", tolower(content_visibility_auto)),
            sprintf("opacityProperty: %s", tolower(opacity_property)),
            sprintf("visibilityProperty: %s", tolower(visibility_property))
          )
        )
      )
    },
```

averissimo:
# Pull Request

Fixes #1181

#### Changes description

- Adds new method that uses JS API to check if element is visible


R/TealAppDriver.R
```diff
@@ -255,6 +248,35 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
     },
     #' @description
+    #' Test if CSS elements are visible on the page via Javascript call.
+    #' @param selector (`character`) CSS selector to test visibility.
+    #' A `CSS` id will return only one element if the UI is well formed.
+    #' @param content_visibility_auto,opacity_property,visibility_property (`logical`) See more information
+    #' on <https://developer.mozilla.org/en-US/docs/Web/API/Element/checkVisibility>.
+    #'
+    #' @return Logical vector with all occurrences of the selector.
+    test_visibility = function(selector,
```
gogonzo, line 258: Let's stick to convention where `is` suggests logical output. `test` is not very specific and can mean many things.

```suggestion
    is_visible = function(selector,
```

m7pr: like the idea of `is_visible` or `check_visibility` but `is_visible` is more direct

m7pr:
@averissimo maybe you can also adjust tests where this method would be used? So that once this is merged, tests are already updated?

R/TealAppDriver.R
```diff
@@ -257,6 +250,35 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
       available_datasets[displayed_datasets_index]
     },
     #' @description
+    #' Test if CSS elements are visible on the page via Javascript call.
```
m7pr, line 253:
```suggestion
    #' Test if CSS elements are visible on the page via JavaScript call.
```

R/TealAppDriver.R
```diff
@@ -257,6 +250,35 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
+    #' @param selector (`character`) CSS selector to test visibility.
```
m7pr, line 254:
```suggestion
    #' @param selector (`character(1)`) CSS selector to test visibility.
```

R/TealAppDriver.R
```diff
@@ -257,6 +250,35 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
+    #' A `CSS` id will return only one element if the UI is well formed.
+    #' @param content_visibility_auto,opacity_property,visibility_property (`logical`) See more information
```
m7pr, line 256:
```suggestion
    #' @param content_visibility_auto,opacity_property,visibility_property (`logical(1)`) See more information
```

m7pr: Where is this `more` information : P ?

averissimo: The more information is on the line below

`#' on <https://developer.mozilla.org/en-US/docs/Web/API/Element/checkVisibility>.`

inst/WORDLIST
```diff
@@ -2,6 +2,7 @@ Biomarker
 CDISC
 Forkers
 Hoffmann
+Javascript
```
m7pr, line 5: If we put `JavaScript` in quotes in the documentation, then you will not have this entry

averissimo: `JavaScript` is a known well spelled word. It was removed from this file with your previous suggestion :+1:

R/TealAppDriver.R
```diff
@@ -257,6 +250,35 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
       available_datasets[displayed_datasets_index]
     },
     #' @description
+    #' Test if CSS elements are visible on the page via JavaScript call.
+    #' @param selector (`character(1)`) CSS selector to check visibility.
+    #' A `CSS` id will return only one element if the UI is well formed.
```
m7pr, line 255: @averissimo one last thing. In this sentence CSS is quoted where in other two previous sentences is not. Please unify. Other than that. This is ready to go

m7pr (approved):

R/TealAppDriver.R
averissimo: Good catch, given that `include_css_files` uses `CSS` with backticks, I'm changing all of them to use backticks

Thanks! :bow:

m7pr:
Just used it today in here https://github.com/insightsengineering/teal.modules.general/pull/718/commits/1c631086f45a62017a369084cea11b5f1bcc5b54#diff-9e1057767ea879b12e16e9c988cc2ec02991e30b18e4d87a2cfe91b79ce3548bR75-R81

This is soo great

Issue #1189, opened 2024-04-01
[Feature Request]: Update `TealAppDriver` method for   `mae`  object Handling


kartikeyakirar:
### Feature description

Objective:
Update the `TealAppDriver` class to improve its functionality with Multi-Assay Experiment `(MAE`) objects, focusing on extending existing capabilities and introducing new methods specifically for adding, setting, and removing filters related to MAE objects.

Background:
The MAE object filtering mechanism in our application is structured into two main sections: one for adding subjects and the other for adding experiments. The latter includes managing gene IDs and sample data, such as row and column data.

![image](https://github.com/insightsengineering/teal/assets/6700955/a0418635-553a-4716-af63-b1ceb263a8d6)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [X] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [X] I agree to follow this project's Security Policy.

PR #1190, opened 2024-04-01
updated TealAppDriver method for adding , removing and setting filter for `MAE` objects

kartikeyakirar:
fixes https://github.com/insightsengineering/teal/issues/1189

- [x] Adding filter
- [x] Setting filter
- [x] removing filter

### DIFF
--- a/R/TealAppDriver.R
+++ b/R/TealAppDriver.R
@@ -392,50 +392,50 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
           self$active_filters_ns(),
           dataset_name
         )
-      } else{
+      } else {
         remove_selector <- switch(filter_type,
-                                  `data.frame` = {
-                                    sprintf(
-                                      "#%s-active-%s-filter-%s_%s-remove",
-                                      self$active_filters_ns(),
-                                      dataset_name,
-                                      dataset_name,
-                                      var_name
-                                    )
-                                  },
-                                  subject = {
-                                    sprintf(
-                                      "#%s-active-%s-subjects-%s_%s-remove",
-                                      self$active_filters_ns(),
-                                      dataset_name,
-                                      dataset_name,
-                                      var_name
-                                    )
-                                  },
-                                  gene = {
-                                    checkmate::check_string(sample_name)
-                                    sprintf(
-                                      "#%s-active-%s-%s-%s_%s_%s_subset-remove",
-                                      self$active_filters_ns(),
-                                      dataset_name,
-                                      sample_name,
-                                      dataset_name,
-                                      var_name,
-                                      sample_name
-                                    )
-                                  },
-                                  sample = {
-                                    checkmate::check_string(sample_name)
-                                    sprintf(
-                                      "#%s-active-%s-%s-%s_%s_%s_select-remove",
-                                      self$active_filters_ns(),
-                                      dataset_name,
-                                      sample_name,
-                                      dataset_name,
-                                      var_name,
-                                      sample_name
-                                    )
-                                  }
+          `data.frame` = {
+            sprintf(
+              "#%s-active-%s-filter-%s_%s-remove",
+              self$active_filters_ns(),
+              dataset_name,
+              dataset_name,
+              var_name
+            )
+          },
+          subject = {
+            sprintf(
+              "#%s-active-%s-subjects-%s_%s-remove",
+              self$active_filters_ns(),
+              dataset_name,
+              dataset_name,
+              var_name
+            )
+          },
+          gene = {
+            checkmate::check_string(sample_name)
+            sprintf(
+              "#%s-active-%s-%s-%s_%s_%s_subset-remove",
+              self$active_filters_ns(),
+              dataset_name,
+              sample_name,
+              dataset_name,
+              var_name,
+              sample_name
+            )
+          },
+          sample = {
+            checkmate::check_string(sample_name)
+            sprintf(
+              "#%s-active-%s-%s-%s_%s_%s_select-remove",
+              self$active_filters_ns(),
+              dataset_name,
+              sample_name,
+              dataset_name,
+              var_name,
+              sample_name
+            )
+          }
         )
       }
       self$click(
--- a/R/TealAppDriver.R
+++ b/R/TealAppDriver.R
@@ -289,79 +289,53 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
       active_filters[[dataset_name]]
     },
     #' @description
-    #' Add a new variable from the dataset to be filtered.
-    #'
-    #' @param dataset_name (character) The name of the dataset to add the filter variable to.
-    #' @param var_name (character) The name of the variable to add to the filter panel.
-    #' @param ... Additional arguments to be passed to `shinytest2::AppDriver$set_inputs`
-    #'
-    #' @return The `TealAppDriver` object invisibly.
-    add_filter_var = function(dataset_name, var_name, ...) {
-      checkmate::check_string(dataset_name)
-      checkmate::check_string(var_name)
-      self$set_input(
-        sprintf(
-          "%s-add-%s-filter-var_to_add",
-          self$active_filters_ns(),
-          dataset_name
-        ),
-        var_name,
-        ...
-      )
-      invisible(self)
-    },
-    #' @description
-    # Add a new subject filter variable for MAE datasets.
-    #'
-    #' @param dataset_name (character) The name of the dataset to add the filter variable to.
-    #' @param var_name (character) The name of the variable to add to the filter panel.
-    #' @param ... Additional arguments to be passed to `shinytest2::AppDriver$set_inputs`
-    #'
-    #' @return The `TealAppDriver` object invisibly.
-    add_mae_subjects_filter_var = function(dataset_name, var_name, ...) {
-      checkmate::check_string(dataset_name)
-      checkmate::check_string(var_name)
-      self$set_input(
-        sprintf(
-          "%s-add-%s-subjects-var_to_add",
-          self$active_filters_ns(),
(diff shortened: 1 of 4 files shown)


R/TealAppDriver.R
```diff
@@ -311,39 +311,70 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
       active_filters[[dataset_name]]
     },
     #' @description
-    #' Add a new variable from the dataset to be filtered.
+    #' Adds a new filter variable to the dataset. Supports both standard datasets and MAE datasets.
     #'
     #' @param dataset_name (character) The name of the dataset to add the filter variable to.
+    #' @param sample_name (character) The name of the sample/experiment.
```
vedhav, line 317:
```suggestion
    #' @param sample_name (character) The name of the sample/experiment. Applicable only if the dataset is a `MultiAssayExperiment` object.
```

R/TealAppDriver.R
```diff
@@ -311,39 +311,70 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
     },
     #' @description
     #' Remove an active filter variable of a dataset from the active filter variables panel.
     #'
     #' @param dataset_name (character) The name of the dataset to remove the filter variable from.
     #' If `NULL`, all the filter variables will be removed.
+    #' @param sample_name (character) The name of the sample, required for "gene" and "sample" filter types.
     #' @param var_name (character) The name of the variable to remove from the filter panel.
     #' If `NULL`, all the filter variables of the dataset will be removed.
+    #' @param filter_type (character) Specifies the type of filter: "data.frame", "subject", "gene", or "sample".
```
vedhav, line 370: Same comment as above. As we have `sample_name` as `NULL` so it makes more sense for me when `filter_type` is also `NULL`. Perhaps it can be called something else more specific like `mae_entity`/`mae_element`/`mae_component`.

R/TealAppDriver.R
```diff
@@ -311,39 +311,70 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
     #' @return The `TealAppDriver` object invisibly.
-    add_filter_var = function(dataset_name, var_name, ...) {
+    add_filter_var = function(dataset_name, sample_name, var_name, filter_type = "data.frame", ...) {
       checkmate::check_string(dataset_name)
       checkmate::check_string(var_name)
-      self$set_input(
-        sprintf(
+      checkmate::assert_choice(filter_type, c("data.frame", "subject", "gene", "sample"))
+
+      input_id <- switch(filter_type,
```
gogonzo, line 328: Before I take away PR rejection let's have a discussion about why this should be in teal.

gogonzo (dismissed):

R/TealAppDriver.R
vedhav: The idea was that `teal.slice` would use the `TealAppDriver` to test the teal features. One approach was to extend the `TealAppDriver` in `teal.slice` to drive MAE selections and the alternative was to make all possible "app driving" done in the teal package itself.

gogonzo: Why does `teal.slice` need `teal` to test itself? (It sets a reverse dependency - on Suggests)

kartikeyakirar: I believe the `TealAppDriver` is primarily meant for testing the tm_modules in `teal` ,  such as `teal.modules.general`, `teal.modules.clinical`, and `teal.modules.hermes`, among others.  and incorporating methods directly into the Driver for `MAE` would be more beneficial. For `teal.slice` all examples using conventional methods, and we can test it with `AppDriver`. I think we don't need  TealAppDriver for `teal.slice` specifically

R/TealAppDriver.R
```diff
@@ -311,39 +311,70 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
     #'
     #' @param dataset_name (character) The name of the dataset to add the filter variable to.
+    #' @param sample_name (character) The name of the sample/experiment.
     #' @param var_name (character) The name of the variable to add to the filter panel.
+    #' @param filter_type (character) Specifies the type of filter. Default is "data.frame".
     #' @param ... Additional arguments to be passed to `shinytest2::AppDriver$set_inputs`
     #'
     #' @return The `TealAppDriver` object invisibly.
-    add_filter_var = function(dataset_name, var_name, ...) {
+    add_filter_var = function(dataset_name, sample_name, var_name, filter_type = "data.frame", ...) {
```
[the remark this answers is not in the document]
kartikeyakirar: I think  setting  specific default value (e.g., "data.frame" for filter_type) rather than NULL makes the function's intended use and behaviour more clearer to the user.  we internally anyways going to change it 'data.frame`. It communicates what the expected default behavior in definition  without requiring the user to dive into the function's logic.

R/TealAppDriver.R
kartikeyakirar: okay in this case keeping it null make sense as we won't have `data.frame` only c("subject", "gene", "sample"). I am okay with this . I will make changes anyway because the filter type was a bit vague.

R/TealAppDriver.R
gogonzo: >  believe the TealAppDriver is primarily meant for testing the tm_modules in teal (if I am not mistaken) such as teal.modules.general, teal.modules.clinical, and teal.modules.hermes

Then let's test the modules and not filters in the module packages. If we need to test filters outside of `teal.slice`  then we probably need a testing class for `FilteredData` and incorporate it somehow in `tealAppDriver` - `tealAppDriver` shouldn't handle stuff like id of the filter-panel element. Reason for this is to limit dependency on internal stuff in `teal.slice` (like id of the card element). Changes in `teal.slice` will require us to make changes in teal tests also (even if the public api remains unchanged).

gogonzo: I think we can resolve problem of violating encapsulation of the teal.slice in two ways:
1. Don't test filter panel outside of the teal.slice.
2. Enable teal.slice testing methods for teal

I guess consequences of (1) are clear. I don't think teal.modules.hermes should test whether it is possible to add a MAE filter. If we cover it in `teal.slice` there is no need to test it again in teal.modules.hermes - but we need to make sure that namespace of the filter-panel is set correctly.

Design is a challange for (2) - `tealAppDriver` can't inherit from both `appDriver` and `tealSliceAppDriver`. Which makes me thinking that maybe `tealAppDriver` class is not needed but rather just functions?

Consider this:
```r
appDr <- shinytest2::appDriver(<tealapp>)

teal:::get_html_rvest(appDr)
teal.slice:::add_filter(appDr, <filter-id>, filter_panel_id = "teal-main_ui-filter_panel")

```

WDYT?

kartikeyakirar: > I guess consequences of (1) are clear. I don't think teal.modules.hermes should test whether it is possible to add a MAE filter. If we cover it in teal.slice there is no need to test it again in teal.modules.hermes - but we need to make sure that namespace of the filter-panel is set correctly.

I agree `MAE` filter can be just checked in teal.slice and we dont't need to test it again in modules.  

> tealAppDriver class is not needed but rather just functions?

This is a good idea to keep testing functions separate. We can still keep the TealAppDriver and create testing functions in teal.slice, for example. (@vedhav  can you share your thoughts on this)
```

appDr <- teal:::TealAppDriver(<tealapp>)
teal.slice:::add_filter(appDr, <filter-id>) # return TealAppDriver
```

vedhav: I think we have to forget about `TealAppDriver` completely when it comes to testing `teal.slice`.
We are testing it using a classic shiny app where it is easy to control the namespace. So the `tealSliceAppDriver` will just use the `appDriver`. Some of the driving logic for this can be copied and maintained in `teal.slice` separately, there is some overlap for sure. But, we do not want the entire `tealAppDriver` to drive a shiny app created with `teal.slice`.

When it comes to this PR, it is adding the MAE filter functions so that the module packages can use it. We still need it. And `teal.slice` is a dependency of `teal` and it is okay to add driving code that is dependent on `teal.slice`.
We can test it as part of testing the `TealAppDriver` itself, there are some features of the `TealAppDriver` that are not tested yet.

gogonzo: @vedhav `teal` is data format independent. There is no assumption on a class of the data. This solution is bad, it checks the filter type which is a internal part of the `teal.slice` package. The link is not obvious and  we will have to deal with this each time there will be changes in relevant code in `teal.slice`.  I will not block the sprint objective, there are more important things to discuss.

kartikeyakirar: Yes, this feature was only required for `teal.modules.hermes` and `teal.modules.helios `. And Following your suggestion I am drafting a plan to incorporate the MAE test within `teal.slice`,  I am also considering whether to consolidate all `teal.slice` related methods, such as removing, adding, and setting filters, exclusively within `teal.slice`.

kartikeyakirar: I'm moving this PR to draft for now. I will close it once the Teal.slice PR is live.

kartikeyakirar: Not yet. I prioritize `tmc` so I did not get time to work on it.

kartikeyakirar:
@marcin, I've kept this PR as a placeholder for now. till the new PR on `teal.slice` to test the filters for ADaM and MAE data.

TODO:

 

- [ ] Remove the methods for adding, removing, and setting filters in tealAppDriver.
- [ ]  Add tests for filters in the teal_slices package.

This isn't exactly blocking the `teal.module.heremrs` and teal.module.hleios shinytest2 conversations, but I marked it as blocked due to some ongoing discussions at that time.

As per our conclusion, we will not be testing filters on modules. Once we remove the methods mentioned above, we will also remove the related tests using filtering functionalities.

Issue #1195, opened 2024-04-09
Failed integration test


donyunardi:
We got failed integration test and need to fix this:

```r
* checking tests ...
  Running ‘testthat.R’
 ERROR
Running the tests in ‘tests/testthat.R’ failed.
Complete output:
  > pkg_name <- "teal"
  > library(pkg_name, character.only = TRUE)
  Loading required package: shiny
  Loading required package: teal.data
  Loading required package: teal.code
  Loading required package: teal.slice
  Registered S3 method overwritten by 'teal':
    method        from      
    c.teal_slices teal.slice
  
  You are using teal version 0.15.2.9022
  
  Attaching package: 'teal'
  
  The following objects are masked from 'package:teal.slice':
  
      as.teal_slices, teal_slices
  
  > testthat::test_check(pkg_name)
  [ FAIL 8 | WARN 0 | SKIP 0 | PASS 479 ]
  
  ══ Failed tests ════════════════════════════════════════════════════════════════
  ── Error ('test-shinytest2-filter_panel.R:48:3'): e2e: filtering a module-specific filter is refected in other shared module ──
  Error in `app$get_active_data_filters("iris")`: Assertion on 'dataset_name' failed: Must be a subset of the empty set, i.e. also empty.
  Backtrace:
      ▆
   1. ├─testthat::expect_equal(...) at test-shinytest2-filter_panel.R:48:3
   2. │ └─testthat::quasi_label(enquo(object), label, arg = "object")
   3. │   └─rlang::eval_bare(expr, quo_get_env(quo))
   4. └─app$get_active_data_filters("iris")
   5.   └─checkmate::assert_subset(dataset_name, datasets)
   6.     └─checkmate::makeAssertion(x, res, .var.name, add)
   7.       └─checkmate:::mstop(...)
  ── Error ('test-shinytest2-filter_panel.R:87:3'): e2e: filtering a module-specific filter is not refected in other unshared modules ──
  Error in `app$get_active_data_filters("mtcars")`: Assertion on 'dataset_name' failed: Must be a subset of the empty set, i.e. also empty.
  Backtrace:
      ▆
   1. ├─testthat::expect_equal(...) at test-shinytest2-filter_panel.R:87:3
   2. │ └─testthat::quasi_label(enquo(object), label, arg = "object")
   3. │   └─rlang::eval_bare(expr, quo_get_env(quo))
   4. └─app$get_active_data_filters("mtcars")
[...]

PR #1200, opened 2024-04-11
Tests wait for idle by default after setting module input

averissimo:
# Pull Request

### Changes description

- [feature] Wait for app to be idle by default (skips if `wait = FALSE` parameter is passed)
- [fix] Removes 5s waiting on every `app_driver$click()`
  - It was waiting for UI to be stable for 5s, instead of setting the timeout.
  - Current default timeout is set on constructor at 20s

### DIFF
--- a/R/TealAppDriver.R
+++ b/R/TealAppDriver.R
@@ -232,6 +232,8 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
         value,
         ...
       )
+      dots <- rlang::list2(...)
+      if (!isFALSE(dots[[wait]])) self$wait_for_idle() # Default behavior is to wait
       invisible(self)
     },
     #' @description
--- a/R/TealAppDriver.R
+++ b/R/TealAppDriver.R
@@ -73,7 +73,7 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
     #' @param ... arguments passed to parent [`shinytest2::AppDriver`] `click()` method.
     click = function(...) {
       super$click(...)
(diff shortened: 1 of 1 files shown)


R/TealAppDriver.R
```diff
@@ -73,7 +73,7 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
     #' @param ... arguments passed to parent [`shinytest2::AppDriver`] `click()` method.
     click = function(...) {
       super$click(...)
-      self$wait_for_idle(5000)
+      self$wait_for_idle()
```
m7pr, line 76: @averissimo who set 5000 in here :D?

m7pr (approved):

R/TealAppDriver.R
averissimo: Oh I should've run git blame first. Do you think that we need to change the duration parameter from 500ms to something else?

Or was this a confusion with the `timeout` parameter?

averissimo: Note that the `duration` is the milliseconds duration that the UI needs to be stable to be considered idle.

It will try to do this for 20s (as that's the default timeout value for `TealAppDriver`)

donyunardi: That was me. I thought we wanted to increase the wait time for idle.
https://github.com/insightsengineering/teal/issues/1195#issuecomment-2047788486

averissimo: Currently, the timeout is being set to 20s, so I think we can merge this and change it appropriately if the integration fails.

PR #1202, opened 2024-04-12
Fixes integration tests when Chrome >=105 is not installed

donyunardi:
Related to #1195 

I'm focusing on `test-shinytest2-filter_panel` to gain more insight during the integration test.

#### Changes description

- Use alternative method of checking visibility of element when `el.checkVisibility()` method does not exists (is `null`)
- Fix tests that were relying on the order of unordered filters (categorical)
  - `expect_setequal` instead of `expect_equal` to compare vectors.

### DIFF
--- a/R/TealAppDriver.R
+++ b/R/TealAppDriver.R
@@ -65,6 +65,20 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
         )
       )
 
+      # Check for minimum version of Chrome that supports the tests
+      #  - Element.checkVisibility was added on 105
+      chrome_version <- numeric_version(
+        gsub(
+          "[[:alnum:]_]+/", # Prefix that ends with forward slash
+          "",
+          self$get_chromote_session()$Browser$getVersion()$product
+        ),
+        strict = FALSE
+      )
+
+      testthat::skip_if_not(!is.na(chrome_version) && chrome_version >= 105)
+      # end od check
+
       private$set_active_ns()
       self$wait_for_idle()
     },
--- a/R/TealAppDriver.R
+++ b/R/TealAppDriver.R
@@ -76,8 +76,20 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
         strict = FALSE
       )
 
-      testthat::skip_if(is.na(chrome_version), "Problem getting Chrome version.")
-      testthat::skip_if(chrome_version < 105, "Chrome version is not supported, please upgrade.")
+      required_version <- 105
+
+      testthat::skip_if(
+        is.na(chrome_version),
+        "Problem getting Chrome version, please contact the developers."
+      )
+      testthat::skip_if(
+        chrome_version < required_version,
+        sprintf(
+          "Chrome version '%s' is not supported, please upgrade to %d or above.",
+          chrome_version,
+          required_version
+        )
+      )
       # end od check
 
       private$set_active_ns()
--- a/R/TealAppDriver.R
+++ b/R/TealAppDriver.R
@@ -85,7 +85,7 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
       testthat::skip_if(
         chrome_version < required_version,
         sprintf(
-          "Chrome version '%s' is not supported, please upgrade to %d or above.",
+          "Chrome version '%s' is not supported, please upgrade to '%d' or higher",
           chrome_version,
           required_version
         )
--- a/R/TealAppDriver.R
+++ b/R/TealAppDriver.R
@@ -249,6 +249,10 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
           self$active_filters_ns()
         )
       )
+
+      self$log_message(sprintf("Value for displayed_datasets_index: %s", toString(available_datasets)))
+      self$log_message(sprintf("Value for available_datasets: %s", toString(displayed_datasets_index)))
+
       available_datasets[displayed_datasets_index]
     },
     #' @description
--- a/R/TealAppDriver.R
+++ b/R/TealAppDriver.R
@@ -250,8 +250,8 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
         )
       )
 
-      self$log_message(sprintf("Value for displayed_datasets_index: %s", toString(available_datasets)))
-      self$log_message(sprintf("Value for available_datasets: %s", toString(displayed_datasets_index)))
+      self$log_message(sprintf("Value for displayed_datasets_index: %s", toString(displayed_datasets_index)))
+      self$log_message(sprintf("Value for available_datasets: %s", toString(available_datasets)))
 
       available_datasets[displayed_datasets_index]
     },
--- a/R/TealAppDriver.R
+++ b/R/TealAppDriver.R
@@ -268,7 +268,13 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
       checkmate::assert_flag(content_visibility_auto)
       checkmate::assert_flag(opacity_property)
       checkmate::assert_flag(visibility_property)
-      result <- unlist(
+
+      testthat::skip_if(
+        app$get_js("typeof Element.prototype.checkVisibility === 'function'"),
+        "Element.prototype.checkVisibility is not supported in the current browser."
+      )
+
+      unlist(
         self$get_js(
           sprintf(
             "Array.from(document.querySelectorAll('%s')).map(el => el.checkVisibility({%s, %s, %s}))",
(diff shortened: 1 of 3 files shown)


tests/testthat/test-shinytest2-filter_panel.R
```diff
@@ -45,6 +45,10 @@ testthat::test_that("e2e: filtering a module-specific filter is refected in othe
     )
   )
 
+  get_active_filter_vars <- app$get_active_filter_vars()
+  print(app$get_logs())
+  print(app$get_html("#teal-main_ui_container"))
```
donyunardi, line 50: This will give us the complete html code.

pawelru, line 50: are you really sure you want to introduce print in the codebase?

donyunardi: I should have been clearer on my intention.
My plan was to just temporarily set this up to find more insight on what's going on.
Another alternative is to use the `log_message()` method and write it somewhere so we can retrieve it (need to confirm with IDR where that location would be).

Since then, I also learned that we can run CI test using a branch.
If that's the case, the let's just run it with this branch and close the branch once we learn more.
I'll reach out to IDR to get more info.

donyunardi:
Convert to draft so it won't be accidentally merged.

donyunardi:
I downloaded the non-validated ocean image, install chrome, install teal and families, and I still can't reproduce the error.

One thing to note is that I originally tried to install just chromium-browser, but then I encountered a blocker because it wanted to install it from Snap. Apparently, there's another setting that you'd need to do to get snap running and after couple tries, I gave up. Instead, I downloaded the chrome debian package and install it to the container.
```
wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
apt install ./google-chrome-stable_current_amd64.deb
```

I'm curious about the contents of our CI image. Is it a full Chrome installation like what I just did, or is it the Snap version of Chromium? Does it make a difference?

@cicdguy @walkowif 
Please, any advise here?

donyunardi:
Oh okay, so you install the whole Chrome as well. 
Then, I am clueless on what to do next.

Next thing we can try is to run CI test based on this branch where I use `print` to quickly see what happened.
Or, I can save the log it in current directory and IDR can help locate this file.

Any other ideas?

donyunardi:
@walkowif helped setup running CI test on this branch.

Here's the result log:
https://nest.pages.roche.com/-/automation/systems-integration-tests/-/jobs/45805756/artifacts/public/logs/check-teal.html

My suspicion is confirmed that there's a problem with `get_active_filter_vars` method during CI, specifically on `displayed_datasets_index` variable:
https://github.com/insightsengineering/teal/blob/6042f560e66a9317f2b6e905f54cc3101851ffd5/R/TealAppDriver.R#L241-L257

The value of `displayed_datasets_index` is null as you can see in the log:
```
  {shinytest2} R  info   16:12:43.47 Value for displayed_datasets_index: 
  {shinytest2} R  info   16:12:43.47 Value for available_datasets: iris, mtcars
```

However, it's odd that it's able to retrieve the values for `available_datasets`, which is embedded in the `active-filter_active_vars_contents` class. 

I wonder if it has something to do with the selector statement inside `is_visible` method because I can clearly see the span tag in the log file. Here's an example for one of the module:
```
<div id=\"teal-main_ui-root-module_1-module_filter_panel-active-filter_active_vars_contents\">\n                  <span id=\"teal-main_ui-root-module_1-module_filter_panel-active-iris\">
```
`is_visible` should return TRUE.

Any idea? @insightsengineering/nest-core-dev

donyunardi:
@averissimo 
Feel free to use this branch (add new commits, etc) to investigate further.

If you need to trigger the CI on this branch, please go to:
https://code.roche.com/nest/automation/systems-integration-tests/-/pipelines/new
Select branch `troubleshoot-teal` and `Run pipeline`.

tests/testthat/test-shinytest2-filter_panel.R
averissimo: removed

averissimo:
Fixed the problem with this PR, the integration tests are running again without the logs before making it ready for review.

An alternative solution would be to install Chrome >=105 on the container image. @cicdguy @walkowif  (Chrome 90 is 3 years old)

averissimo:
Pipeline is confirmed successful: https://nest.pages.roche.com/-/automation/systems-integration-tests/-/jobs/46148438/artifacts/public/logs/check-teal.html

The job fails, but not due to teal.

m7pr:
You are the legend @averissimo

walkowif:
@averissimo I checked that it should be possible to install Chromium 120 in the images with R 4.3.
@cicdguy We can discuss if there are any potential implications of such change.

averissimo:
Good to know!

This PR patches the problem and creates a fallback solution for the lack of support of the relevant API.

However, I think it would be best to bridge the gap between the versions of chrome among platforms (local dev, GH and integration).

R/TealAppDriver.R
```diff
@@ -279,6 +280,22 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
+            sprintf(
+              "Array.from(document.querySelectorAll('%s')).map(%s)",
+              selector,
+              fun_body
+            )
+          )
+        )
+      } else {
+        result
+      }
```
pawelru, line 298: I really appreciate the finding here. Now I am wondering - is it better to provide the support for an old dependency or skip the test completely. The rationale here is that after upgrading (that everyone agrees it's needed) some part of code would leave not executed which I believe is a bad thing.
Taking this idea even further - we can introduce such requirements (I believe that this is not the only one) in the class constructor and stop very early before app init.
What do you think?

averissimo: I considered this and there's a solution to check for API support and skip: `typeof Element.prototype.checkVisibility === "function"` (js)

However, it didn't feel right to skip a test during integration tests, hence the fallback solution.

I'm fine with adding an automatic skip \*\*as long as\*\* the integration tests image is updated (see previous phrase :sweat_smile:)

>  Taking this idea even further - we can introduce such requirements (I believe that this is not the only one) in the class constructor and stop very early before app init.

This can be done with user agent (`navigator.userAgent` (js)). I didn't see a way of getting this info from `{chromote}` session, but maybe there is.

**By skipping I'm just worried about no actions being taken** if the image is reverted to a previous version or a new CI is introduced with an older version of chrome.

Failure here shows us that something is going on.

walkowif:
Just wanted to confirm that the images used for integration tests have been updated to include Chromium 120.

R/TealAppDriver.R
pawelru: My assumption here was that we will be looking into the test logs periodically and capture that some tests are unintentionally skipped. (That's actually the reason behind introducing non --as-cran pipeline but that's a separate story)
Thinking about this more - I think that skip would be more in line with testthat2 ~philosophy~ common practice. Please have a look at the purpose `skip_if_not_installed()` or any other `skip_if_xyz()` function - it is to do the skip (with a meaningful message) if some condition is met. That means that missing / insufficient version of test pre-requisites does not fail the tests. I consider testthat2, chromote and JS version as test pre-requisites. Note that we are not providing alternatives if they are missing.

averissimo: Ok, that sounds good as there is (will be) a plan in place to catch the skipped tests :+1: 

I'll try to find a way to get the versioning in the least `hacky` way possible. Hopefully, via `{chromote}` itself :sweat_smile:

averissimo: Added both a minimum version and a API-specific check.

I've added the general check to allow for future requirements that cannot be checked explicitly. But realistically, it could be removed for now.

### Minimum version

https://github.com/insightsengineering/teal/blob/09bfc1290dd6d089feabcb86e9ce264f8004704e/R/TealAppDriver.R#L70-L80

note that the version needs to be parsed

```r
# app <- TealAppDriver$new(...)
app$get_chromote_session()$Browser$getVersion()$product
#> [1] "HeadlessChrome/124.0.6367.60"
```

### API specific (less arbitrary and easier to maintain)

https://github.com/insightsengineering/teal/blob/09bfc1290dd6d089feabcb86e9ce264f8004704e/R/TealAppDriver.R#L287-L290

R/TealAppDriver.R
```diff
@@ -65,6 +65,21 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
+        gsub(
+          "[[:alnum:]_]+/", # Prefix that ends with forward slash
+          "",
+          self$get_chromote_session()$Browser$getVersion()$product
+        ),
+        strict = FALSE
+      )
+
+      testthat::skip_if(is.na(chrome_version), "Problem getting Chrome version.")
+      testthat::skip_if(chrome_version < 105, "Chrome version is not supported, please upgrade.")
```
pawelru, line 80: please include `chrome_version` in the err message - it would help debugging

R/TealAppDriver.R
pawelru: awesome! thank you!

R/TealAppDriver.R
averissimo: Oh, great idea!

Error message in the case required version is 125 (to force it, instead of 105 that we're targetting)

```r
Error: Reason: Chrome version '124.0.6367.60' is not supported, please upgrade to '125' or higher
```

pawelru (approved):
This looks good. Thank you! Great work done here!

averissimo:
@walkowif , I believe the pipeline that was created for this branch can be removed (`nest/automation/systems-integration-tests@troubleshoot-teal`)

Thanks for the help with setting that up and for updating Chromium :1st_place_medal:

PR #1210, opened 2024-04-29
Extend the table fetcher to handle DT tables too

vedhav:
Extend the `get_active_module_tws_output` to also work for `DT::DTOutput` which is now renamed to `get_active_module_table_output` along with `get_active_module_plot_output`

Example to test:

```r
devtools::load_all("../teal")
library(teal.modules.clinical)

data <- teal_data()
data <- within(data, {
  ADSL <- teal.data::rADSL
  ADLB <- teal.data::rADLB
})

datanames <- c("ADSL", "ADLB")
teal.data::datanames(data) <- datanames
teal.data::join_keys(data) <- teal.data::default_cdisc_join_keys[datanames]

app_driver <- TealAppDriver$new(
  data = data,
  modules = tm_t_pp_laboratory(
    label = "Vitals",
    dataname = "ADLB",
    patient_col = "USUBJID",
    paramcd = teal.transform::choices_selected(
      choices = teal.transform::variable_choices(data[["ADLB"]], "PARAMCD"),
      selected = "PARAMCD"
    ),
    param = teal.transform::choices_selected(
      choices = teal.transform::variable_choices(data[["ADLB"]], "PARAM"),
      selected = "PARAM"
    ),
    timepoints = teal.transform::choices_selected(
      choices = teal.transform::variable_choices(data[["ADLB"]], "ADY"),
      selected = "ADY"
    ),
    anrind = teal.transform::choices_selected(
      choices = teal.transform::variable_choices(data[["ADLB"]], "ANRIND"),
      selected = "ANRIND"
    ),
    aval_var = teal.transform::choices_selected(
      choices = teal.transform::variable_choices(data[["ADLB"]], "AVAL"),
      selected = "AVAL"
    ),
    avalu_var = teal.transform::choices_selected(
      choices = teal.transform::variable_choices(data[["ADLB"]], "AVALU"),
      selected = "AVALU"
    )
  )
)

app_driver$get_active_module_table_output("lab_values_table", 2)
```

### DIFF
--- a/R/TealAppDriver.R
+++ b/R/TealAppDriver.R
@@ -246,19 +246,19 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
     #' Get the output from the module's `teal.widgets::table_with_settings` or `DT::DTOutput` in the `teal` app.
     #' This function will only access outputs from the name space of the current active teal module.
     #'
-    #' @param tws (`character(1)`) `table_with_settings` namespace name.
+    #' @param table_id (`character(1)`) The id of the table in the active teal module's name space.
     #' @param which (integer) If there is more than one table, which should be extracted.
     #' @param is_dt_table (logical) If the table is a `DT::DTOutput`.
     #' By default it will look for  a table that is built using `teal.widgets::table_with_settings`.
     #'
     #' @return The data.frame with table contents.
-    get_active_module_table_output = function(tws, which = 1, is_dt_table = FALSE) {
+    get_active_module_table_output = function(table_id, which = 1, is_dt_table = FALSE) {
       checkmate::check_number(which, lower = 1)
-      checkmate::check_string(tws)
+      checkmate::check_string(table_id)
       table_selector <- ifelse(
         is_dt_table,
-        self$active_module_element(tws),
-        self$active_module_element(sprintf("%s-table-with-settings", tws))
+        self$active_module_element(table_id),
+        self$active_module_element(sprintf("%s-table-with-settings", table_id))
       )
       table <- table_selector %>%
         self$get_html_rvest() %>%
@@ -273,13 +273,13 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
     #' Get the output from the module's `teal.widgets::plot_with_settings` in the `teal` app.
     #' This function will only access plots from the name space of the current active teal module.
     #'
-    #' @param pws (`character(1)`) `plot_with_settings` namespace name.
+    #' @param plot_id (`character(1)`) The id of the plot in the active teal module's name space.
     #'
     #' @return The `src` attribute as `character(1)` vector.
-    get_active_module_plot_output = function(pws) {
-      checkmate::check_string(pws)
+    get_active_module_plot_output = function(plot_id) {
+      checkmate::check_string(plot_id)
       self$get_attr(
-        self$active_module_element(sprintf("%s-plot_main > img", pws)),
+        self$active_module_element(sprintf("%s-plot_main > img", plot_id)),
         "src"
       )
     },
--- a/R/TealAppDriver.R
+++ b/R/TealAppDriver.R
@@ -248,19 +248,13 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
     #'
     #' @param table_id (`character(1)`) The id of the table in the active teal module's name space.
     #' @param which (integer) If there is more than one table, which should be extracted.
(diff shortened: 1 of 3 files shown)


R/TealAppDriver.R
```diff
@@ -253,13 +253,14 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
       checkmate::check_number(which, lower = 1)
       checkmate::check_string(tws)
-      table <-
-        rvest::html_table(
-          self$get_html_rvest(
-            self$active_module_element(sprintf("%s-table-with-settings", tws))
-          ),
-          fill = TRUE
-        )
+      table_selector <- ifelse(
```
kartikeyakirar, line 256: what if we have both table in module ?

gogonzo (commented):
Here we are, decision about including method to `TealAppDriver` has already some consequences. 

Name `get_active_module_tws_output` is not accurate anymore, but what is worse is that function doesn't handle single tws case but has hardcoded condition for not-tws. 


Probably it should be called like this from the beginning. 

```r
app_driver$get_active_module_table_output(
  table_id = "my_table-table-with-settings" 
)
```

R/TealAppDriver.R
vedhav: I did not see cases like this in all the 36 modules in tmc.

vedhav: But, I see your point, I will make this as an argument to make the selection.

vedhav:
Yeah, true. This should be `get_active_module_table_output`, And, it's companion `get_active_module_plot_output` which will slowly extend to other plot types if observed.

R/TealAppDriver.R
vedhav: Done in b36be57b69269c7703ed2843ce7cc84eba087996

R/TealAppDriver.R
```diff
@@ -273,7 +276,7 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
     #' @param pws (`character(1)`) `plot_with_settings` namespace name.
     #'
     #' @return The `src` attribute as `character(1)` vector.
-    get_active_module_pws_output = function(pws) {
+    get_active_module_plot_output = function(pws) {
```
kartikeyakirar, line 279: same here as well https://github.com/insightsengineering/teal/pull/1210/files#r1582990384

R/TealAppDriver.R
```diff
@@ -243,23 +243,26 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
     #' This function will only access outputs from the name space of the current active teal module.
     #'
     #' @param tws (`character(1)`) `table_with_settings` namespace name.
     #' @param which (integer) If there is more than one table, which should be extracted.
+    #' @param is_dt_table (logical) If the table is a `DT::DTOutput`.
+    #' By default it will look for  a table that is built using `teal.widgets::table_with_settings`.
     #'
     #' @return The data.frame with table contents.
-    get_active_module_tws_output = function(tws, which = 1) {
+    get_active_module_table_output = function(tws, which = 1, is_dt_table = FALSE) {
```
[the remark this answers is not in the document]
vedhav: True, `table_id` makes more sense here.

R/TealAppDriver.R
vedhav: Done

gogonzo:
> ```r
> app_driver$get_active_module_table_output(
>   table_id = "my_table-table-with-settings" 
> )
> ```

Function name changed so we'll need to update test everywhere downstream. Why not to apply my suggestion . This should be relatively straightfoward. 

```r
app_driver$get_active_module_table_output(
  table_id = "my_table-table-with-settings" 
)
```


While `"my_table-table-with-settings"` theoreticaly could be handled by `ns_pws` (same as `pws_des_input`) but it is simple enoygh that I could write it manually.

vedhav:
I am also okay with this change. But, I would still prefer to have the `table-with-settings` in once place for two reasons:
1. All the input ids used in our teal module tests can be directly found in the module code. So, I would like to keep it that way if possible. As, it is easy to search for "table" in the code with `teal.widgets::table_with_settings_ui(ns("table"))`.
2. Changing this class in the future will be easy if it is in one place.

vedhav:
Okay, I already see that some of the ids are implemented as you suggested. So, I'm onboard. I'll remove this logic from there.

kartikeyakirar (approved):
LGTM

PR #1211, opened 2024-05-06
Fix shinytest2 app init on R 4.4.0+

vedhav:
Fixes the broken [shinytest2 tests](https://github.com/insightsengineering/teal/actions/runs/8961636987/job/24609457272)

The root cause is that `numeric_version` only accepts character to convert into `numeric_version` class. So typecasting of the numeric version was giving the error.

Example code to test. This works in older R versions. But, in R 4.4.0 it does not work.
```r
# Broken
numeric_version('123') > 120

#Works
numeric_version('123') > '120'
```

### DIFF
--- a/R/TealAppDriver.R
+++ b/R/TealAppDriver.R
@@ -76,7 +76,7 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
         strict = FALSE
       )
 
-      required_version <- 105
+      required_version <- "105"
 
       testthat::skip_if(
         is.na(chrome_version),
@@ -85,7 +85,7 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
       testthat::skip_if(
         chrome_version < required_version,
         sprintf(
-          "Chrome version '%s' is not supported, please upgrade to '%d' or higher",
+          "Chrome version '%s' is not supported, please upgrade to '%s' or higher",
           chrome_version,
           required_version
         )


R/TealAppDriver.R
```diff
@@ -76,7 +76,7 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
         strict = FALSE
       )
 
-      required_version <- 105
+      required_version <- "105"
```
kartikeyakirar, line 79: Can we add a comment explaining the reason for this change? This will help prevent confusion for anyone reviewing it in the future.

kartikeyakirar (approved):
LGTM

R/TealAppDriver.R
vedhav: I don't think so. The `?numeric_version` doc is clear that it only works with character vector.

```r
Arguments:

       x: for the creators, a character vector with suitable numeric
          version strings (see ‘Details’); for ‘package_version’,
          alternatively an R version object as obtained by ‘R.version’.
          For ‘as.numeric_version’ and ‘as.package_version’, suitable
          character vectors as above, or numeric version objects.  For
          ‘is.numeric_version’ and ‘is.package_version’, arbitrary R
          objects.
```
So, our earlier usage was incorrect.

kartikeyakirar: makes sense.

Issue #1128, opened 2024-03-05
remove fluff in `teal.R`


chlebowa:
#### current state

```
#' @importFrom magrittr %>%
#' @export
magrittr::`%>%`

# Fix R CMD check notes
#' @import shiny teal.data teal.slice
#' @importFrom stats setNames
NULL

# Fix R CMD check notes
#' @importFrom lifecycle badge
lifecycle::badge
```

Some of this is superfluous and misleading.

#### sufficient and accurate
```
#' @import shiny teal.data teal.slice
#' @importFrom stats setNames
#' @importFrom magrittr %>%
#' @export magrittr::`%>%`
NULL
```
checks will pass

m7pr:
feel free to open a PR. looks good

PR #1214, opened 2024-05-09
1128 remove fluff in imports

m7pr:
Closes https://github.com/insightsengineering/teal/issues/1128


R/teal.R
```diff
@@ -9,15 +9,8 @@
...
 
-#' @importFrom magrittr %>%
-#' @export
-magrittr::`%>%`
-
-# Fix R CMD check notes
 #' @import shiny teal.data teal.slice
 #' @importFrom stats setNames
+#' @importFrom magrittr %>%
+#' @export %>%
```
kartikeyakirar, line 15: According to  [R Packages - Dependencies in Practice](https://r-pkgs.org/dependencies-in-practice.html#exports), the better way to re-export a method would be. And run devtools::check() to re-generate NAMESPACE and document.

```suggestion
#' @export
#' @importFrom magrittr %>%
magrittr::`%>%`
```

The @export directive requires fully qualified function names. ANd  it can not find documentation for %>% by placing them directly after the `@export` tag, as this would lead to CI errors. 

To correctly export the %>% operator, you must use the complete notation magrittr::\%>%`.

kartikeyakirar:
@m7pr I have one comment. rest of the PR looks Great !!

m7pr:
Thanks @kartikeyakirar

m7pr:
Let's see if this passes R CMD

Issue #1234, opened 2024-05-24
How quickly "should" an app start?


chlebowa:
### What is your question?

An app built with the following `data` and `modules` (no filters) takes over a minute to start*. The data is not particularly large but larger than the typical example data used in NEST packages. Pre-processing takes about 10 seconds. Since modules do not start until viewed, the (near) entirety of app startup taken up by `teal` itself. I suppose building the filter panel is one of the more time-consuming steps.

*) Measuring in a fresh session, from runnig the script that attaches all necessary packages, creates data and modules, and runs the app, until the moment the notification `Data loaded - App fully started up` appears on screen. 

My question is: at what point should one consider the app taking too long to start? Is there a way to improve startup time?

<details> <summary> 10 datasets of ~30 MB in total and 13 modules </summary>
```
> data
✅︎ verified teal_data object
<environment: 0x55a7eec656c0> [L]
Parent: <environment: package:teal.widgets>
Bindings:
• ADRS: <tibble[,49]> [L]
• ADEX: <tibble[,74]> [L]
• ADLB: <tibble[,103]> [L]
• ADAE: <tibble[,69]> [L]
• ADVS: <tibble[,75]> [L]
• ADSL: <tibble[,39]> [L]
• ADDV: <tibble[,62]> [L]
• ADCM: <tibble[,95]> [L]
• ADTTE: <tibble[,50]> [L]
• ADMH: <tibble[,75]> [L]
```

```
> sapply(ls(get_env(data)), \(x) dim(get_var(data, x))) 
     ADAE ADCM ADDV ADEX ADLB  ADMH ADRS ADSL ADTTE ADVS
[1,] 1934 9216   11  773 8400 24399 1200  400  1600 8982
[2,]   69   95   62   74  103    75   49   39    50   75
```

```
> modules
+ root
 + Summary Tables
  + Data Table
  + Cross Table
 + Data Quality
  + Variable Browser
  + Missing Data
  + Outliers
 + Exploratory Plots
  + Count Barchart
  + Distribution Plot
  + Bivariate Plot
  + Scatter Plot
  + Scatterplot Matrix
 + Statistical Analyses
  + Regression
  + Response Plot
  + Principal Component Analysis
```

</details>

Tagging @gogonzo for visibility.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [X] I agree to follow this project's Contribution Guidelines.

### Security Policy

[...]

PR #1236, opened 2024-05-29
Improve progress bars

chlebowa:
Related to #1234.

Update progress bars to improve user experience.

- `modules_datasets` receives optional `progress` argument that accepts a `Progress` object
- `ui_nested_tabs` and `ui_tabs_with_filters` receive `progress` argument that accepts a `Progress` object
- rewrote progress bar management in `srv_teal`

Added a large app for testing in `__stress_test.R`, to be removed before merging.

#### NOTE

Once the module UI is prepared, there is still a delay before the app starts, so the progress bar can be improved further. So far I was unable to identify the reason for the delay.


m7pr:
Hey @chlebowa thanks for the input! I think it's gonna be highly useful!

I allowed CI check to be executed on this PR. I see a Warning in R CMD Check - would you be able to check in your free time?

chlebowa:
Thanks @m7pr, I had forgotten to re-create documentation after removing the faulty defaults.

chlebowa:
There is one more [failure](https://github.com/insightsengineering/teal/actions/runs/9364265492/job/25779942899?pr=1236), I think this is something on your end.

m7pr:
Yeah, this is something @walkowif is investing on our end

walkowif:
The `urlchecker` failure should be gone after retriggering the checks.

R/utils.R
```diff
@@ -189,25 +189,46 @@ check_filter_datanames <- function(filters, datanames) {
 #' @param modules (`teal_modules`) object
 #' @param filters (`teal_slices`) object
 #' @param filtered_data_singleton A result of `teal_data_to_filtered_data` applied to `data`.
+#' @param progress (`Progress`) object, optional; will be filled
```
m7pr, line 192: @chlebowa what do you mean by `will be filled`?

R/utils.R
```diff
@@ -189,25 +189,46 @@ check_filter_datanames <- function(filters, datanames) {
...
+                             filtered_data_singleton = teal_data_to_filtered_data(data),
+                             progress) {
   checkmate::assert_class(data, "teal_data")
   checkmate::assert_multi_class(modules, c("teal_modules", "teal_module"))
   checkmate::assert_class(filters, "modules_teal_slices")
   checkmate::assert_r6(filtered_data_singleton, "FilteredData")
+  if (!missing(progress)) checkmate::assert_r6(progress, "Progress")
 
   if (!isTRUE(attr(filters, "module_specific"))) {
+    if (!missing(progress)) {
```
m7pr, line 208: why and when `progress` can be missing, if it's created in `[module_teal.R](https://github.com/insightsengineering/teal/pull/1236/files#diff-7b625abea8aaccebc90e0fddf962570dbeb41b6b470c7fed09892c6c88094b17)`?

R/utils.R
```diff
@@ -189,25 +189,46 @@ check_filter_datanames <- function(filters, datanames) {
...
     filtered_data_singleton$set_filter_state(slices)
+
     return(modules_structure(modules, filtered_data_singleton))
   }
 
   if (inherits(modules, "teal_module")) {
+    if (!missing(progress)) {
+      progress$inc(
+        amount = 1,
+        detail = sprintf("%.0f%%", progress$getValue() / progress$getMax() * 100)
```
m7pr, line 228: maybe instead of `"%.0f%%"` we can use `round()`? 

```r
sprintf("%s", round(progress$getValue() / progress$getMax() * 100, 2))
```

R/utils.R
chlebowa: Just in case the argument isn't provided but I guess if it's an internal function that will only ever be called here, it can be assumed that the argument is always there. I will remove the condition.

R/module_teal.R
```diff
@@ -160,18 +160,19 @@ srv_teal <- function(id, modules, teal_data_rv, filter = teal_slices()) {
       )
     }
 
-    env <- environment()
     datasets_reactive <- eventReactive(teal_data_rv(), {
-      env$progress <- shiny::Progress$new(session)
-      env$progress$set(0.25, message = "Setting data")
-
+      progress_d <- Progress$new(
```
m7pr, line 164: what if we use names `progress_data` and `progress_modules`
```suggestion
      progress_data <- Progress$new(
```

R/module_teal.R
```diff
@@ -160,18 +160,19 @@ srv_teal <- function(id, modules, teal_data_rv, filter = teal_slices()) {
...
+        max = length(unlist(module_labels(modules)))
+      )
+      progress_d$set(message = "Preparing data filtering")
+      on.exit(progress_d$close())
```
m7pr, line 168: maybe instead `on.exit` go with `shiny::onStop`?

R/utils.R
chlebowa: `sprintf` is used throughout the code base, I though it would be more consistent.

R/module_teal.R
chlebowa: Then one line will run over the 120 character limit and another `# nolint` exception will have to be added.

R/module_teal.R
chlebowa: No, you want the bar do disappear when this funciton exits, not when `shiny` stops.

R/module_tabs_with_filters.R
```diff
@@ -29,15 +30,16 @@
 NULL
 
 #' @rdname module_tabs_with_filters
-ui_tabs_with_filters <- function(id, modules, datasets, filter = teal_slices()) {
+ui_tabs_with_filters <- function(id, modules, datasets, filter = teal_slices(), progress) {
```
m7pr, line 33: what if `ui_tabs_with_filters` gets a parameter called `progress_modules` and `modules_datasets` gets parameter called `progress_data`, so that we can differentiate between various progress types used here? I see we have 2 progress objects in the type: one for data, one for modules.

R/utils.R
m7pr: Yeah. `sprintf` with `%s` not`%.0f%%`. `sprintf`  is fine : ) I meant it's argument

R/utils.R
```diff
@@ -189,25 +189,46 @@ check_filter_datanames <- function(filters, datanames) {
...
   checkmate::assert_multi_class(modules, c("teal_modules", "teal_module"))
   checkmate::assert_class(filters, "modules_teal_slices")
   checkmate::assert_r6(filtered_data_singleton, "FilteredData")
+  if (!missing(progress)) checkmate::assert_r6(progress, "Progress")
 
   if (!isTRUE(attr(filters, "module_specific"))) {
+    if (!missing(progress)) {
+      progress$inc(
+        amount = progress$getMax(),
+        detail = sprintf("%.0f%%", progress$getMax() / progress$getMax() * 100)
```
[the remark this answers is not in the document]
chlebowa: No, actually, the bar is set to max in one go. Might change `detail` to `100%` manually.

R/module_teal.R
m7pr: I think we can leave with that :)

R/module_teal.R
m7pr: got it! ignore me on this one

R/utils.R
m7pr: It's internal - let's remove, so there is less clutter in the code.

R/module_tabs_with_filters.R
chlebowa: I don't follow. These two functions are different scopes, what is the benefit of using different variable names?

R/utils.R
```diff
@@ -189,25 +189,46 @@ check_filter_datanames <- function(filters, datanames) {
...
+modules_datasets <- function(data,
+                             modules,
+                             filters,
+                             filtered_data_singleton = teal_data_to_filtered_data(data),
+                             progress) {
   checkmate::assert_class(data, "teal_data")
   checkmate::assert_multi_class(modules, c("teal_modules", "teal_module"))
   checkmate::assert_class(filters, "modules_teal_slices")
   checkmate::assert_r6(filtered_data_singleton, "FilteredData")
+  if (!missing(progress)) checkmate::assert_r6(progress, "Progress")
```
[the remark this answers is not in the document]
m7pr: Then I guess this can also be simplified to `checkmate::assert_r6(progress, "Progress")` based on this conversation https://github.com/insightsengineering/teal/pull/1236#discussion_r1627618621

m7pr:
Thank you @chlebowa for the time and energy you invested to provide this feature. I just test in on the branched cloned from your fork. Works like a charm! I think we are good to merge this excellent feature, once we figure out the details from the comments.

@kumamiao would you mind providing a NEWS entry for this feature?

https://github.com/insightsengineering/teal/assets/133694481/ca98b4af-f994-48c7-a796-f37de1ad49da

R/utils.R
chlebowa: Ok then.

R/utils.R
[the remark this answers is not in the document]
chlebowa: Argument is now mandatory, which means `NULL` is not allowed.

R/utils.R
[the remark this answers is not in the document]
m7pr: So it's gonna display
- Progress 0%,
- Progress 100%
and then the progress info will disappear or will be overwritten by another progress ?

R/utils.R
chlebowa: The progress bar will be filled during the (possibly recursive) funciton call.

R/module_teal.R
chlebowa: done

R/utils.R
chlebowa: done

chlebowa: Ah, I remember now. If the argument is mandatory, unit tests fail. Would you like me to include a `Progress` object in the tests?

R/utils.R
chlebowa: And here is why I didn't:
`Error in `initialize(...)`: Can only use Progress$new() inside a Shiny app`

R/utils.R
chlebowa: And here is why I didn't:
`Error in `initialize(...)`: Can only use Progress$new() inside a Shiny app`

R/utils.R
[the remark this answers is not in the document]
m7pr: Cool. Maybe we can put a comment next to this line, that there is only one go

R/utils.R
m7pr: hmm, so maybe `(Progress) an optional progress object to be filled`

chlebowa: I expanded the argument description to 
```
#' @param progress (`Progress`) object from `shiny`, optional.
#'  The progress bar will be filled during the (possibly recursive) call.
```

R/module_tabs_with_filters.R
m7pr: My point was that it's hard to understand if those are the same progresses used across the package if you look for `progress` term with a global search/find

image

m7pr: It's not a big deal, I'm fine having the same name as it's shorter and within the scope of one function is not confusing.

R/utils.R
m7pr: amazing! thanks

R/utils.R
m7pr: coolio

R/module_tabs_with_filters.R
chlebowa: Would you do the same for `filter` ,`modules`, `datasets` or `datanames`?

chlebowa:
I had a thought this morning that may be a slightly more orderly implementation. Rather than creating tro `Progress` objects, one in a reactive expression and one in an observer, instantiate one `Progress` in `srv_teal` and work with it throughout the function by first passing it to `modules_datasets` and later reset the progress and pass to `ui_tabs_with_filters`. 

The user would see only one progress bar, whereas now two bars can be seen, the filter one is moved up and later disappears as the module one appears. On the code seide, there would be a little less clutter.

What do you think?

m7pr:
If one waits for the other, then it's probably better for the end user to see just one progress bar at a time :)
Depending on how much time you have. This is already in good condition

R/module_tabs_with_filters.R
m7pr: Forget about it. I would not do that for other modules like modules, filters or datasets

chlebowa:
Have a look.

m7pr:
Wow @chlebowa nice and clean. Really ready to be merged!

Can you 
- delete `__stres_test.R` file
- extend NEWS.md with an Enhancement `* Provided progress bar for modules loading and data filtering during teal app startup.`

Then I will merge. Thank you. Really great job!

R/module_teal.R
m7pr: @chlebowa can we get back to progress_data from the initial review?

chlebowa: Yup, sorry about that.

m7pr (approved):
really useful stuff!

pawelru:
Hi all. Apologies for jumping into the old PR - just got back from the break of mine. May I ask for a follow-up PR before the release that will make `progress` argument optional for all modified functions? The rationale here is simplicity - we would like to keep the requirements to run the funcs to the meaningful minimum which is in these case: `id`, `modules` and `datasets`. If you look at the functions definitions all the other arguments (incl. `filter` which is also quite important) has some defaults. In case of `progress` this could be `NULL` and if it's `Progress` we will increment it. Another benefits are in testing - no need to create a dummy progress object to execute functions.

(*) I know the `is.missing()` stuff but IMHO having `x = NULL` in the function definition exposes this better without reading docs or function body

m7pr:
@pawelru I created a ticket for that in here https://github.com/insightsengineering/teal/issues/1243

PR #1237, opened 2024-05-29
Fix the integration tests by alternative 

vedhav:
Closes https://github.com/insightsengineering/coredev-tasks/issues/554

Changes:

1. ~The container in the integration test does not seem to evaluate the JS code to check visibility when `app_driver$is_visible` is called with `visibility_property = TRUE`~ Bumping the chrome version to resolve this issue.
2. Increasing the timeout threshold for `wait_for_page_stability`. For an unknown reason this timeout does not work in certain containers. Initial timeout (500ms) was set as a 2x of what was needed in my local machine. But looks like we need at least 800 ms to work in some loaded conditions. So, setting it to 2000 ms now.

### DIFF
--- a/R/TealAppDriver.R
+++ b/R/TealAppDriver.R
@@ -338,36 +338,18 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
         "Element.prototype.checkVisibility is not supported in the current browser."
       )
 
-      if (visibility_property) {
-        visibility <- unlist(
-          self$get_js(
-            sprintf(
-              "var selector = '%s';
-              var elements = document.querySelectorAll(selector);
-              var visibilityList = [];
-              elements.forEach(function(element) {
-                var computedStyles = window.getComputedStyle(element);
-                visibilityList.push(computedStyles.visibility !== 'hidden');
-              });
-              visibilityList;",
-              selector
-            )
+      unlist(
+        self$get_js(
+          sprintf(
+            "Array.from(document.querySelectorAll('%s')).map(el => el.checkVisibility({%s, %s, %s}))",
+            selector,
+            # Extra parameters
+            sprintf("contentVisibilityAuto: %s", tolower(content_visibility_auto)),
+            sprintf("opacityProperty: %s", tolower(opacity_property)),
+            sprintf("visibilityProperty: %s", tolower(visibility_property))
           )
         )
-      } else {
-        visibility <- unlist(
-          self$get_js(
-            sprintf(
-              "Array.from(document.querySelectorAll('%s')).map(el => el.checkVisibility({%s, %s}))",
-              selector,
-              # Extra parameters
-              sprintf("contentVisibilityAuto: %s", tolower(content_visibility_auto)),
-              sprintf("opacityProperty: %s", tolower(opacity_property))
-            )
-          )
-        )
-      }
-      visibility
+      )
     },
     #' @description
     #' Get the active filter variables from a dataset in the `teal` app.
--- a/R/TealAppDriver.R
+++ b/R/TealAppDriver.R
@@ -76,7 +76,7 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
         strict = FALSE
       )
 
-      required_version <- "105"
+      required_version <- "121"
 
       testthat::skip_if(
         is.na(chrome_version),
--- a/R/TealAppDriver.R
+++ b/R/TealAppDriver.R
@@ -672,7 +672,7 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
     # @param stability_period (`numeric(1)`) The time in milliseconds to wait till the page to be stable.
     # @param check_interval (`numeric(1)`) The time in milliseconds to check for changes in the page.
     # The stability check is reset when a change is detected in the page after sleeping for check_interval.
-    wait_for_page_stability = function(stability_period = 500, check_interval = 50) {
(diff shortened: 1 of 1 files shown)


vedhav:
@averissimo would appreciate your input with 1.
@kartikeyakirar added you as you were involved in 2.

R/TealAppDriver.R
```diff
@@ -338,18 +338,36 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
+              "var selector = '%s';
+              var elements = document.querySelectorAll(selector);
+              var visibilityList = [];
+              elements.forEach(function(element) {
+                var computedStyles = window.getComputedStyle(element);
+                visibilityList.push(computedStyles.visibility !== 'hidden');
+              });
+              visibilityList;",
+              selector
+            )
```
averissimo, line 354: I think we should use an alternative approach just as we chatted on slack:
- Force minimum version to be `121` and keep old code (just changing that versioning requirement)
- Use `105` parameters instead and remove `content_visibility_auto` parameter as that is from a newer API.
  - See diff below

```diff
diff --git a/R/TealAppDriver.R b/R/TealAppDriver.R
index bad441567..e666ecf1f 100644
--- a/R/TealAppDriver.R
+++ b/R/TealAppDriver.R
@@ -323,7 +323,6 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
     #'
     #' @return Logical vector with all occurrences of the selector.
     is_visible = function(selector,
-                          content_visibility_auto = FALSE,
                           opacity_property = FALSE,
                           visibility_property = FALSE) {
       checkmate::assert_string(selector)
@@ -338,36 +337,18 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
         "Element.prototype.checkVisibility is not supported in the current browser."
       )
 
-      if (visibility_property) {
-        visibility <- unlist(
-          self$get_js(
-            sprintf(
-              "var selector = '%s';
-              var elements = document.querySelectorAll(selector);
-              var visibilityList = [];
-              elements.forEach(function(element) {
-                var computedStyles = window.getComputedStyle(element);
-                visibilityList.push(computedStyles.visibility !== 'hidden');
-              });
-              visibilityList;",
-              selector
-            )
-          )
-        )
-      } else {
-        visibility <- unlist(
-          self$get_js(
-            sprintf(
-              "Array.from(document.querySelectorAll('%s')).map(el => el.checkVisibility({%s, %s}))",
-              selector,
-              # Extra parameters
-              sprintf("contentVisibilityAuto: %s", tolower(content_visibility_auto)),
-              sprintf("opacityProperty: %s", tolower(opacity_property))
-            )
+      unlist(
+        self$get_js(
+          sprintf(
+            "Array.from(document.querySelectorAll('%s')).map(el => el.checkVisibility({%s, %s}))",
+            selector,
+            # Extra parameters
+            # Add support for contentVisibilityAuto when minimum version is at 121
+            sprintf("checkOpacity: %s", tolower(opacity_property)),
+            sprintf("checkVisibilityCSS: %s", tolower(visibility_property))
           )
         )
-      }
-      visibility
+      )
     },
     #' @description
     #' Get the active filter variables from a dataset in the `teal` app.
```

averissimo: Here's the diff for the first bullet point:

It basically reverts the changes and changes `required_version <- "121"`

<details>

<summary>Diff</summary>

```diff
diff --git a/R/TealAppDriver.R b/R/TealAppDriver.R
index bad441567..acd950ae6 100644
--- a/R/TealAppDriver.R
+++ b/R/TealAppDriver.R
@@ -76,7 +76,7 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
         strict = FALSE
       )
 
-      required_version <- "105"
+      required_version <- "121"
 
       testthat::skip_if(
         is.na(chrome_version),
@@ -338,36 +338,18 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
         "Element.prototype.checkVisibility is not supported in the current browser."
       )
 
-      if (visibility_property) {
-        visibility <- unlist(
-          self$get_js(
-            sprintf(
-              "var selector = '%s';
-              var elements = document.querySelectorAll(selector);
-              var visibilityList = [];
-              elements.forEach(function(element) {
-                var computedStyles = window.getComputedStyle(element);
-                visibilityList.push(computedStyles.visibility !== 'hidden');
-              });
-              visibilityList;",
-              selector
-            )
-          )
-        )
-      } else {
-        visibility <- unlist(
-          self$get_js(
-            sprintf(
-              "Array.from(document.querySelectorAll('%s')).map(el => el.checkVisibility({%s, %s}))",
-              selector,
-              # Extra parameters
-              sprintf("contentVisibilityAuto: %s", tolower(content_visibility_auto)),
-              sprintf("opacityProperty: %s", tolower(opacity_property))
-            )
+      unlist(
+        self$get_js(
+          sprintf(
+            "Array.from(document.querySelectorAll('%s')).map(el => el.checkVisibility({%s, %s, %s}))",
+            selector,
+            # Extra parameters
+            sprintf("contentVisibilityAuto: %s", tolower(content_visibility_auto)),
+            sprintf("opacityProperty: %s", tolower(opacity_property)),
+            sprintf("visibilityProperty: %s", tolower(visibility_property))
           )
         )
-      }
-      visibility
+      )
     },
     #' @description
     #' Get the active filter variables from a dataset in the `teal` app.

```

</details>

vedhav: Now, I will make sure to check the browser compatibility with every JS change we add.

donyunardi (approved):

PR #1252, opened 2024-06-26
Update bootstrap-themes-in-teal.Rmd

Alexandra01:
This update is crucial as it was needed in the past, but the fix has already been released, hence adding the requirement for the minimum version rather than downloading from main.
The related fix was released with shinyWidgets 0.7.4: https://github.com/dreamRs/shinyWidgets/commit/d82d8ad287c952cb37c8f5ae3e7db134c37314b8


Polkas:
I confirm that the related issue was merged in shinyWidgets 0.7.4

donyunardi (approved):
Thanks for improving the docs @Alexandra01 !

PR #1254, opened 2024-06-28
Update teal-options.Rmd

pawelru:
Add info to set the option before loading packages

FYI: @chlebowa

### DIFF
--- a/vignettes/teal-options.Rmd
+++ b/vignettes/teal-options.Rmd
@@ -66,14 +66,14 @@ This option controls the dots per inch of the graphs rendered and downloaded whe
 Default: 72
 
 ### `teal.log_layout` (`character`)
-This defines the layout of a log message used in a `teal` application. `teal` uses this layout to format the emitted log messages. Read the documentation of `teal.logger::register_logger` for more information. This must be set before loading any `teal` package(s).
+This defines the layout of a log message used in a `teal` application. `teal` uses this layout to format the emitted log messages. Read the documentation of `teal.logger::register_logger` for more information. This must be set before loading any `teal` package(s). In case the option is set after attaching the packages, please re-attach the packages again to use it correctly.
 
 Default: `"[{level}] {format(time, \"%Y-%m-%d %H:%M:%OS4\")} pid:{pid} token:[{token}] {ans} {msg}"`.
 
 Note that this layout is formatted by the `glue` package.
 
 ### `teal.log_level` (`character`)
-This is the logging level threshold used in a `teal` application. A `teal` application will not emit logs below this level. See the documentation of `logger::TRACE` for all possible values of logging threshold and more information on what it does. This must be set before loading any `teal` package(s).
+This is the logging level threshold used in a `teal` application. A `teal` application will not emit logs below this level. See the documentation of `logger::TRACE` for all possible values of logging threshold and more information on what it does. This must be set before loading any `teal` package(s). In case the option is set after attaching the packages, please re-attach the packages again to use it correctly.
 
 Default: `"INFO"`.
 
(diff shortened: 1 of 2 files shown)


chlebowa:
Thank you 👍

vignettes/teal-options.Rmd
```diff
@@ -66,14 +66,14 @@ This option controls the dots per inch of the graphs rendered and downloaded whe
 Default: 72
 
 ### `teal.log_layout` (`character`)
-This defines the layout of a log message used in a `teal` application. `teal` uses this layout to format the emitted log messages. Read the documentation of `teal.logger::register_logger` for more information.
+This defines the layout of a log message used in a `teal` application. `teal` uses this layout to format the emitted log messages. Read the documentation of `teal.logger::register_logger` for more information. This must be set before loading any `teal` package(s).
```
m7pr, line 69: @pawelru I would rather say that:

If you set those options, you need to load `teal` package again if it was already loaded.

pawelru: Hmmm... How would you phrase it?

I'm thinking about the use cases, i.e. how this will used. To me, most of the people will be creating some kind of project structure or even a simple `app.R` file and I want them to instruct how to order stuff. This is slightly different to how we are working, that is: interactive session with a lots of load/attach on-the-fly. My point is that in that project structure it's quite uncommon (or I would even say: not recommended) to reload packages. WDYT?

chlebowa: > If you set those options, you need to load `teal` package again if it was already loaded.

That can't be done without detaching, can it? `library` exits early if the package is already attache, I suppose it is the same for loading.

m7pr: `Recommended to set before loading teal package(s). If packages were loaded first, and the option was set after, please reload the package to enable the option.`

?

pawelru: > > If you set those options, you need to load `teal` package again if it was already loaded.
> 
> That can't be done without detaching, can it? `library` exits early if the package is already attache, I suppose it is the same for loading.

Yes of course. This would have to be detaching and attached again.

Coming back to my previous thought - I actually don't really mind. The documentation is also for developers and they are oftentimes working in a "more interactive" way. I think I will apply your suggestion

pawelru, line 69:
```suggestion
This defines the layout of a log message used in a `teal` application. `teal` uses this layout to format the emitted log messages. Read the documentation of `teal.logger::register_logger` for more information. This must be set before loading any `teal` package(s). In case the option is set after attaching the packages, please re-attach the packages again to use it correctly.
```

vignettes/teal-options.Rmd
```diff
@@ -66,14 +66,14 @@ This option controls the dots per inch of the graphs rendered and downloaded whe
...
 
 Default: `"[{level}] {format(time, \"%Y-%m-%d %H:%M:%OS4\")} pid:{pid} token:[{token}] {ans} {msg}"`.
 
 Note that this layout is formatted by the `glue` package.
 
 ### `teal.log_level` (`character`)
-This is the logging level threshold used in a `teal` application. A `teal` application will not emit logs below this level. Read the documentation of `teal.logger::register_logger` for more information. Possible values: `"TRACE"`, `"INFO"`, `"WARNING"`, `"ERROR"`. See the documentation of `logger::TRACE` for all possible values of logging threshold and more information on what it does.
+This is the logging level threshold used in a `teal` application. A `teal` application will not emit logs below this level. See the documentation of `logger::TRACE` for all possible values of logging threshold and more information on what it does. This must be set before loading any `teal` package(s).
```
pawelru, line 76:
```suggestion
This is the logging level threshold used in a `teal` application. A `teal` application will not emit logs below this level. See the documentation of `logger::TRACE` for all possible values of logging threshold and more information on what it does. This must be set before loading any `teal` package(s). In case the option is set after attaching the packages, please re-attach the packages again to use it correctly.
```

m7pr (approved):

PR #1232, opened 2024-05-22
479 add `renv::snapshot()` for `.lockfile` with `future` + `shiny::ExtendedTask`

m7pr:
Fixes https://github.com/insightsengineering/coredev-tasks/issues/479
Alternative for https://github.com/insightsengineering/teal/pull/1224

This PR includes new imports
- `{renv}` 
    - so that we can use a function that creates a `.lockfile` for future reproducibility
- `{future}` 
    - so that we can run `renv::snapshot()` as a `shiny::ExtendedTask()$new` in a parallel process to the shiny session

Tested with

```r
# pass your own .lockfile
# options(teal.renv.lockfile = "session.lock")

# allow to create the .lockfile based on DESCRIPTION file
# renv::settings$snapshot.type("explicit")

app <- init(
  data = teal_data(iris = iris),
  modules = example_module(label = "example teal module")
)
if (interactive()) {
  shinyApp(app$ui, app$server)
}
```

PR #1255, opened 2024-07-01
substitute `future::promises` lockfile creation with `callr::r_bg`

m7pr:
An alternative for https://github.com/insightsengineering/teal/pull/1232 that does not require `future` nor `promises`, hence does not require handling of the parallel computations plan.

### DIFF
--- a/R/init.R
+++ b/R/init.R
@@ -231,6 +231,8 @@ init <- function(data,
       if (!is.null(landing_module)) {
         do.call(landing_module$server, c(list(id = "landing_module_shiny_id"), landing_module$server_args))
       }
+      session$userData$lockfile_process <- process
+      reactivePoll(1000, session, checkFunc = lockfile_status, valueFunc = function(){})
       srv_teal_with_splash(id = id, data = data, modules = modules, filter = deep_copy_filter(filter))
     }
   )
--- a/R/init.R
+++ b/R/init.R
@@ -154,7 +154,7 @@ init <- function(data,
   teal.logger::log_system_info()
 
   # invoke lockfile creation
-  teal_lockfile()
+  process <- teal_lockfile()
 
   # argument transformations
   ## `modules` - landing module
--- a/R/init.R
+++ b/R/init.R
@@ -232,7 +232,7 @@ init <- function(data,
         do.call(landing_module$server, c(list(id = "landing_module_shiny_id"), landing_module$server_args))
       }
       session$userData$lockfile_process <- process
-      reactivePoll(1000, session, checkFunc = lockfile_status, valueFunc = function(){})
+      reactivePoll(1000, session, checkFunc = lockfile_status, valueFunc = function() {})
       srv_teal_with_splash(id = id, data = data, modules = modules, filter = deep_copy_filter(filter))
     }
   )
--- a/R/init.R
+++ b/R/init.R
@@ -231,8 +231,7 @@ init <- function(data,
       if (!is.null(landing_module)) {
         do.call(landing_module$server, c(list(id = "landing_module_shiny_id"), landing_module$server_args))
       }
-      session$userData$lockfile_process <- process
-      reactivePoll(1000, session, checkFunc = lockfile_status, valueFunc = function() {})
+      lockfile_status_tracker(process)
       srv_teal_with_splash(id = id, data = data, modules = modules, filter = deep_copy_filter(filter))
     }
   )
--- a/R/teal_lockfile.R
+++ b/R/teal_lockfile.R
@@ -43,8 +43,8 @@ teal_lockfile <- function() {
 
   # # capture.output is not needed if stdout = "|"
   # utils::capture.output( # Needed to suppress: 'Opening fd 1' message
-  callr::r_bg(
-    func = create_renv_lockfile_2,
+  process <- callr::r_bg(
+    func = create_renv_lockfile,
     args = list(
       lockfile_path = lockfile_path,
       opts = options()
@@ -61,13 +61,7 @@ teal_lockfile <- function() {
   # )
 
   logger::log_trace("lockfile creation started.")
-}
-
-create_renv_lockfile_2 <- function(lockfile_path = NULL, opts = NULL) {
-  lapply(1:15, function(x) {
-    print(paste0('teal:', x)))
-    Sys.sleep(1)
-  })
+  process
 }
 
 create_renv_lockfile <- function(lockfile_path = NULL, opts = NULL) {
--- a/R/teal_lockfile.R
+++ b/R/teal_lockfile.R
@@ -47,7 +47,7 @@ teal_lockfile <- function() {
 create_renv_lockfile <- function(lockfile_path = NULL) {
   checkmate::assert_string(lockfile_path, na.ok = TRUE)
 
-  renv_logs <- utils::capture.output(
+  utils::capture.output(
     renv::snapshot(
       lockfile = lockfile_path,
       prompt = FALSE,
--- a/R/teal_lockfile.R
+++ b/R/teal_lockfile.R
@@ -40,17 +40,30 @@ teal_lockfile <- function() {
   }
 
   shiny::onStop(function() file.remove(lockfile_path))
-  callr::r_bg(
-    func = create_renv_lockfile,
-    args = list(
-      lockfile_path = lockfile_path,
-      opts = options()
+
+  # Both capture.output are not needed if stdout = "|"
+  callr_output <-
+  capture.output( # Needed to suppress: 'Opening fd 1' message
+    capture.output( # Needed to suppress: 'PROCESS 'Rterm', running, pid' output
+      callr::r_bg(
+        func = create_renv_lockfile,
+        args = list(
+          lockfile_path = lockfile_path,
+          opts = options()
+        ),
+        # print results to the console
+        stdout = "",
+        # renv setup is orchestrated by its special S3 objects: renv::settings, renv::config and renv::paths
+        #     `package` parameter include `renv` namespace inside the environment of `func = create_renv_lockfile` fun
+        package = "renv"
+        # default env = NULL # means that callr process will use environmental variables
+        #     from the main session (parent process)
+      ),
+      type = "message"
     ),
-    # renv setup is orchestrated by its special S3 objects: renv::settings, renv::config and renv::paths
-    #     `package` parameter include `renv` namespace inside the environment of `func = create_renv_lockfile` function
-    package = "renv"
-    # default env = NULL # means that callr process will use environmental variables from the main session (parent process)
+    type = "output"
   )
+
   logger::log_trace("lockfile creation started.")
 }
 
@@ -60,13 +73,20 @@ create_renv_lockfile <- function(lockfile_path = NULL, opts) {
 
   options(opts)
 
-  renv::snapshot(
-    lockfile = lockfile_path,
-    prompt = FALSE,
-    force = TRUE
-    # type = is taken from renv::settings$snapshot.type()
+  renv_logs <- utils::capture.output(
+    renv::snapshot(
+      lockfile = lockfile_path,
+      prompt = FALSE,
+      force = TRUE
+      # type = is taken from renv::settings$snapshot.type()
+    )
   )
-  lockfile_path
+  # print() is required for callr::r_bg to track console output
+  if (any(grepl("Lockfile written", renv_logs))) {
+    print("lockfile created successfully.")
+  } else {
+    print("lockfile created with issues.")
+  }
 }
 
 teal_lockfile_downloadhandler <- function() {
--- a/R/teal_lockfile.R
+++ b/R/teal_lockfile.R
@@ -40,12 +40,15 @@ teal_lockfile <- function() {
   }
 
   shiny::onStop(function() file.remove(lockfile_path))
-  callr::r_bg(create_renv_lockfile, args = list(lockfile_path = lockfile_path))
+  callr::r_bg(create_renv_lockfile, args = list(lockfile_path = lockfile_path, opts = options()), package = 'renv')
   logger::log_trace("lockfile creation started.")
 }
 
-create_renv_lockfile <- function(lockfile_path = NULL) {
+create_renv_lockfile <- function(lockfile_path = NULL, opts) {
   checkmate::assert_string(lockfile_path, na.ok = TRUE)
+  checkmate::assert_list(opts)
+
+  options(opts)
 
   renv::snapshot(
     lockfile = lockfile_path,
--- a/R/teal_lockfile.R
+++ b/R/teal_lockfile.R
@@ -41,28 +41,24 @@ teal_lockfile <- function() {
 
   shiny::onStop(function() file.remove(lockfile_path))
 
-  # Both capture.output are not needed if stdout = "|"
-  callr_output <-
-    utils::capture.output( # Needed to suppress: 'Opening fd 1' message
-      utils::capture.output( # Needed to suppress: 'PROCESS 'Rterm', running, pid' output
-        callr::r_bg(
-          func = create_renv_lockfile,
-          args = list(
-            lockfile_path = lockfile_path,
-            opts = options()
-          ),
-          # print results to the console
-          stdout = "",
-          # renv setup is orchestrated by its special S3 objects: renv::settings, renv::config and renv::paths
-          #     `package` parameter include `renv` namespace inside the environment of `func = create_renv_lockfile` fun
-          package = "renv"
-          # default env = NULL # means that callr process will use environmental variables
-          #     from the main session (parent process)
-        ),
-        type = "message"
+  # capture.output is not needed if stdout = "|"
+  utils::capture.output( # Needed to suppress: 'Opening fd 1' message
+    callr::r_bg(
+      func = create_renv_lockfile,
+      args = list(
+        lockfile_path = lockfile_path,
+        opts = options()
       ),
-      type = "output"
-    )
+      # print results to the console
+      stdout = "",
+      # renv setup is orchestrated by its special S3 objects: renv::settings, renv::config and renv::paths
+      #     `package` parameter include `renv` namespace inside the environment of `func = create_renv_lockfile` fun
+      package = "renv"
+      # default env = NULL # means that callr process will use environmental variables
+      #     from the main session (parent process)
+    ),
+    type = "message"
+  )
 
   logger::log_trace("lockfile creation started.")
 }
--- a/R/teal_lockfile.R
+++ b/R/teal_lockfile.R
@@ -49,13 +49,13 @@ teal_lockfile <- function() {
         lockfile_path = lockfile_path,
         opts = options()
       ),
-      # print results to the console
+      # prints results to the console
       stdout = "",
       # renv setup is orchestrated by its special S3 objects: renv::settings, renv::config and renv::paths
-      #     `package` parameter include `renv` namespace inside the environment of `func = create_renv_lockfile` fun
-      package = "renv"
-      # default env = NULL # means that callr process will use environmental variables
-      #     from the main session (parent process)
+      #     `package` parameter include `renv` namespace inside the environment of `func = create_renv_lockfile`
+      package = "renv",
+      # callr process will use environmental variables from the main session (parent process)
+      env = NULL
     ),
     type = "message"
   )
--- a/R/teal_lockfile.R
+++ b/R/teal_lockfile.R
@@ -43,21 +43,21 @@ teal_lockfile <- function() {
 
   # Both capture.output are not needed if stdout = "|"
   callr_output <-
-  utils::capture.output( # Needed to suppress: 'Opening fd 1' message
-    utils::capture.output( # Needed to suppress: 'PROCESS 'Rterm', running, pid' output
-      callr::r_bg(
-        func = create_renv_lockfile,
-        args = list(
-          lockfile_path = lockfile_path,
-          opts = options()
-        ),
-        # print results to the console
-        stdout = "",
-        # renv setup is orchestrated by its special S3 objects: renv::settings, renv::config and renv::paths
-        #     `package` parameter include `renv` namespace inside the environment of `func = create_renv_lockfile` fun
-        package = "renv"
-        # default env = NULL # means that callr process will use environmental variables
-        #     from the main session (parent process)
+    utils::capture.output( # Needed to suppress: 'Opening fd 1' message
+      utils::capture.output( # Needed to suppress: 'PROCESS 'Rterm', running, pid' output
+        callr::r_bg(
+          func = create_renv_lockfile,
+          args = list(
+            lockfile_path = lockfile_path,
+            opts = options()
+          ),
+          # print results to the console
+          stdout = "",
+          # renv setup is orchestrated by its special S3 objects: renv::settings, renv::config and renv::paths
+          #     `package` parameter include `renv` namespace inside the environment of `func = create_renv_lockfile` fun
+          package = "renv"
+          # default env = NULL # means that callr process will use environmental variables
+          #     from the main session (parent process)
         ),
         type = "message"
       ),
--- a/R/teal_lockfile.R
+++ b/R/teal_lockfile.R
@@ -64,7 +64,7 @@ teal_lockfile <- function() {
 }
 
 create_renv_lockfile_2 <- function(lockfile_path = NULL, opts = NULL) {
-  lapply(1:15, function(x) {print(x); Sys.sleep(1)})
+  lapply(1:15, function(x) {print(paste0('teal:', x)); Sys.sleep(1)})
 }
 
 create_renv_lockfile <- function(lockfile_path = NULL, opts = NULL) {
--- a/R/teal_lockfile.R
+++ b/R/teal_lockfile.R
@@ -47,13 +47,11 @@ teal_lockfile <- function() {
 create_renv_lockfile <- function(lockfile_path = NULL) {
   checkmate::assert_string(lockfile_path, na.ok = TRUE)
 
-  utils::capture.output(
-    renv::snapshot(
-      lockfile = lockfile_path,
-      prompt = FALSE,
-      force = TRUE
-      # type = is taken from renv::settings$snapshot.type()
-    )
+  renv::snapshot(
+    lockfile = lockfile_path,
+    prompt = FALSE,
+    force = TRUE
+    # type = is taken from renv::settings$snapshot.type()
   )
   lockfile_path
 }
--- a/R/teal_lockfile.R
+++ b/R/teal_lockfile.R
@@ -40,7 +40,17 @@ teal_lockfile <- function() {
   }
 
   shiny::onStop(function() file.remove(lockfile_path))
-  callr::r_bg(create_renv_lockfile, args = list(lockfile_path = lockfile_path, opts = options()), package = "renv")
+  callr::r_bg(
+    func = create_renv_lockfile, 
+    args = list(
+      lockfile_path = lockfile_path, 
+      opts = options()
+    ), 
+    # renv setup is orchestrated by its special S3 objects: renv::settings, renv::config and renv::paths
+    #     `package` parameter include `renv` namespace inside the environment of `func = create_renv_lockfile` function
+    package = "renv" 
+    # default env = NULL # means that callr process will use environmental variables from the main session (parent process)
+  )  
   logger::log_trace("lockfile creation started.")
 }
 
--- a/R/teal_lockfile.R
+++ b/R/teal_lockfile.R
@@ -39,52 +39,23 @@ teal_lockfile <- function() {
     }
   }
 
-  if (!(is_in_test() || is_r_cmd_check())) {
-    # old_plan <- future::plan()
-    # # If there is already a parallel (non-sequential) backend, reuse it.
-    # if (inherits(old_plan, "sequential")) {
-    #   future::plan(future::multisession, workers = 2)
-    # }
-    #
-    # lockfile_task <- ExtendedTask$new(create_renv_lockfile)
-    # lockfile_task$invoke(close = inherits(old_plan, "sequential"), lockfile_path)
-    callr::r_bg(create_renv_lockfile, args = list(lockfile_path = lockfile_path))
-    logger::log_trace("lockfile creation invoked.")
-  }
+  shiny::onStop(function() file.remove(lockfile_path))
+  callr::r_bg(create_renv_lockfile, args = list(lockfile_path = lockfile_path))
+  logger::log_trace("lockfile creation started.")
 }
 
-create_renv_lockfile <- function(#close = FALSE,
-                                 lockfile_path = NULL) {
-  #checkmate::assert_flag(close)
+create_renv_lockfile <- function(lockfile_path = NULL) {
   checkmate::assert_string(lockfile_path, na.ok = TRUE)
-  #promise <- promises::future_promise({
-    # Below we can not use a file created in tempdir() directory.
-    # If a file is created in tempdir(), it gets deleted on 'then(onFulfilled' part.
-    shiny::onStop(function() file.remove(lockfile_path))
 
-    renv_logs <- utils::capture.output(
-      renv::snapshot(
-        lockfile = lockfile_path,
-        prompt = FALSE,
-        force = TRUE
-        # type = is taken from renv::settings$snapshot.type()
-      )
+  renv_logs <- utils::capture.output(
+    renv::snapshot(
+      lockfile = lockfile_path,
+      prompt = FALSE,
+      force = TRUE
+      # type = is taken from renv::settings$snapshot.type()
     )
-    if (any(grepl("Lockfile written", renv_logs))) {
-      logger::log_trace("lockfile created successfully.")
-    } else {
-      logger::log_trace("lockfile created with issues.")
-    }
-
-    lockfile_path
-  #})
-  # if (close) {
-  #   # If the initial backend was only sequential, bring it back.
-  #   promises::then(promise, onFulfilled = function() {
-  #     future::plan(future::sequential)
-  #   })
-  # }
-  # promise
+  )
+  lockfile_path
 }
 
 teal_lockfile_downloadhandler <- function() {
@@ -105,12 +76,4 @@ teal_lockfile_downloadhandler <- function() {
     },
     contentType = "application/json"
   )
-}
-
-is_r_cmd_check <- function() {
-  ("CheckExEnv" %in% search()) || any(c("_R_CHECK_TIMINGS_", "_R_CHECK_LICENSE_") %in% names(Sys.getenv()))
-}
-
-is_in_test <- function() {
-  identical(Sys.getenv("TESTTHAT"), "true")
 }
--- a/R/teal_lockfile.R
+++ b/R/teal_lockfile.R
@@ -41,24 +41,18 @@ teal_lockfile <- function() {
 
   shiny::onStop(function() file.remove(lockfile_path))
 
-  # # capture.output is not needed if stdout = "|"
-  # utils::capture.output( # Needed to suppress: 'Opening fd 1' message
-  process <- callr::r_bg(
+  process <- callr::r_bg( # callr process needs to be assigned to an object so does shiny doesn't kill it on shinyApp
     func = create_renv_lockfile,
     args = list(
       lockfile_path = lockfile_path,
       opts = options()
     ),
-    # prints results to the console
-    stdout = "",
     # renv setup is orchestrated by its special S3 objects: renv::settings, renv::config and renv::paths
     #     `package` parameter include `renv` namespace inside the environment of `func = create_renv_lockfile`
     package = "renv",
     # callr process will use environmental variables from the main session (parent process)
     env = Sys.getenv()
-  ) # ,
-  #   type = "message"
-  # )
+  )
 
   logger::log_trace("lockfile creation started.")
   process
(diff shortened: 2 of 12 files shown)


R/teal_lockfile.R
```diff
@@ -39,50 +39,23 @@ teal_lockfile <- function() {
     }
   }
 
-  if (!(is_in_test() || is_r_cmd_check())) {
```
m7pr, line 42: `is_in_test()` and `is_r_cmd_check()` no longer needed as there is no parallel cluster socket process that gets started with `future` plan

R/teal_lockfile.R
```diff
@@ -39,50 +39,23 @@ teal_lockfile <- function() {
...
+      lockfile = lockfile_path,
+      prompt = FALSE,
+      force = TRUE
+      # type = is taken from renv::settings$snapshot.type()
     )
-    if (any(grepl("Lockfile written", renv_logs))) {
-      logger::log_trace("lockfile created successfully.")
-    } else {
-      logger::log_trace("lockfile created with issues.")
-    }
```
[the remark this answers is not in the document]
gogonzo: How would you implement this on a feature branch? I don't see immediately  how this could be done.

R/teal_lockfile.R
```diff
@@ -39,50 +39,23 @@ teal_lockfile <- function() {
...
-    shiny::onStop(function() file.remove(lockfile_path))
 
-    renv_logs <- utils::capture.output(
-      renv::snapshot(
-        lockfile = lockfile_path,
-        prompt = FALSE,
-        force = TRUE
-        # type = is taken from renv::settings$snapshot.type()
-      )
+  renv_logs <- utils::capture.output(
```
pawelru, line 50: do you still need `utils::capture()` you are not passing `renv_logs` variable...

R/teal_lockfile.R
```diff
@@ -39,50 +39,23 @@ teal_lockfile <- function() {
...
-    if (inherits(old_plan, "sequential")) {
-      future::plan(future::multisession, workers = 2)
-    }
-
-    lockfile_task <- ExtendedTask$new(create_renv_lockfile)
-    lockfile_task$invoke(close = inherits(old_plan, "sequential"), lockfile_path)
-    logger::log_trace("lockfile creation invoked.")
-  }
+  shiny::onStop(function() file.remove(lockfile_path))
+  callr::r_bg(create_renv_lockfile, args = list(lockfile_path = lockfile_path))
```
pawelru, line 43: are you able to pass R options? This might be relevant for the lockfile creation

m7pr: For sure `.Renvrion` is shared between R and `callr` sessions. Need to double check what happens with `options` or `env` (environment variables).

pawelru: I was more referring to the options set during the session. Say, an `app.R` script that has `options()` calls followed by `teal::init()` (inside which we have callr call).

m7pr: Yes, need to double check `options()`. I think this is the most crucial, next to environmental variables

R/teal_lockfile.R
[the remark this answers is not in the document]
m7pr: I actually don't think that's possible..
Even if `create_renv_lockfile` would return `renv` message with the status, something at the top level of `teal_lockfile` would need to wait until `create_renv_lockfile` is done to pass the message to the `logger`.

R/teal_lockfile.R
m7pr: f0ee2fd6465322240a5f1b088c5afcc84ad13b89

R/teal_lockfile.R
m7pr: Environmental Variables are propagated

```r
> library(callr)
> rx <- r_bg(function() print(Sys.getenv('RSTUDIO_USER_IDENTITY')))
> rx$read_all_output()
[1] "[1] \"kosinsm4\"\r\n"
```

m7pr: `renv` uses `config`, `paths` and `settings` S3 objects in its setup. Those are not passed to `callr` unless you specify to set function environment to the specific package

```r
library(callr)
rx <- r_bg(function() print(names(settings)))
rx$read_all_output()
[1] ""
rx <- r_bg(function() print(names(settings)), package = 'renv')
rx$read_all_output()
> [1] " [1] \"bioconductor.version\"      \"ignored.packages\"         \r\n [3] \"external.libraries\"        \"package.dependency.fields\"\r\n [5] \"ppm.enabled\"               \"ppm.ignored.urls\"         \r\n [7] \"r.version\"                 \"snapshot.type\"            \r\n [9] \"use.cache\"                 \"vcs.manage.ignores\"       \r\n[11] \"vcs.ignore.cellar\"         \"vcs.ignore.library\"       \r\n[13] \"vcs.ignore.local\"         \r\n"
```

m7pr: Sounds like a bigger problem but maybe there is a way to pass `teal` as a reference package, that would somehow load/use `renv::settings`, `renv::config` and `renv::paths`. However this is a potential blocker

m7pr: I think we need to pass `options()`, `renv::settings`, `renv::config` and `renv::paths` as parameters to `create_renv_lockfile`

gogonzo: @pawelru made a good point with `options`, it looks like options are not passed from the current session (see code below). I think we don't need to pass all the options but only those needed by `renv` to correctly reproduce packages (this is why `repos` argument needs to be specified). This is why we don't need all options but only some:

1. app is hosted on the remote server where some options are set by `.Rprofile` file
2. app is hosted on the remote server where some envvars are set by `.Renviron` file 
3. app developer can set `options()` and `Sys.setenv()` (they should be included in the `qenv` as it is part of the preprocessing)
4. app is run with app the options and envvars combined.
5. app user runs the app and need to restore the default (system) environment. This is given by the `.Rprofile` and `.Renviron` which seems to be included in the `r_bg` by default.
6. If `renv.lock` includes sysvars and options then user is able to reproduce initial environment (before preprocessing)
7. `qenv` code includes necessary `options` and `envvars`  

We need to make sure that (5) and (6) are true, (7) is the responsibility of app developer.


```r
library(callr)
options(example_option = "hi there")
rx <- r_bg(function() {
  print(getOption("repos")) # this is from .Rprofile
  print(getOption("example_option")) # this is from current session
})
rx$read_all_output() |> cat()

#                                devel                              release 
# "https://pharmaverse.r-universe.dev"        "https://cloud.r-project.org" 
#                                CRAN 
#       "https://cloud.r-project.org" 
# 
# NULL
```

gogonzo: > `options()` **do not** get propagated by default (maybe there is a way to manually propagate them
> 
> ```r
> library(callr)
> options(user = 'MARCIN')
> rx <- r_bg(function() print(getOption('user', default = 'NOUSER')))
> rx$read_all_output()
> > [1] "[1] \"NOUSER\"\r\n"
> ```

Are you sure you didn't have an option `user` by default but make with different value?

m7pr: Passing all options is easier then some options because you can set multiple options with one call.

```r
# getting options
session_options <- options()
# setting options
options(session_options)
```

```r
library(callr)

rx <- r_bg(function(the_options) {print(options(the_options))}, list(the_options = options()))
rx$read_all_output()
```

So setting/passing options is fairy easy. What is problematic is passing `renv::settings`, `renv:;config` and `renv::config` that are locked bindings. IMHO `renv` doesn't utilize any options.

m7pr: I think those 4 should be enough for `renv` to have the same setup as in the main session

image


However those S3 objects in `renv` package are locked

```r
> assign('settings', value = settings, envir = getNamespace('renv'))
Error in assign("settings", value = settings, envir = getNamespace("renv")) : 
  cannot change value of locked binding for 'settings'
```

gogonzo: > IMHO renv doesn't utilize any options.

Uses but only `getOption("repos")`

image

pawelru: > I think we don't need to pass all the options but only those needed by renv to correctly reproduce packages

> Uses but only getOption("repos")

I disagree with the above. We don't really know right now what options are used. `renv` might call other packages that utilises its own set of options. I don't want to go through the whole functionality to identify list of options used. Moreover, this list might evolve in time so this introducing significant overhead on the maintenance. Please choose simple and low-maintenance solutions if possible.

2. Have you experimented with `package` argument of `callr::r_bg`. What would happen if you set it to `"renv"`?

gogonzo: @pawelru fair point. So, passing all options and system vars to the `callr::r_bg` is needed.

R/teal_lockfile.R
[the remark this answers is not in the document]
gogonzo: Ok, fair enough. I think it is more important to not change `future::plan` then caring so much about this log

R/teal_lockfile.R
gogonzo:
```
> assign('settings', value = settings, envir = getNamespace('renv')) Error in assign("settings", value = settings, envir = getNamespace("renv")) : cannot change value of locked binding for 'settings'
```

Can anyone explain me why to set `renv` settings? `renv` settings are changes through `options(renv.settings)` so passing `options` should be enough. Or not?

m7pr: - passing options is doable ✅ https://github.com/insightsengineering/teal/pull/1255#discussion_r1661366619
- system environmental variables are passed by default ✅ https://github.com/insightsengineering/teal/pull/1255#discussion_r1661140720

What is left to be passed are `renv::setting`, `renv::config` and `renv::paths`.

> Have you experimented with package argument of callr::r_bg. What would happen if you set it to "renv"?

Yes, I tried in here

image

we just need to check if this is the same `renv::setting` as in the main session or is this new `renv::setting` in the new session that is clean.

m7pr: I don't think there is anything like `options(renv.settings)`. `renv` uses it's own 3 objects to track its own options

m7pr: If we pass `package = 'renv'` then `renv::settings` and all objects from `renv` get propagated with their state from the main session

```r

library(callr)
settings$bioconductor.version('3.2222')
settings$bioconductor.version()
> [1] "3.2222"

rx <- r_bg(function() {print(settings$bioconductor.version())}, package = 'renv')
rx$read_all_output()
> [1] "[1] \"3.2222\"\r\n"
```

gogonzo: > I don't think there is anything like `options(renv.settings)`. `renv` uses it's own 3 objects to track its own options

`?renv::settings`
image

m7pr: Yes, but those `options` are empty

```r
> settings$snapshot.type()
[1] "explicit"
> options('renv.settings.snapshot.type')
$renv.settings.snapshot.type
NULL
````

```r
> settings$snapshot.type("explicit")
> settings$snapshot.type()
[1] "explicit"
> options('renv.settings.snapshot.type')
$renv.settings.snapshot.type
NULL
```

So I assume something is off with this `options` part. The description and examples states to use objects (`settings`, `config`, `paths`) directly

image

image

m7pr: If you set them, renv will get updated

```r
> options(renv.settings.snapshot.type = 'marcin')
> settings$snapshot.type()
[1] "marcin"
```

but at the beginning of the session, everything in `options(renv.settings*)` is NULL.

If someone changes options via `settings$function()` those are not reflected in `options()`. It's the other way around: if something is set via `options()` then it is reflected in `settings$function()`. So for the case where users set their specification via `settings$function()`, we need to pass `settings` object.

R/teal_lockfile.R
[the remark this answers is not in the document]
m7pr: @gogonzo I was able to bring back logging with this commit e04fff129ce0088b9d28748ca4d4d543957c798b
However there is a lot of information that you need to suppress if you want the `r_bg` function to print results to the console. It has it's own message and it's own custom log, so you need to run `capture.output` twice.

You can experiment with

```r
  x <- 
    utils::capture.output(
      utils::capture.output(
        r_bg(
          teal:::create_renv_lockfile, 
          args = list(lockfile_path = 'marcin.lock', opts = options()),
          stdout = ""
        ),
        type = "message"
      ),
    type = "output"
  )
```

The first `capture.output` is needed to suppress: `'Opening fd 1'` message
The second `capture.output` is needed to suppress: `'PROCESS 'Rterm', running, pid'` output

R/teal_lockfile.R
```diff
@@ -39,50 +39,54 @@ teal_lockfile <- function() {
...
+      prompt = FALSE,
+      force = TRUE
+      # type = is taken from renv::settings$snapshot.type()
+    )
+  )
+  # print() is required for callr::r_bg to track console output
+  if (any(grepl("Lockfile written", renv_logs))) {
+    print("lockfile created successfully.")
+  } else {
+    print("lockfile created with issues.")
```
gogonzo, line 88: I can't see this message in the console. Does it need `stdout = "|"`?

```r
options(
  teal.log_level = "TRACE",
  teal.show_js_log = TRUE,
)
library(scda)
pkgload::load_all("teal")
# pkgload::load_all("teal.slice")

data <- teal_data_module(
  ui = function(id) {
    ns <- NS(id)
    tagList(
    )
  },
  server = function(id, ...) {
      eventReactive(input$submit, {
        data <- teal_data() |>
          within(
            {
              logger::log_trace("Loading data")
              iris <- iris
            },
            username = input$username,
            password = input$password
          )

        join_keys(data) <- default_cdisc_join_keys[c("ADSL", "ADTTE")]

        data
      })
    })
  }
)

app <- teal::init(
  data = data,
  modules = modules(
    teal.modules.general::tm_data_table("Data Table"),
    example_module("Example Module", datanames = "ADTTE"),
    module(
      ui = function(id) {
        ns <- NS(id)
        tagList(
          tableOutput(ns("filter_summary"))
        )
      },
      server = function(id, datasets) {
          output$filter_summary <- renderTable({
            datasets$get_filter_overview(datanames = datasets$datanames())
          })
        })
      }
    )
  ),
  filter = teal_slices(
    teal_slice("ADSL", "SEX"),
    teal_slice("ADSL", "AGE", selected = c(18L, 65L)),
    module_specific = TRUE,
    mapping = list(
      global_filters = "ADSL AGE",
      `Example Module` = "ADSL SEX"
    )
  )
)

shinyApp(app$ui, app$server)

```

m7pr: I see it when I run `app <- teal::init` and I don't run `shinyApp(app$ui, app$server)`.
It seems like once you run `shinyApp(app$ui, app$server)` before lockfile is created, lockfile will not be created.
Once you end app session, you can see `Error: invoke_wrapped: throwing std::runtime_error`
Like `shiny` would block `callr` from running.

Unfortunately I was only testing today till `app <- teal::init` . I have seen it working during shiny runtime yesterday. Will troubleshoot what brought this inconvenience

tests/testthat/test-utils.R
```diff
@@ -254,12 +254,9 @@ testthat::test_that("defunction recursively goes down a list", {
 })
 
 testthat::test_that("create_renv_lockfile creates a lock file during the execution", {
-  old_plan <- future::plan(future::sequential)
-  withr::defer(future::plan(old_plan))
-
   renv_file_name <- "teal_app.lock"
   withr::defer(file.remove(renv_file_name))
-  promise <- create_renv_lockfile(TRUE, renv_file_name)
+  create_renv_lockfile(renv_file_name)
```
gogonzo, line 259: Please don't test private `utils` but rather test snapshot creation through `teal::init`.

m7pr: d4005709149db97ccc74455e160972b24f0c7b6b

R/teal_lockfile.R
gogonzo: This is the last thing which blocks me from approval. Please investigate if it is possible or remove `print`

m7pr: Yes, will try to figure out why this is not displayed when app is run

R/teal_lockfile.R
```diff
@@ -39,50 +39,54 @@ teal_lockfile <- function() {
...
+    utils::capture.output( # Needed to suppress: 'Opening fd 1' message
+      utils::capture.output( # Needed to suppress: 'PROCESS 'Rterm', running, pid' output
+        callr::r_bg(
+          func = create_renv_lockfile,
+          args = list(
+            lockfile_path = lockfile_path,
+            opts = options()
+          ),
+          # print results to the console
+          stdout = "",
```
pawelru, line 55: I don't really follow here.
We print to the console and capture the output not to print to console but to save to var instead but that var is nowhere used.

```suggestion
  callr_output <-
    utils::capture.output( # Needed to suppress: 'Opening fd 1' message
      utils::capture.output( # Needed to suppress: 'PROCESS 'Rterm', running, pid' output
        callr::r_bg(
          func = create_renv_lockfile,
          args = list(
            lockfile_path = lockfile_path,
            opts = options()
          ),
          # print results to the console
          stdout = "",
```

R/teal_lockfile.R
```diff
@@ -39,50 +39,54 @@ teal_lockfile <- function() {
...
+          # renv setup is orchestrated by its special S3 objects: renv::settings, renv::config and renv::paths
+          #     `package` parameter include `renv` namespace inside the environment of `func = create_renv_lockfile` fun
+          package = "renv"
+          # default env = NULL # means that callr process will use environmental variables
+          #     from the main session (parent process)
```
pawelru, line 60: why it is there? you provide the explanation to the commented out code?

gogonzo: Looks alike, but not. It is an explanation of the defaults to understand how `env` is propagated to `r_bg`. It is not obvious from the documentation, we needed to dig to the docs of `processx::process`. Probably comment should be rephrased then to not suggest that it explains a commented-out code.

pawelru: if this is really needed: maybe uncomment this? - even if that's the default param value
right now it's look a little bit odd to me

m7pr: I think uncommenting the `env = NULL` and leaving the comment after it would be the cleanest

R/teal_lockfile.R
m7pr: What is returned by `callr::r_bg(` with `stdout = ""` is printed directly to the console through `callr` process. On top of it, main session print a message and a regular text to the console. With two `capture.outputs` we disable what main session prints to the console. This way we only get what `callr` process printed. However I don't think `callr_output <-` is necessary.

This whole logging thing might not be working while shiny will be occupying the console output during the runtime. I am investigating this in https://github.com/insightsengineering/teal/pull/1255#discussion_r1662467473

R/teal_lockfile.R
m7pr: I just checked that `r_bg` logging is not blocked if shiny app is running. e.g.

```r
library(callr)
rx <- r_bg(function() {Sys.sleep(5); print('Hi!'); renv::snapshot(prompt = FALSE, force = TRUE)}, stdout = "")

app <- init(
  modules = example_module(label = "example teal module")
)

shinyApp(app$ui, app$server)

```

Produces renv.lock file and renv output is pasted in the console. Trying to figure out why it's not happening in teal

R/teal_lockfile.R
m7pr: Removed the var and one `caputre.output` as it was not needed in the end e7c7c7b5fbea4f8850e07f3d1c445e4a77a348d5

R/teal_lockfile.R
m7pr: In the end we figured out that we can pass
- `options()` as a parameter to `create_renv_lockfile`
- `renv` namespace to `create_renv_lockfile` environment with `r_bg` parameter (`package  = 'renv'`)
- environmental variables from main session by default to the `callr` process with `r_bg` parameter (`env = NULL`)

```r
callr::r_bg(
      func = create_renv_lockfile,
      args = list(
        lockfile_path = lockfile_path,
        opts = options()
      ),
      # prints results to the console
      stdout = "",
      # renv setup is orchestrated by its special S3 objects: renv::settings, renv::config and renv::paths
      #     `package` parameter include `renv` namespace inside the environment of `func = create_renv_lockfile`
      package = "renv" ,
      # callr process will use environmental variables from the main session (parent process)
      env = NULL
    )
```

R/teal_lockfile.R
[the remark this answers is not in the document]
m7pr: Simplified it so we only run

```r
utils::capture.output(
        r_bg(
          teal:::create_renv_lockfile, 
          args = list(lockfile_path = 'marcin.lock', opts = options()),
          stdout = ""
        ),
        type = "message"
      )
```

to capture `'Opening fd 1'` message

R/teal_lockfile.R
m7pr: For the `env` parameter I actually think we need to go with `env = Sys.getenv()` because `r_bg` by default uses `env = rcmd_safe_env()` 

```r
CYGWIN            R_TESTS          R_BROWSER        R_PDFVIEWER 
"nodosfilewarning"                 ""            "false"            "false"
```

`r_bg` also does not allow `env = NULL` even though `processx::process$new()` allows that. In `processx::process$new()` `env = NULL` means all parent system variables, but when `r_bg` limits them I think it's safer to use `Sys.getenv()`. 

`processx::process$new()` docs claims that you can append some variables to the list of all system variables
image
but I don't think `callr` is appending anything with `current` value 
https://github.com/search?q=repo%3Ar-lib%2Fcallr%20%22current%22&type=code

R/teal_lockfile.R
m7pr: Hey @gogonzo this is very strange behavior. I think I reached the wall.
I pushed a test version of `create_renv_lockfile_2` to demonstrate one behavior.
When you run `shinyApp()` all `callr` processes keep running, but the ones started inside `teal` get frozen

Try this code on the current branch

```r
library(callr)
rx <- r_bg(function() {lapply(1:15, function(x) {print(paste0('custom: ', x)); Sys.sleep(1)})}, stdout = "")

app <- init(
  modules = example_module(label = "example teal module")
)

shinyApp(app$ui, app$server)
````

Counter for the `r_bg` process started in the main session gets till the end of the counting, but the counter in teal stops. This is just to show that teal-callr process is stopped/killed and the lockfile will never be created when `shinyApp()` is run

image

File: R/teal_lockfile.R [hunk 9]
```diff
@@ -39,50 +39,57 @@ teal_lockfile <- function() {
...
+    # callr process will use environmental variables from the main session (parent process)
+    env = Sys.getenv()
+  ) # ,
+  #   type = "message"
+  # )
+
+  logger::log_trace("lockfile creation started.")
+}
+
+create_renv_lockfile_2 <- function(lockfile_path = NULL, opts = NULL) {
```
m7pr, line 66: This is just for testing, if the callr process is working or not, and when it's stopped/killed

R/teal_lockfile.R
m7pr: As you mentioned, I needed to store the process to a variable to prevent it from being killed 027faee48cda8a1c067a0b59d195fb05385f0472

R/teal_lockfile.R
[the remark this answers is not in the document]
m7pr: We stored the `callr` in a variable that is accessible in `init`. Last thing is to use `reactivePoll` or `reactiveTimer` to detect that the process finished, so that we can print logs about the finalization of the process.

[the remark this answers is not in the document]
m7pr: Useful internet resources to read and watch 
https://www.youtube.com/watch?v=DTMVzK7iZFU&t=1629s
https://towardsdev.com/a-simple-workflow-for-async-shiny-with-callr-f32393b087be

R/init.R
```diff
@@ -231,6 +231,8 @@ init <- function(data,
       if (!is.null(landing_module)) {
         do.call(landing_module$server, c(list(id = "landing_module_shiny_id"), landing_module$server_args))
       }
+      session$userData$lockfile_process <- process
+      reactivePoll(1000, session, checkFunc = lockfile_status, valueFunc = function(){})
```
m7pr, line 235: This does not yet produce `logger::log_trace` messages.
TODO for the further investigation.

Potential helpful resources on the `callr`

https://www.youtube.com/watch?v=DTMVzK7iZFU&t=1629s
https://towardsdev.com/a-simple-workflow-for-async-shiny-with-callr-f32393b087be

R/teal_lockfile.R
```diff
@@ -39,50 +39,41 @@ teal_lockfile <- function() {
...
+
+  logger::log_trace("lockfile creation started.")
+  process
 }
 
-create_renv_lockfile <- function(close = FALSE, lockfile_path = NULL) {
-  checkmate::assert_flag(close)
+create_renv_lockfile <- function(lockfile_path = NULL, opts = NULL) {
   checkmate::assert_string(lockfile_path, na.ok = TRUE)
-  promise <- promises::future_promise({
```
pawelru, line 58: I'm sorry if I am undermining the whole callr work but isn't it that easy as adding `seed = NULL` argument?

I'm circling back as I can't find the rationale behind switch to callr in there (please redirect me if it's there). It looks like the new standard is ExtendedTask with future-like objects. This goes completely the opposite way. Wanted to understand the reasons.

gogonzo: @pawelru the biggest issue with `future` is that we change the `future::plan` and trying to restore it but we can't promise we will restore it on time. Any plan change by the app developer or module developer is a potential risk and highlighted by `future` [authors that one should avoid setting it](https://future.futureverse.org/articles/future-7-for-package-developers.html#avoid-changing-the-future-strategy).


I give you another example which fails the app. Please do following: 
- run the app on @main. You should get an error. 
- run the app again in the same session. App gets through and returns "42" in the module.
- Then please comment out `teal_lockfile()` in teal and try again in a new session. App works.

Problem we're facing is that `future` loses socket to the jobs in previous plans.

```r
plan(multisession)
v1 %<-% {
  Sys.sleep(1)
  cat("Hello world!\n")
  3.14
}
plan(sequential)
v2 %<-% {
  cat("Hello world!\n")
  Sys.sleep(1)
  3.14
}
v1
# Error: MultisessionFuture (<none>) failed to receiving message from cluster RichSOCKnode #1 (PID 93358 on localhost ‘localhost’). The reason reported was ‘Connection to the worker is corrupt’. 
# Post-mortem diagnostic: No process exists with this PID, i.e. the localhost worker is no longer alive. The socket connection to the worker of MultisessionFuture future (<none>) is lost or corrupted: 
# Connection (connection: index=4, description="NA", class="NA", mode="NA", text="NA", opened="NA", can read="NA", can write="NA", id=1786, raw_id="<pointer: 0x6fa>") is no longer valid. 
# There is currently no registered R connection with that index 4. As an example, this may happen if base::closeAllConnections() have been called, for instance via base::sys.save.image() which in turn is called if the R session (pid 90548) is forced to terminate
v2
```

Feel free to suggest any solution involving `future` which will make it more robust. So far I found `callr::r_bg` solution have not thrown any errors.

<details>
<summary> app </summary>

```r
# library(teal)
pkgload::load_all("~/Documents/nest/teal")
setwd("~/Documents/nest/sample_apps/sample_app_renv_lockfile")


app <- teal::init(
  data = teal_data(obj = data.frame()),
  modules = modules(
    module(
      ui = function(id) {
        ns <- NS(id)
        tagList(
          textOutput(ns("text1")),
          textOutput(ns("text2"))
        )
      },
      server = function(id, data) {
          output$text1 <- renderPrint({
            future::plan("multicore")
            future::future({
              a <- 7
              b <- 3
              c <- 2
              a * b * c
            })
          })
        })
      }
    )
  )
)

runApp(app)

```

</details>

m7pr: Changing the plan closes/looses other tasks so we are using `promise::then` at the end of the execution of the task so that we close the plan when the task is finished. But starting and closing multiple apps one after another might cause that some tasks did not succeed

pawelru: Thank you for the explanation. I think it all makes sense and it's coming down into the topic of changing the plan. This has to be done more carefully on all parties - app developer, core tool developers (us) and also module developers. I think that the golden rule should be as follows: if the plan has been changed (very easy to detect - see below) - do not change it.

So in other words - we (from the core tool perspective) are going to change the plan if it was not set by ~app developer~ anyone else before. This would have to be documented in a prominent place.

(As an extra extension - we can add an option to better control the settings used. Right now it's hardcoded multisession with 2 workers - maybe one wanted to alter it?)

As a result:
- we will not change the plan set by app developers before init
- module developers cannot change the plan set by either of app developers or us

The latter point is actually true right now. Take the lockfile away - if there are multiple modules and each set its own plan then the app would have a problem. I see your example in this way - take teal away and use base shiny with two modules each trying to set its own, different plan. Module developers also needs to be very careful with this.

See my proposal below:
image

pawelru: (and a little bit of more context behind my ask)
I am having a high hopes for `ExtendedTask`. This was advertised as the way to go for long-running tasks. I think I can name a few in which this would be useful. Therefore I think that the plan setting should be at the {teal} core (inside `init()`; optionally controlled with set of options) and all the modules and teal extensions would be able to use it as they wish.

gogonzo: @pawelru all makes sense, but what about the case when a module developer has any `future::plan` in the teal_module code,  then:
- we are setting the `plan("multisession")` for renvlock-process in `teal::init`
- custom `teal_module` is called with `plan("sequential")` which automatically loses renvlock-process

If the answer is "module developer should be more careful" I'd say "module developer has already a lot to learn about `teal` and this would be yet another thing to keep in mind"

pawelru: Yes and no :)
Yes as my answer is unfortunately "module developer should be more careful".
No because this has nothing to do with `teal`. These are all `future` requirements and limitations you linked above. Module should be seen as a function and in [here](https://future.futureverse.org/articles/future-7-for-package-developers.html#avoid-changing-the-future-strategy) we have:

> For reasons like the ones mentioned above, refrain from setting [plan()](https://future.futureverse.org/reference/plan.html) in a function. It is better to leave it to the end-user to decided how they want to parallelize. One reason for this is that we can never know how and in what context our code will run. (...) 

> If you still think it is necessary to set [plan()](https://future.futureverse.org/reference/plan.html), make sure to undo when the function exits, also on errors.

Change "function" to "module", "context" into "app" - you know what I mean here.

See my previous message that this is true even without us touching this at all. I will give you an example. What if you are app developer and me and Marcin are module developers. I'm doing `future::plan` so does Marcin. Then you have a problem. My point is that everyone (incl. module developers) who is doing `plan()` needs to be careful with it and adhere to the docs you linked above. That's completely independent of us touching this.

gogonzo: @pawelru 

| What if you are app developer and me and Marcin are module developers. I'm doing future::plan so does Marcin. Then you have a problem.

| you might also call a single module (that runs plan()) multiple times (because why not?). And then even with a one module - you as an app developer - would have a problem.

I think both are not teal-specific problems. They can happen in any shiny app and  we shouldn't complicate this even more in `teal`. I think that `teal` messing with `future` is a mistake and brings unnecessary complications. 

I thing about following scenarios:

1. App developer don't use future and none of teal_modules uses future. In `teal` we are allowed set `plan("multisession")` as it hasn't been changed by app dev. When task is finished (`on.exit`) we set it back to original plan.
2. App developer sets `plan(...)`, teal_modules don't use future.  In `teal` we detect that plan has been changed and we call `teal_lockfile()` on the user-specified-plan. Plan set to "sequential" delays an app `init` because teal app awaits `teal_lockfile()` to finish. Luckily, this delays only `init`, so the next session will be started without lags.
3. App developer don't use future and teal_module uses future. And here are two possible ways for this:
  a. teal_module sets `plan(xyz)` - if we keep the same condition - to change `plan(multisession)` only if app developer didn't set, then we introduce a bug, because `on.exit(...)` is possibly invoked too late as our task runs asynchronously and teal_module with `plan(...)` can be called. Yes, I agree that setting the `plan` by module developer is the same sin as us setting this in `teal::init`, but I think we shouldn't immediately punish module developer for one's wrongdoing. Problem is that we can't detect if any module in the app sets `plan(...)` so we can't even prevent the failure. Failure message could be very cryptic and it will be hard to point exactly which `plan()` call caused a problem.
  b. teal_module can't call a `plan` as generally recommended and app developer sets the plan - then we have scenario (2) 
  
Because of (2), (3a) and (3b) I'm against using `future`. I think lockfile should be created asynchronously always and not being dependent on any app-developer or module-developer decision. Looking at `@main` `ExtendedTask` is not utilized properly so I don't see the point of using this at all (`ExtendedTask` is used just to invoke a task) as invoking future can be done without `ExtendedTask` too.

My general opinion is - why to complicate this with `future` while we can simply call `r_bg` without being worried about (2), (3a) and (3b).

m7pr: After experimenting with `callr` and `r_bg` on this branch I personally believe `callr` is simpler to run and easier to manage and does not mess up with `plan`.  There are also `future` warnings and errors pasted to a console when an app is stopped by the process is still running

R/init.R
m7pr: Had a chance to watch the youtube tutorial on callr today but didn't learn more than what we already knew

m7pr: Hey @gogonzo the `reactivePoll` wasnt the right tool (or I wasn't able to properly use it!).
I managed to execute `lockfile_creation` status with `reactiveTimer` + `observe()` + `$destroy()` combination in here 3c480a7be0524dbcaa30f2a2d6165069df813e9c

R/teal_lockfile.R
m7pr: Even though `ExtendedTask` and `future::future_promises` is the new way to go, I think we've been abusing this solution and not utilizing the whole aspects of the flow. I would advise to go with `callr` solution as I see less issues with hanging and unfinished processes. Happy to meet and discuss next week if you guys feel that this conversation is not over yet

pawelru: Thanks for the comments guys. Let's go with callr. I think that it's quite likely that the future will come back to us but we can revisit this later on (if at all).  For now let's go with callr. Resolving this

R/teal_lockfile.R
```diff
@@ -105,10 +96,23 @@ teal_lockfile_downloadhandler <- function() {
   )
 }
 
-is_r_cmd_check <- function() {
-  ("CheckExEnv" %in% search()) || any(c("_R_CHECK_TIMINGS_", "_R_CHECK_LICENSE_") %in% names(Sys.getenv()))
+lockfile_status <- function(process) {
+  renv_logs <- process$read_output()
+  if (any(grepl("Lockfile written", renv_logs))) {
```
gogonzo, line 101: 1. This `if else` is incomplete. I made `renv` to fail and I printed the message:

image
I think that success or failure should be followed by `showNotification("Lockfile created successfully.")` or `showNotification("Lockfile failed to create.")`


2. I think download lockfile shouldn't be shown when lockfile is not created.

tests/testthat/test-init.R
```diff
@@ -77,3 +77,20 @@ testthat::test_that("init throws when dataname in filter incompatible w/ datanam
     "Filter 'iris Species' refers to dataname not available in 'data'"
   )
 })
+
+testthat::test_that("init creates a lockfile during the execution", {
```
gogonzo, line 81: test for failed lockfile creation would be nice also

m7pr: We use `renv::snapshot` with `force = TRUE` - this forces the generation of a lockfile even when pre-flight validation checks have failed. So the file will be created anyway.

It's not like that the file gets created or not. The file gets created even though it might have data that can only be partially restorable. So for the notifications you mentioned in here https://github.com/insightsengineering/teal/pull/1255#discussion_r1672837226

Instead of `Lockfile failed to create.` we put an info that `lockfile created with issues.`

However, we can try to develop a test that returns this lockfile that was created with issues

R/teal_lockfile.R
m7pr: > I think download lockfile shouldn't be shown when lockfile is not created.

This simplified the code a bit, as we no longer need to wait once the download button is clicked but the process is not finished 6010c4fb1d08b9de8c4c52afafb398e66421570f

R/teal_lockfile.R
```diff
@@ -92,23 +83,33 @@ teal_lockfile_downloadhandler <- function() {
     },
     content = function(file) {
       teal_lockfile <- "teal_app.lock"
-      iter <- 1
-      while (!file.exists(teal_lockfile) && iter <= 100) {
-        logger::log_trace("lockfile not created yet, retrying...")
-        Sys.sleep(0.25)
-        iter <- iter + 1 # max wait time is 25 seconds
-      }
```
m7pr, line 100: This can be removed as `downloadHandler` will be only shown once `lockfile_status()` is run
so just after the `process <- teal_lockfile()` is finished (`process$is_alive() == FALSE`).

tests/testthat/test-init.R
m7pr: But it's hard to know if the lockfile has issues in it, once it's created. I think the only way is the attempt to `renv::restore()` that fails.

m7pr:
Hey @gogonzo this is what I get for the simple app you asked for to test with

```r
options(teal.log_level = "TRACE")

library(shiny)
pkgload::load_all(".")

fun <- function() {
  process <- teal_lockfile()
  
  app <- shinyApp(
    ui = fluidPage(),
    server = function(input, output, session) {
      lockfile_status_tracker(process)
    }
  )
}

app <- fun()


runApp(app)
```

image

m7pr:
In the end `callr` was lost on macOS when we started shiny app. Closing this PR in favour of `mirai` approach https://github.com/insightsengineering/teal/pull/1263

PR #1256, opened 2024-07-01
Adds unfiltered data to data passed to modules

### LINKED ISSUE #1229: Add unfiltered data to the teal_data
When passing data to the module all filtered datasets are included in `teal_data` (by datanames). We need to include unfiltered data (use `filtered_data$get_data(filtered = FALSE)`) and they should be called `<dataname>_raw` ([look here](https://github.com/insightsengineering/teal/blob/5c3e3fd4380ef0fc7efc53ec1cdf4f5a7c922dd3/R/module_nested_tabs.R#L278). Inclusion of this dataset should be reflected in the `@code` slot

P.S. `_raw` because it is the shortest name. I wanted to avoid `_unfiltered`.

averissimo:
# Pull Request

Fixes #1229

### Changes description

- Adds unfiltered data to `teal_data` being passed to module
- Adds relevant code `<dataname>_raw <- <dataname>` after hashes

#### Notes

- This approach keeps original code from `teal_data` initialization and copies it before "filter statements"

Alternative approach would create expression and convert it to character vector

```r
paste(
  vapply(
    datanames,
    function(x) {
      deparse1(
        substitute(
          raw_var <- original_var,
          list(raw_var = as.name(sprintf("%s_raw", x)), original_var = as.name(x)
          )
        )
      )
    },
    character(1L)
  ),
  collapse = "\n"
)
```

#### Test

```r
pkgload::load_all("../teal")

  eval_code("
    IRIS <- iris
    IRIS <- IRIS[IRIS$Species != \"setosa\", ] # @linksto IRIS
    MTCARS <- mtcars
  ")
teal.data::datanames(data) <- c("IRIS", "MTCARS")

teal.code::get_code(data, datanames = "IRIS")

app <- teal::init(
  data = data,
  modules = list(
    teal.modules.general::tm_data_table("Data Table"),
    example_module("Example Module"),
    example_module("Example Module (iris only)", datanames = "IRIS")
  ),
  filter = teal::teal_slices(
    teal.slice::teal_slice("IRIS", "Species", selected = c("setosa", "virginica"))
    # module_specific = TRUE,
    # mapping = list(
    #   global_filters = "IRIS Species",
    #   `Example Module` = "IRIS Species"
    # )
  )
)

runApp(app)
```


averissimo:
I placed the raw copying below the data integrity check and after the unchanged code from `teal.data` / `teal_data_module`

Adds function `.get_raw_code` to keep the same convention of calling function to generate code (although the function body is very simple).

R/1.0_module_teal_modules.R
```diff
@@ -302,20 +302,31 @@ srv_teal_module.teal_module <- function(id,
 }
 
 .make_teal_data <- function(modules, data, datasets, datanames) {
-  filtered_datasets <- sapply(datanames, function(x) datasets$get_data(x, filtered = TRUE), simplify = FALSE)
+  new_datasets <- c(
+    # Filtered data
+    sapply(datanames, function(x) datasets$get_data(x, filtered = TRUE), simplify = FALSE),
+    # Raw (unfiltered data)
+    rlang::set_names(
```
gogonzo, line 309: why not setNames in this case?

R/1.0_module_teal_modules.R
```diff
@@ -345,3 +356,14 @@ srv_teal_module.teal_module <- function(id,
...
+
+#' Get code that creates a backup of the original (unfiltered) data
+#'
+#' @param datanames (`character`) names of datasets
+#' @return A character vector with code lines that copy existing `<datanames>` to
+#' `<datanames>_raw`.
+#'
+#' @keywords internal
+.get_raw_code <- function(datanames) {
+  paste(sprintf("%s_raw", datanames), "<-", datanames, collapse = "\n")
```
gogonzo, line 368:
```suggestion
  sprintf("%1$s <- %1$s_raw", datanames)
```

gogonzo: Looking how simple this function can be I suggest to just make inline `sprintf` instead of making extra function

R/1.0_module_teal_modules.R
averissimo: Done in 6a704dba2

R/1.0_module_teal_modules.R
averissimo: True, I don't use sprintf enough

R/1.0_module_teal_modules.R
```diff
@@ -302,20 +302,31 @@ srv_teal_module.teal_module <- function(id,
...
+  raw_data_code <- sprintf("%1$s_raw <- %1$s", datanames)
+  filter_code <- get_filter_expr(datasets, datanames = datanames)
 
-  all_code <- paste(unlist(c(data_code, "", hashes_code, "", filter_code)), collapse = "\n")
+  all_code <- paste(unlist(c(data_code, "", hashes_code, raw_data_code, "", filter_code)), collapse = "\n")
   tdata <- do.call(
     teal.data::teal_data,
     c(
-      list(code = all_code),
+      list(code = trimws(all_code, which = "right")),
```
gogonzo, line 324: `trimws` doesn't make sense here because it doesn't recognise end of the line. If you want to remove whitespaces for each line you need to do this earlier, because in this line you're dealing with concatenated single character.
```suggestion
      list(code = all_code),
```

R/1.0_module_teal_modules.R
```diff
@@ -302,20 +302,31 @@ srv_teal_module.teal_module <- function(id,
...
       list(join_keys = teal.data::join_keys(data)),
-      filtered_datasets
+      new_datasets
     )
   )
+
   tdata@verified <- TRUE # todo: change to original value from from data
```
gogonzo, line 330: @averissimo please address this todo: after you merge this branch into feature branch. You can commit directly

gogonzo (approved):

PR #1259, opened 2024-07-04
fix `noSuggests` check

pawelru:
This is my testing branch of r-hub workflows. If you see this as non-draft that means that I have removed my workflow file and therefore making it ready to be merged (this is do be done separately on the next propagation round).

The result of my tests are here: https://github.com/insightsengineering/teal/actions/workflows/rhub.yaml

One of the checks is "noSuggests" check and it runs the check without suggested packages. This fixes encountered errors.

Please check if the downstream packages requires similar updates. If so, please kindly do this or let me know.


tests/testthat/test-validate_inputs.R
pawelru, line 1: FYI: the first test (and only one!) does not require `shinyvalidate` hence I cannot introduce `skip` at the beggining. There is also an option to introduce skip somewhere in line ~11 (i.e. right after the first test) but if we already have a test independent on `shinyvalidate` - I can definitely imagine a new test(s). Hence decided to introduce skip inside test_that. It's a little bit more clutter but that would keep this file easily extendable.

.github/workflows/rhub.yaml
pawelru, line 1: TODO: remove this file before merge

m7pr:
@pawelru how I would be able to trigger r-hub workflow https://github.com/insightsengineering/teal/actions/workflows/rhub.yaml ?
I would like to run in on master for comparison. Other workflows have `This workflow has a workflow_dispatch event trigger.` notification and `Run workflow`  button but this one doesn't.

I see there is a `workflow_dispatch` in the configuration file, but the indentation might be improper?
image
Checkout indentation in this workflow yaml https://github.com/insightsengineering/teal/actions/runs/8196271417/workflow

m7pr:
@pawelru did you delete `.github/workflows/rhub.yaml` file? I think this is the cause why I can't trigger or run `R-Hub` workflow

m7pr:
Is this the check `noSuggests` in this workflow `r-hub.yaml` that we are testing? https://github.com/insightsengineering/r.pkg.template/blob/main/.github/workflows/rhub.yaml#L34

pawelru:
Yes exactly. This workflow is to be added in https://github.com/insightsengineering/teal/pull/1285 and this feature branch was used when developing this functionality. I have removed this assuming that this will be added separately (with all the other stuff waiting for propagation). See my commit https://github.com/insightsengineering/teal/pull/1259/commits/5e4d2a468b76b80c3ac63255bb30b0e6c1649f36 in which I removed the workflow.

For testing, I would recommend to remove all the suggested packages and run the tests. We expect no errors but a couple of skips.

m7pr:
@pawelru which of the workflows in teal https://github.com/insightsengineering/teal/tree/main/.github/workflows actually use `rhub` job? Or we do not plan to run `rhub` job in this repository?

m7pr:
Ah I see. Scheduled will be extended with this new check https://github.com/insightsengineering/teal/pull/1285/files#diff-bff33451e6a8cdc8353268c5ec4a59426869f0d7c4caf410da4b267b9cb3ff57R34

m7pr:
Last job failed for this branch for rhub workflow
image

I tried to restart it but got the following error
image

I guess because after this PR was merged https://github.com/insightsengineering/r.pkg.template/pull/232, this branch ` test-rhub-workflows` was deleted and the old configuration that pointed to this branch can not  be run anymore. I base my acceptance based on the previous run that was successful https://github.com/insightsengineering/teal/actions/runs/9809126389

m7pr:
So as far as I understand: all packages will now have an extra run `rhub` that is run with `scheduled` action. So each scheduled period of time this is executed. For this package we got tests improved in this PR, and for other packages we might see errors that needs attention once the new workflow propagation gets merged to master branches.

m7pr (approved):

PR #1263, opened 2024-07-12
479 mirai lockfile@main

gogonzo:
https://github.com/insightsengineering/teal/issues/1276

@pawelru @m7pr 
This is a `mirai` alternative. Package seems to address the biggest issues reported during a research:
- `mirai` has a native support of `ExtendedTask`
- `mirai` is not being killed when `runApp` is executed (like `callr` does)
- `mirai` by default opens a deamon in parallel R session without a need to handle the `future::plan`.
- `mirai` has only one dependency in the whole dependency tree.

#### Disadvantages so far: 
- ~~we need to pass and set `options`, system vars, working directory and `.libPaths` https://github.com/shikokuchuo/mirai/issues/122~~

#### How does it work:
- lockfile creation is invoked in `init` before application starts. This prevents to start the process each time when a new shiny session starts. Process is invoked as a promise and eventually `teal_app.lock` will be created
- When shiny session starts `download lockfile` button is hidden by default. If promise is eventually resolved and lockfile is created then download button is shown.
- alternatively, app developer can pre-compute lockfile and provide its path in `teal.renv.lockfile` option. In such case `renv::snapshot` will be skipped and user lockfile will be used in an app.

#### Logs and notifications
Logs are printed for app developer while notifications are presented to the app user:
1. When app uses precomputed file:
- log in init: `Lockfile set using option "teal.renv.lockfile" - skipping automatic creation.`
- no notification to the app user.
2. When app automatically determines snapshot:
- log in init: `Lockfile creation started based on { getwd() }.`
- log If lockfile created: `Lockfile {path} containing { n-pkgs } packages created{ with errors or warnings  }.`
- notification if lockfile created: `Lockfile available to download`
- log if lockfile not created: `Lockfile creation failed.` 
- notification if lockfile not created: `Lockfile creation failed.`


R/teal_lockfile.R
```diff
@@ -23,92 +44,140 @@
...
-    # If the initial backend was only sequential, bring it back.
-    promises::then(promise, onFulfilled = function() {
-      future::plan(future::sequential)
-    })
+    logger::log_trace("Lockfile {renv_logs$lockfile_path} containing { renv_logs$length } packages created{ with }.")
+    shiny::showNotification("Lockfile is available to download.")
+    shinyjs::show("teal-lockFile")
+  } else {
+    warning("Lockfile creation failed.")
+    shiny::showNotification("Lockfile creation failed.", type = "warning")
```
m7pr, line 165: I would show it anyway with a NOTE, that you download the file on your own risk, since the file was created but `renv::restore` might not be able to restore project based on this lockfile.

```suggestion
    shiny::showNotification("Lockfile creation failed.", type = "warning")
    shinyjs::show("teal-lockFile")
```

m7pr:
Overall looks good based on findings researched during `future+shiny::ExtendedTask` and `callr` approaches :)

R/teal_lockfile.R
```diff
@@ -23,92 +44,140 @@
...
     }
   }
 
-  if (!(is_in_test() || is_r_cmd_check())) {
-    old_plan <- future::plan()
-    # If there is already a parallel (non-sequential) backend, reuse it.
-    if (inherits(old_plan, "sequential")) {
-      future::plan(future::multisession, workers = 2)
-    }
+  process <- ExtendedTask$new(function(...) {
```
m7pr, line 75:
```suggestion
  process <- shiny::ExtendedTask$new(function(...) {
```

chlebowa: Is it necessary to prefix `ExtendedTask`? The whole `shiny` is imported and it's functions are not prefixed throuhgout.

R/teal_lockfile.R
```diff
@@ -23,92 +44,140 @@
...
-    renv_logs <- utils::capture.output(
-      renv::snapshot(
-        lockfile = lockfile_path,
-        prompt = FALSE,
-        force = TRUE
-        # type = is taken from renv::settings$snapshot.type()
-      )
+  # mirai starts in vanilla session in the R.home directory. We need to pass all session related info
+  options(opts)
+  lapply(names(sysenv), function(sysvar) do.call(Sys.setenv, sysenv[sysvar]))
```
chlebowa, line 106: I think this is sufficient.
```suggestion
  do.call(Sys.setenv, sysenv)
```

R/teal_lockfile.R
m7pr: unsure, then I suppose not need for the prefix - it was just a live edit during a call to figure out what are the real imports/dependencies of `mirai` package

R/teal_lockfile.R
gogonzo: Good point

```r
Sys.setenv(ZZZ = "1")
vars <- Sys.getenv()
Sys.setenv(ZZZ = "2")
do.call(Sys.setenv, as.list(vars))
Sys.getenv()[["ZZZ"]] # expects 1
#> [1] "1"
```

R/teal_lockfile.R
```diff
@@ -23,92 +44,140 @@
...
+  }
+}
+
+#' @rdname teal_lockfile
+#' @keywords internal
+teal_lockfile_handler <- function(process) {
+  # todo: make sure that we catch all possible status outputs
+  #     - normally getwd should be set to the app directory (snapshot based on the app files)
+  #     - what happens if app directory is set to a different location?
+  #     - what if setwd is set to a directory which doesn't contain anything
```
m7pr, line 148: For the empty directory a lockfile with `renv` only is created. Like for an empty, just created project with `renv` setup.

R/teal_lockfile.R
```diff
@@ -23,92 +44,140 @@
...
+  #     - check out possible ERROR and WARNING
```
m7pr, line 149: renv has +40 `R/*.R` files https://github.com/rstudio/renv/tree/main/R and 174 occurrences of `ERROR/error` in it's repo https://github.com/search?q=repo%3Arstudio%2Frenv%20ERROR&type=code

I think it's hard to troubleshoot all for their edge cases, so I would rather just seek for WARNING/ERROR in the renv log.

gogonzo: Cool 👍 we can implement some custom behaviour if somebody request it in the future 👍

R/teal_lockfile.R
```diff
@@ -23,92 +44,140 @@
...
+    shinyjs::show("teal-lockFile")
+  }
+}
+
+#' @rdname teal_lockfile
+#' @keywords internal
+teal_lockfile_handler <- function(process) {
+  # todo: make sure that we catch all possible status outputs
+  #     - normally getwd should be set to the app directory (snapshot based on the app files)
+  #     - what happens if app directory is set to a different location?
```
m7pr, line 147: App directory can be specified with `shinyAppDir()`, to point to a directory containing `server.R` or `app.R`. If you run shiny app with `shinyAppDir(path)` the `getwd()` shows `path` which is good, and as expected. 

However in `teal` use case `teal::init` creates `ui` and `server` functions, that can be used in `shinyApp(ui = ui, server = server)`. It is possible that the file that launches `shinyApp(ui = init$ui, server = init$srv)` is not in working directory and the `getwd()` is used and points to something else. Investigating on how to track actual directory with the app.

m7pr: I think we may want to use `rstudioapi::getActiveDocumentContext()$path` which tracks the active document context to figure out the app directory

image

m7pr:
There is also a NOTE in R CMD CHECK that we need to eliminate

```r
* checking R code for possible problems ... NOTE
teal_lockfile_invoke : <anonymous>: no visible global function
  definition for ‘run’
teal_lockfile_invoke : <anonymous>: no visible binding for global
  variable ‘opts’
teal_lockfile_invoke : <anonymous>: no visible binding for global
  variable ‘sysenv’
teal_lockfile_invoke : <anonymous>: no visible binding for global
  variable ‘libpaths’
teal_lockfile_invoke : <anonymous>: no visible binding for global
  variable ‘wd’
Undefined global functions or variables:
  libpaths opts run sysenv wd
```

m7pr:
I think the function in process creation needs to be defined as a regular function

```r
  process <- ExtendedTask$new(function(...) {
    mirai::mirai(
      run(lockfile_path = lockfile_path, opts = opts, sysenv = sysenv, libpaths = libpaths, wd = wd),
      ...
    )
  })
```

```r
  process <- ExtendedTask$new(run_mirai)
  run_mirai <- function(lockfile_path, opts, sysenv, libpaths, wd) {
  mirai::mirai(
      renv_snapshot(lockfile_path = lockfile_path, opts = opts, sysenv = sysenv, libpaths = libpaths, wd = wd),
    )
 }
```

R/teal_lockfile.R
m7pr: I think this is too complicated to get this detected by `teal`. @gogonzo what if we introduce new option `options(teal.app.dir)` that hosts the directory of the teal app. This probably would have more applications in other use cases. This can be set to `getwd()` by default.

gogonzo:
> ```r
> run(lockfile_pa
> ```

you probably need to `teal:::renv_lockfile` then.

R/teal_lockfile.R
gogonzo: I think we can give up and try something in the future if this problem occurs. I guess it will be very rare to change working directory especially on deployed app.

m7pr: Cool, this would only require use to orchestra paths in `renv::snapshot()` call. So there is a way to include a custom path in there if there is a need in the futrue https://github.com/insightsengineering/teal/pull/1263/files#diff-c389a151f2967297002c1253ee7819f95d9017f250147356c6e8d6d66165638cR79

m7pr:
I see some errors on vignettes creations during R CMD CHECK
https://github.com/insightsengineering/teal/actions/runs/9956572489/job/27506766811#step:23:35

```r
Quitting from lines 106-112 [unnamed-chunk-4] (adding-support-for-reporting.Rmd)
Error: processing vignette 'adding-support-for-reporting.Rmd' failed with diagnostics:
Don't know how to convert object of class NULL into a promise
--- failed re-building ‘adding-support-for-reporting.Rmd’

--- re-building ‘bootstrap-themes-in-teal.Rmd’ using rmarkdown
--- finished re-building ‘bootstrap-themes-in-teal.Rmd’

--- re-building ‘creating-custom-modules.Rmd’ using rmarkdown

Quitting from lines 174-183 [unnamed-chunk-4] (creating-custom-modules.Rmd)
Error: processing vignette 'creating-custom-modules.Rmd' failed with diagnostics:
Don't know how to convert object of class NULL into a promise
--- failed re-building ‘creating-custom-modules.Rmd’

--- re-building ‘data-as-shiny-module.Rmd’ using rmarkdown

Quitting from lines 40-68 [unnamed-chunk-2] (data-as-shiny-module.Rmd)
Error: processing vignette 'data-as-shiny-module.Rmd' failed with diagnostics:
Don't know how to convert object of class NULL into a promise
--- failed re-building ‘data-as-shiny-module.Rmd’

--- re-building ‘filter-panel.Rmd’ using rmarkdown
--- finished re-building ‘filter-panel.Rmd’

--- re-building ‘getting-started-with-teal.Rmd’ using rmarkdown

Quitting from lines [33](https://github.com/insightsengineering/teal/actions/runs/9956572489/job/27506766811#step:23:34)-52 [unnamed-chunk-1] (getting-started-with-teal.Rmd)
Error: processing vignette 'getting-started-with-teal.Rmd' failed with diagnostics:
Don't know how to convert object of class NULL into a promise
--- failed re-building ‘getting-started-with-teal.Rmd’
```

R/teal_lockfile.R
```diff
@@ -23,92 +44,141 @@
...
-    if (inherits(old_plan, "sequential")) {
-      future::plan(future::multisession, workers = 2)
-    }
-
-    lockfile_task <- ExtendedTask$new(create_renv_lockfile)
-    lockfile_task$invoke(close = inherits(old_plan, "sequential"), lockfile_path)
-    logger::log_trace("lockfile creation invoked.")
+  run_renv_mirai <- function(lockfile_path, opts, sysenv, libpaths, wd) {
+    mirai::mirai(
+      renv_snapshot(
```
gogonzo, line 77: how this suppose to work in a vanilla R session? `renv_snapshot` doesn't exist in any public namespace. This is why `run` was substituted in previous commits

R/teal_lockfile.R
```diff
@@ -23,92 +44,141 @@
...
+        lockfile_path = lockfile_path,
+        opts = opts,
+        sysenv = sysenv,
+        libpaths = libpaths,
+        wd = wd
+      ),
```
gogonzo, line 83:
```suggestion
      )
```

gogonzo (commented):
Can you confirm it works in your case @m7pr 
image

R/teal_lockfile.R
m7pr: got it! because it is started in a new thread and `renv_snapshot` is not visible there as we are no longer in `teal` namespace.

m7pr: now I understand the point of this comment https://github.com/insightsengineering/teal/pull/1263#issuecomment-2230711936

m7pr: `mirai::mirai` need to have arguments passed through `...` or `.args` so they are visible in `.expr` argument. all is clear now

R/teal_lockfile.R
```diff
@@ -23,92 +44,139 @@
...
     }
+  )
 
-    lockfile_task <- ExtendedTask$new(create_renv_lockfile)
-    lockfile_task$invoke(close = inherits(old_plan, "sequential"), lockfile_path)
-    logger::log_trace("lockfile creation invoked.")
-  }
+  suppressWarnings({ # 'package:stats' may not be available when loading
+    process$invoke(
+      run = teal:::renv_snapshot,
```
chlebowa, line 86: Remember, CRAN doesn't take kindly to using `:::`.
It's unclear to my why it is neede, though, so maybe you have exhausted other options.

m7pr: Yeah, there is even our own R CMD CHECK Notes blocker that does not allow that

```r
* checking dependencies in R code ... NOTE
There are ::: calls to the package's namespace in its code. A package
  almost never needs to use ::: for its own objects:
  ‘renv_snapshot’
```

I was trying to overcome different issue with https://github.com/insightsengineering/teal/pull/1263#issuecomment-2230583352
Let's try if we can pass this directly without the specs of the namespace

m7pr (dismissed):
Hey @gogonzo this is really solid work. If ever any user needs customization for the apps deployed in non-working directories, it will be easy to extend the current approach.

I needed to extend function definition in `ExtendTask$new(` by direct parameter names to overcome notes in R CMD CHECK

R/teal_lockfile.R
```diff
@@ -23,92 +44,139 @@
...
 
-  if (!(is_in_test() || is_r_cmd_check())) {
-    old_plan <- future::plan()
-    # If there is already a parallel (non-sequential) backend, reuse it.
-    if (inherits(old_plan, "sequential")) {
-      future::plan(future::multisession, workers = 2)
+  process <- ExtendedTask$new(
+    function(run, lockfile_path = lockfile_path, opts = opts, sysenv = sysenv, libpaths = libpaths, wd = wd) {
+      mirai::mirai(
+        run(lockfile_path = lockfile_path, opts = opts, sysenv = sysenv, libpaths = libpaths, wd = wd),
```
pawelru, line 78: can we avoid `run` argument here and call `renv_snapshot()` directly? If `:::` is a problem then let's try `getFromNamespace()` trick. This will slightly reduce complexity.

R/teal_lockfile.R
```diff
@@ -23,92 +44,139 @@
...
+    length = length(renv$Packages)
+  )
+}
+
+#' @rdname teal_lockfile
+#' @keywords internal
+teal_lockfile_tracker <- function(process) {
+  checkmate::assert_class(process, "ExtendedTask", null.ok = TRUE)
+  if (inherits(process, "ExtendedTask")) {
+    tracker <- observeEvent(process$status(), {
```
pawelru, line 135: variable defined but not used
```suggestion
    observeEvent(process$status(), {
```

R/teal_lockfile.R
```diff
@@ -23,92 +44,139 @@
...
     }
 
-    lockfile_path
-  })
-  if (close) {
-    # If the initial backend was only sequential, bring it back.
-    promises::then(promise, onFulfilled = function() {
-      future::plan(future::sequential)
-    })
+    logger::log_trace("Lockfile {renv_logs$lockfile_path} containing { renv_logs$length } packages created{ with }.")
```
pawelru, line 159: let's assume I got warnings or errors - how can I see them? 
I think we should log the contents of `renv_logs$out`.

R/init.R
```diff
@@ -231,6 +231,7 @@ init <- function(data,
       if (!is.null(landing_module)) {
         do.call(landing_module$server, c(list(id = "landing_module_shiny_id"), landing_module$server_args))
       }
+      teal_lockfile_tracker(process)
```
pawelru, line 234: can we group the related functionalities close together? 

```suggestion
      teal_lockfile_tracker(teal_lockfile_invoke())
```

Alternatively we can have one "main" function
```
      create_lockfile()
```

R/teal_lockfile.R
```diff
@@ -23,92 +44,139 @@
...
       file.copy(user_lockfile, lockfile_path)
-      return(invisible(NULL))
+      logger::log_trace('Lockfile set using option "teal.renv.lockfile" - skipping automatic creation.')
+      return(NULL)
     } else {
       stop("lockfile provided through options('teal.renv.lockfile') does not exist.")
     }
   }
 
-  if (!(is_in_test() || is_r_cmd_check())) {
```
pawelru, line 42: IIRC these functions were created for this and are used only here
if removed - let's remove them as well

m7pr: They are removed here https://github.com/insightsengineering/teal/pull/1263/files#diff-c389a151f2967297002c1253ee7819f95d9017f250147356c6e8d6d66165638cL108-L114

pawelru: ok missed that somehow
closing

R/teal_lockfile.R
```diff
@@ -23,92 +44,139 @@
...
 
-    lockfile_task <- ExtendedTask$new(create_renv_lockfile)
-    lockfile_task$invoke(close = inherits(old_plan, "sequential"), lockfile_path)
-    logger::log_trace("lockfile creation invoked.")
-  }
+  suppressWarnings({ # 'package:stats' may not be available when loading
+    process$invoke(
+      run = renv_snapshot,
+      lockfile_path = lockfile_path,
+      opts = options(),
```
pawelru, line 88: let's take `opts` as a representative but this applies also to `libpaths`, `sysenv` and `wd`

currently this is present in multiple places increasing overall complexity (starting from the bottom):
- argument of `teal:::renv_snapshot()`
- argument of `mirai::mirai` passed to the call
- argument of anonymous function in `ExtendedTask$new`

I think it's a little too much and we would be able to find a way to avoid this. A couple of ideas from my side:
- `mirai::mirai(run = {options(opts); ...; renv_snapshot(lockfile_path)})` - this is to avoid extending `renv_snapshot()`
- `ExtendedTask$new({options(opts); ...; function(lockfile_path){mirai::mirai(renv_snapshot(lockfile_path), lockfile_path))} })`

R/module_teal.R
```diff
@@ -112,7 +112,7 @@ ui_teal <- function(id,
         footer,
         teal.widgets::verbatim_popup_ui(ns("sessionInfo"), "Session Info", type = "link"),
         br(),
-        downloadLink(ns("lockFile"), "Download .lock file"),
+        shinyjs::hidden(downloadLink(ns("lockFile"), "Download .lock file")),
```
pawelru, line 115: It goes a little bit beyond this PR but you actually changes this :)

I think it's a bad user experience to be in the state where you don't know if something is running. We should expose that information somewhere somehow. Right now that's the case - if you don't see the link you don't know if it is still doing snapshot or it's a bug or something else.
I would propose to be more transparent in what's happening. That is: start empty. If the process is invoked - change the status to "running" state. If it's completed - present the download link. If it's error - give this to UI. I hope you get what I mean here.

R/init.R
pawelru: I actually think that introducing yet another level (`create_lockfile`) would be nice to keep file-based-approach separate to the future-based-approach

pseudo-code
```
create_lockfile <- function() {
  if ( file condition ) {
    teal_lockfile_read()
  } else { 
    process <- teal_lockfile_process_invoke()
    teal_lockfile_process_tracker(process)
  }
}
```

Then we can isolate all file based logic from the current `teal_lockfile_invoke` - conceptually there is no need to _invoke_ if file is provided. This would also simplify the current `teal_lockfile_tracker` because you don't have to handle empty `process` argument.

pawelru (commented):
I left a few comments about the implementation but in general I do like this. I like it more that the current approach implemented on main. Mostly because it enables `ExtendedTask` which I believe will be implemented in many places in the future.

R/teal_lockfile.R
m7pr: @pawelru it is because `mirai::mirai` way of passing parameters from the current session to the new session that is starts, which is not ideal of course. Or we could pass the whole environment.

image

I don't think I understand what's going on in here

image

R/module_teal.R
m7pr: You mean, the button that is finally used for the download-handler could have multiple phases/steps: empty, running, ready for download, error?

R/init.R
m7pr: This makes a lot of sense

R/teal_lockfile.R
m7pr: I think we could also decrease

```r
function(run, lockfile_path = lockfile_path, opts = opts, sysenv = sysenv, libpaths = libpaths, wd = wd) 
```

to 

```
function(run, lockfile_path, opts, sysenv, libpaths, wd) 
```

pawelru:
I opened a PR that address most of my comments: https://github.com/insightsengineering/teal/pull/1272

R/teal_lockfile.R
pawelru: https://github.com/insightsengineering/teal/pull/1272/commits/5349788da901388082c561ceddb4a72c744c4746

DESCRIPTION
```diff
@@ -40,14 +40,13 @@ Depends:
...
+    mirai,
     checkmate (>= 2.1.0),
-    future (>= 1.33.2),
     jsonlite,
     lifecycle (>= 0.2.0),
     logger (>= 0.2.0),
     magrittr (>= 1.5),
     methods,
-    promises (>= 1.3.0),
     renv (>= 1.0.7),
```
pawelru, line 50: how about making `renv` a (true) soft dependency and fail gracefully if not available?

m7pr: Also @gogonzo is there any specific minimal `mirai` version that we support? Should we go with the newest CRAN release? 1.1.1? https://cran.r-project.org/web/packages/mirai/index.html

gogonzo: Hard to determine how `teal` would behave with previous versions. It is safe to set to `(>= latest version)` especially that even now it doesn't have anything in reverse dependencies which might change our reverse dependencies.

.pre-commit-config.yaml
```diff
@@ -17,14 +17,13 @@ repos:
...
           - davidgohel/flextable # Error: package 'flextable' is not available
           - davidgohel/gdtools # for flextable
+          - mirai
           - checkmate
-          - future
           - jsonlite
           - lifecycle
           - logger
           - magrittr
           - methods
```
gogonzo, line 26: alphabetical order
```suggestion
          - checkmate
          - jsonlite
          - lifecycle
          - logger
          - magrittr
          - methods
          - mirai
```

DESCRIPTION
pawelru: Please kindly open a new thread for a separate issue. It will help to organise the discussion in a clean way. Here, I want to discuss whether renv can be handled as a true soft dependency.

R/teal_lockfile.R
```diff
@@ -23,92 +44,154 @@
...
+      warning("Lockfile creation failed.")
+      shinyjs::html("teal-lockFileStatus", "Lockfile creation failed.")
+      shinyjs::disable("teal-lockFileLink")
+    }
+  })
+}
+
+#' @rdname teal_lockfile
+#' @keywords internal
 teal_lockfile_downloadhandler <- function() {
```
gogonzo, line 185: IMO separation of shiny namespaces is hard to maintain and since we do `disable(teal-lockFileLink)` and `shinyjs::html("teal-lockFileStatus)` we are risking to break this with any changes in the namespace in teal. We would need to be very careful with this in a future.

I think if we pass `lockfile_process` to the module we would be able to encapsulate everything in a single shiny namespace and we wouldn't have process split into multiple parts. 
```suggestion
ui_teal_lockfile <- function(id) {...}
srv_teal_lockfile <- function(id, process) {
  moduleServer(id, function(input, output, session) {
      if (!is.null(process)) {
        teal_lockfile_process_tracker(process)
      }
      
      ...
  })
}
```

The only problem is how to handover `lockfile_process` - either by argument `srv_teal_with_splash(..., process)` -> `srv_teal(..., process)` or by `session$userData$lockfile_process`. I prefer `userData` as `srv_teal` will be exported in the next release and we don't want to complicate public api.

gogonzo:
After discussion with @ruckip we decide to:

- encapsulate everything in `ui_renv_lockfile` and `srv_renv_lockfile`
- move envocation of the lockfile process to the `srv_renv_lockfile` which should be conditional - if file already exist, don't start the process again. It means it will no longer be in `teal::init`, so we don't need to pass `lockfile_process` to the `srv_teal`
- we need an `option(teal.renv.enable = TRUE/FALSE)` to enable/disable lockfile calculation

m7pr:
> move envocation of the lockfile process to the srv_renv_lockfile which should be conditional - if file already exist, don't start the process again. It means it will no longer be in teal::init, so we don't need to pass lockfile_process to the srv_teal

> It means it will no longer be in teal::init, so we don't need to pass lockfile_process to the srv_teal

I thought we wanted to have lockfile creation in `init` so it is created once per app, and not per each session. Is it still relevant?

DESCRIPTION
```diff
@@ -41,13 +41,12 @@ Depends:
...
 Imports:
     checkmate (>= 2.1.0),
-    future (>= 1.33.2),
     jsonlite,
     lifecycle (>= 0.2.0),
     logger (>= 0.2.0),
     magrittr (>= 1.5),
     methods,
-    promises (>= 1.3.0),
+    mirai,
```
m7pr, line 49:
```suggestion
    mirai (>= 1.1.1),
```

DESCRIPTION
m7pr: @pawelru if we want to make `renv` a soft dependency then I would suggest making `mirai` a soft dependency as well. What is the point of having a teal installation where `mirai` gets installed by default but `renv`not? in our case `mirai` only make sense with a combination with `renv`

pawelru: In the current state you are definitely right and it feels natural to move `mirai` to soft dep as well.
However, there is a slight difference between those two and it's all about it's potential for the future development. For `mirai` I do see a lot of other use cases (e.g. teal.slice, teal.transform, teal.code and even the module codes) as opposed to `renv`. I see this PR as a very first implementation of async that should serve as an example for the future. If we move `mirai` to soft right now - I am afraid that we will be reverting this sooner or later.

m7pr: but we are not reverting yet. this PR is still a proposition, so we can propose to have it in the soft deps and THEN improve to become a regular dependency if we ever figure out more use cases

pawelru: > yet

That's exactly what I was referring to :) As a principle, it's hard to disagree with you. I just think that it will happen sooner than later but I'm also happy to do as you suggested.

R/teal_lockfile.R
```diff
@@ -23,92 +24,149 @@
...
+    process <- .teal_lockfile(lockfile_path)
+
+    observeEvent(process$status(), {
+      if (process$status() == "initial" || process$status() == "running") {
+        shinyjs::html("lockFileStatus", "Creating lockfile...")
+      } else if (process$status() == "success") {
+        result <- process$result()
+        if (any(grepl("Lockfile written to", result$out))) {
+          logger::log_debug("Lockfile {result$path} containing { length(result$res$Packages) } packages created.")
+          if (any(grepl("WARNING:", result$out)) || any(grepl("ERROR:", result$out))) {
```
m7pr, line 58:
```suggestion
          if (any(grepl("WARNING|ERROR:", result$out))) {
```

gogonzo:
I've found a serious problem when invoked ExtendedTask in tests. There is a long queue of running processes which are being cumulated by each `srv_teal` call. I'm trying to find a way to terminate on `shiny::onStop`.

m7pr:
We can introduce the same mechanism as is currently on main - to skip lockfile creation in tests and r cmd check. Those are the environments that we do not control. We invented that because during tests on GitHub Actions some connections were not closed, even though the jobs were finished.

https://github.com/insightsengineering/teal/blob/6fc542196413d4db8f328d84e623915df4b91547/R/teal_lockfile.R#L108-L114

https://github.com/insightsengineering/teal/blob/6fc542196413d4db8f328d84e623915df4b91547/R/teal_lockfile.R#L42

gogonzo:
> We can introduce the same mechanism as is currently on main - to skip lockfile creation in tests and r cmd check. Those are the environments that we do not control. We invented that because during tests on GitHub Actions some connections were not closed, even though the jobs were finished.
> 
> https://github.com/insightsengineering/teal/blob/6fc542196413d4db8f328d84e623915df4b91547/R/teal_lockfile.R#L108-L114
> 
> https://github.com/insightsengineering/teal/blob/6fc542196413d4db8f328d84e623915df4b91547/R/teal_lockfile.R#L42


This will disable lockfile from running in tests. But how to test this feature?

m7pr:
> This will disable lockfile from running in tests. But how to test this feature?

A separate GitHub action job, that doesn't run `devtools::test()` but runs an R script and reviews results based on some assumptions and checks?

R/teal_lockfile.R
```diff
@@ -45,34 +45,69 @@ ui_teal_lockfile <- function(id) {
...
+        logger::log_debug("Removing lockfile after shutting down the app")
+        file.remove(lockfile_path)
+      }
+    })
+
+    # run renv::snapshot only once per app. Once calculated available for all users
     lockfile_path <- "teal_app.lock"
-    process <- .teal_lockfile(lockfile_path)
+    if (file.exists(lockfile_path)) {
+      logger::log_debug("Lockfile have been already created - skipping automatic creation.")
```
m7pr, line 68:
```suggestion
      logger::log_debug("Lockfile has been already created - skipping automatic creation.")
```

R/teal_lockfile.R
```diff
@@ -45,34 +45,69 @@ ui_teal_lockfile <- function(id) {
...
+    # don't run renv::snapshot when option is set
+    user_lockfile <- getOption("teal.renv.lockfile", "")
+    if (!identical(user_lockfile, "")) {
+      if (file.exists(user_lockfile)) {
+        file.copy(user_lockfile, lockfile_path)
+        logger::log_debug('Lockfile set using option "teal.renv.lockfile" - skipping automatic creation.')
+        enable_lockfile_download()
+        return(NULL)
+      } else {
+        warning("lockfile provided through options('teal.renv.lockfile') does not exist.", call. = FALSE)
```
m7pr, line 82:
```suggestion
        warning("Lockfile provided through options('teal.renv.lockfile') does not exist.", call. = FALSE)
```

R/teal_lockfile.R
```diff
@@ -91,58 +126,32 @@ srv_teal_lockfile <- function(id) {
...
+    libpaths = .libPaths(),
+    sysenv = as.list(Sys.getenv()),
+    wd = getwd()
   )
+  shiny::onStop(function() {
+    if (mirai::unresolved(mirai_obj)) {
+      logger::log_debug("Terminating a running lockfile process...")
+      mirai::stop_mirai(mirai_obj) # this doesn't stop running - renv will be created even if session is closed
+    }
+  })
```
m7pr, line 153: Well done on this one!

m7pr:
Hey @gogonzo getting back to this. Do you think this https://github.com/insightsengineering/teal/pull/1263#issuecomment-2288467032 is a blocker for `mirai`? We can always revisit the approach with `callr` and figure out why it didn't work on macOS

DESCRIPTION
pawelru: Hey @gogonzo 
Following up on this thread. If kept in Suggests - we need to exit nicely if `renv` (as well as `mirai` as @m7pr pointed out) is not installed.

gogonzo: Any warning message if they are not installed?

pawelru: Yes please

R/module_teal_lockfile.R
```diff
@@ -0,0 +1,195 @@
...
+    tags$span("", id = ns("lockFileStatus")),
+    shinyjs::disabled(downloadLink(ns("lockFileLink"), "Download lockfile"))
+  )
+}
+
+#' @rdname module_teal_lockfile
+srv_teal_lockfile <- function(id) {
+  if (!isTRUE(getOption("teal.renv.enable")) || !.is_lockfile_deps_installed()) {
+    return(NULL)
+  }
```
pawelru, line 55: there is one more case:
enabled = TRUE
user provided file
suggested packages not installed

R/module_teal_lockfile.R
```diff
@@ -0,0 +1,195 @@
...
+NULL
+
+#' @rdname module_teal_lockfile
+ui_teal_lockfile <- function(id) {
+  if (!isTRUE(getOption("teal.renv.enable"))) {
+    return(NULL)
+  } else if (!.is_lockfile_deps_installed()) {
+    warning("Lockfile feature disabled. `mirai` and `renv` packages must be installed.")
+    return(NULL)
+  }
```
pawelru, line 42: I think it would be better to keep ui independent (i.e. always return disabled link) and handle soft dependencies in server only.
- enable option at the beggining where it is currently
- soft deps in `.teal_lockfile_process_invoke` - i.e. return empty `process` object if any package is missing

R/module_teal_lockfile.R
gogonzo: Thanks, good point 👍

R/module_teal_lockfile.R
gogonzo: > I think it would be better to keep ui independent (i.e. always return disabled link)

I don't like this proposition. I don't see the reason why app-user would like to see disabled button when packages are missing. What if `teal.lockfile.enable = FALSE`, should disabled download button displayed also? 

I don't see the point to run `server` and make some extra `if`s there just to disable download button. Creating empty `process` will require specific code to handle this "exception". I consider this too complicated (more complicated that simple ifs in both UI and server). This is how I view this now.

image

pawelru: I was more concerned about messy implementation of multiple ifs here and there but if you put this in that way (I mean the decision tree from the above) then it's going to be good. I think it's connected with https://github.com/insightsengineering/teal/pull/1263/files#r1745602447 so once implemented let me have another look at this.

R/module_teal_lockfile.R
```diff
@@ -0,0 +1,195 @@
...
+  shiny::tagList(
+    tags$span("", id = ns("lockFileStatus")),
+    shinyjs::disabled(downloadLink(ns("lockFileLink"), "Download lockfile"))
+  )
+}
+
+#' @rdname module_teal_lockfile
+srv_teal_lockfile <- function(id) {
+  if (!isTRUE(getOption("teal.renv.enable")) || !.is_lockfile_deps_installed()) {
+    return(NULL)
```
pawelru, line 54: it would be good to put something in the logs
we have a few exit criteria and it would be good if we can identify which one occured in case of issues

R/module_teal_lockfile.R
```diff
@@ -0,0 +1,195 @@
...
+
+    # run renv::snapshot only once per app. Once calculated available for all users
+    if (file.exists(lockfile_path)) {
+      logger::log_debug("Lockfile has been already created - skipping automatic creation.")
+      enable_lockfile_download()
+      return(NULL)
+    }
+
+    # don't run renv::snapshot when option is set
+    user_lockfile <- getOption("teal.renv.lockfile", "")
```
pawelru, line 85: When I reading this again I started to think that the option is too complex - it consists of two technical terms: `renv` as well as `lockfile`. I think we can do simpler:
`teal.renv.lockfile` -> `teal.lockfile.path`
`teal.renv.enable` -> `teal.lockfile.enable`

The rationale is as follows:
- one technical term `lockfile` instead of two
- `renv` is irrelevant here - it should not patter how (which package) it is implemented

Please let me know your thoughts on that.

gogonzo: Makes absolute sense 👍

R/module_teal_lockfile.R
gogonzo: I'll move UI to renderUI and keep only ifs only on the server side 👍

gogonzo:
@pawelru do you want to look at this again?

R/module_teal_lockfile.R
```diff
@@ -0,0 +1,196 @@
...
+      }
+    })
+
+    lockfile_path <- "teal_app.lock"
+    is_lockfile_enabled <- isTRUE(getOption("teal.lockfile.enable"))
+    user_lockfile_path <- getOption("teal.lockfile.path", default = "")
+    is_user_lockfile_set <- !identical(user_lockfile_path, "")
+
+    if (!is_lockfile_enabled) {
+      logger::log_debug("'teal.lockfile.enable' option is set to false. Hiding a lockfile download button.")
```
m7pr, line 82:
```suggestion
      logger::log_debug("'teal.lockfile.enable' option is set to FALSE. Hiding a lockfile download button.")
```

R/zzz.R
```diff
@@ -1,13 +1,15 @@
 .onLoad <- function(libname, pkgname) {
   # adapted from https://github.com/r-lib/devtools/blob/master/R/zzz.R
-  teal_default_options <- list(teal.show_js_log = FALSE)
+  teal_default_options <- list(
+    teal.show_js_log = FALSE,
+    teal.lockfile.enable = TRUE,
+    shiny.sanitize.errors = FALSE
```
pawelru, line 6: why this?

pawelru: ahh its moved from below
lets leave it then

tests/testthat/setup-options.R
```diff
@@ -7,6 +14,7 @@ opts_partial_match_old <- list(
   warnPartialMatchAttr = getOption("warnPartialMatchAttr")
 )
 opts_partial_match_new <- list(
+  teal.lockfile.enable = FALSE,
```
pawelru, line 17: - please move it elsewhere (I mean outside of `opts_partial_match_new` as it has some meaning)
- coming back to [this](https://github.com/insightsengineering/teal/issues/1276#issuecomment-2288684906) - can we auto-dected that we are in the tests so that there is no need to set this option? This is relevant for tests outside of `teal` - e.g. `teal.modules.general`. I would like to avoid mass update all downstream packages to set this option in the tests - if possible of course.

vignettes/teal-options.Rmd
```diff
@@ -84,10 +84,16 @@ This indicates whether to print the `JavaScript` console logs to the `R` console
 
 Default: `FALSE`.
 
-### `teal.renv.lockfile` (`character`)
+### `teal.lockfile.enable` (`logical`)
 
-The path to the pre-computed `renv` lockfile that will be shared through teal app. To read more about lockfile usage creation check `?teal::teal_lockfile`.
+This enables app users to download `renv` lockfile by clicking "download lockfile" in the footer. If
```
m7pr, line 89:
```suggestion
This enables app users to download `renv` lockfile by clicking `"download lockfile"` in the footer of the app. If 
```

tests/testthat/setup-options.R
m7pr: I think for tests you can detect if you are inside them with

```r

is_in_test <- function() {
  identical(Sys.getenv("TESTTHAT"), "true")
}
```

vignettes/teal-options.Rmd
```diff
@@ -84,10 +84,16 @@ This indicates whether to print the `JavaScript` console logs to the `R` console
 
 Default: `FALSE`.
 
-### `teal.renv.lockfile` (`character`)
+### `teal.lockfile.enable` (`logical`)
 
-The path to the pre-computed `renv` lockfile that will be shared through teal app. To read more about lockfile usage creation check `?teal::teal_lockfile`.
+This enables app users to download `renv` lockfile by clicking "download lockfile" in the footer. If 
+`teal.lockfile.path` is not specified then lockfile is created automatically by `teal`. To read more about 
+lockfile usage creation check `?teal::teal_lockfile`.
```
m7pr, line 91:
```suggestion
lockfile usage creation check `?teal::module_teal_lockfile`.
```

vignettes/teal-options.Rmd
```diff
@@ -84,10 +84,16 @@ This indicates whether to print the `JavaScript` console logs to the `R` console
...
 
+### `teal.lockfile.path` (`character`)
+
+The path to the pre-computed `renv` lockfile linked by app developer. Lockfile will be available to download if 
+`teal.lockfile.enable` is set to `TRUE`.
```
m7pr, line 96: Might be redundant as it's already listed in the previous option, but if someone just looks inside this option then this might be helpful
```suggestion
`teal.lockfile.enable` is set to `TRUE`. To read more about lockfile usage creation check `?teal::module_teal_lockfile`.
```

R/module_teal_lockfile.R
```diff
@@ -0,0 +1,196 @@
...
+#'      working directory.
+#'     - **`DESCRIPTION`-based lockfile**: To generate a lockfile based on a `DESCRIPTION` file in your working
+#'     directory, set `renv::settings$snapshot.type("explicit")`. The naming convention for `type` follows
+#'     `renv::snapshot()`. For the `"explicit"` type, refer to `renv::settings$package.dependency.fields()` for the
+#'     `DESCRIPTION` fields included in the lockfile.
+#'     - **Custom files-based lockfile**: To specify custom files as the basis for the lockfile, set
+#'     `renv::settings$snapshot.type("custom")` and configure the `renv.snapshot.filter` option.
+#'
+#' @section lockfile usage:
+#' After creating the lockfile, you can restore the application environment using `renv::restore()`.
```
m7pr, line 23:
```suggestion
#' After creating the lockfile, you can restore the application's environment using `renv::restore()`.
```

tests/testthat/setup-options.R
m7pr: R CMD check, for which we would also like to disable this option automatically

```r
is_r_cmd_check <- function() {
  ("CheckExEnv" %in% search()) || any(c("_R_CHECK_TIMINGS_", "_R_CHECK_LICENSE_") %in% names(Sys.getenv()))
}
```

m7pr: But using `is_in_test` and `is_r_cmd_check` is not ideal as we won't be able to test this functionality at all. Setting an option in tests in other packages allow us to test this functionality in `teal`

chlebowa: > I think for tests you can detect if you are inside them with
> 
> ```r
> is_in_test <- function() {
>   identical(Sys.getenv("TESTTHAT"), "true")
> }
> ```

Have a look at `testthat::is_testing`.


> R CMD check, for which we would also like to disable this option automatically
> 
> ```r
> is_r_cmd_check <- function() {
>   ("CheckExEnv" %in% search()) || any(c("_R_CHECK_TIMINGS_", "_R_CHECK_LICENSE_") %in% names(Sys.getenv()))
> }
> ```

Checking for presence of env vars rather than their values may be risky because someone may choose to set them to `""` rather than unsetting them.

m7pr: Usage of `testthat::is_testing` would require putting `testthat` in Imports. The code that I put for `is_in_test` is actually the body of `testthat::is_testing`.

When it comes to those env vars - I hope they are created during R CMD CHECK and then they are unset. I don't think this is something a regular user would set.

Anyway I think this is not good practice to omit those in tests and R CMD CHECKS by design as there is no way to test this functionality. The option is better way, as can be turned off for any custom package, but not this one.

chlebowa: Fair enough 👍  
Just to be clear, `is_testing` actually does `isTRUE(as.logical(Sys.getenv("TESTTHAT")))`.

vignettes/teal-options.Rmd
pawelru: I would discourage usage of `"app"` in the documentation - use `"application"` instead

tests/testthat/setup-options.R
m7pr: My testthat version does this

```r
> testthat::is_testing
function () 
{
    identical(Sys.getenv("TESTTHAT"), "true")
}
```

chlebowa: My mistake.
```
> testthat:::env_var_is_true
function (x) 
{
    isTRUE(as.logical(Sys.getenv(x, "false")))
}
<bytecode: 0x558a1bf24230>
<environment: namespace:testthat>
```

m7pr: For other processes: `callr` sets `CALLR_IS_RUNNING` variable.


```r
library(callr)

process <- callr::r_bg(func = function() print(Sys.getenv()))
output <- process$read_output()
env <- strsplit(output, split = "\r\n")[[1]]
names_in_callr <- unique(gsub(" .*$", "", env))
setdiff(names_in_callr, names(Sys.getenv()))

Sys.getenv("CALLR_IS_RUNNING")
process <- callr::r_bg(func = function() print(Sys.getenv("CALLR_IS_RUNNING")))
output <- process$read_output()
output

[1] "CALLR_IS_RUNNING"                
[2] ""                                
[3] "CYGWIN"                          
[4] "PROCESSX_PSLQ3SBE3NYC_1726557642"
[5] "R_BROWSER"                       
[6] "R_PDFVIEWER"
```


`reprex` runs via `callr` : )

image

R/module_teal_lockfile.R
```diff
@@ -0,0 +1,196 @@
...
+        disable_lockfile_download()
+      }
+    })
+
+    NULL
+  })
+}
+
+utils::globalVariables(c("opts", "sysenv", "libpaths", "wd", "lockfilepath", "run")) # needed for mirai call
+#' @rdname module_teal_lockfile
```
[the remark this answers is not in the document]
gogonzo: > This exposes those functions prefixed with . in the documentation

This is not a public documentation. This is for devs. I think it is nice to list all the functions related to the certain topic

R/module_teal_lockfile.R
```diff
@@ -0,0 +1,196 @@
...
+      logger::log_debug("'teal.lockfile.enable' option is set to false. Hiding a lockfile download button.")
+      shinyjs::hide("lockFileLink")
+      return(NULL)
+    }
+
+    if (file.exists(lockfile_path)) {
+      logger::log_debug("Lockfile has already been created for this app - skipping automatic creation.")
+      enable_lockfile_download()
+      return(NULL)
+    }
```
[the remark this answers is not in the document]
gogonzo: I don't think the suggestion makes sense. `lockfile_path` is the location of the file created for the current session. This includes also `teal.lockfile.path` provided by the user. When app developer provides `teal.lockfile.path` then it is copied to `lockfile_path` for the session runtime. Thanks to this `if` teal avoids recalculating/re-copying lockfile which already has been set for the app.

vignettes/teal-options.Rmd
m7pr: updated the comment

R/module_teal_lockfile.R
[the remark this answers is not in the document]
m7pr: Ok, so if this was intended then this is fine.
There is public instruction in vignettes to check this page https://github.com/insightsengineering/teal/pull/1263#discussion_r1760819539

But it's probably more for teal app developers not teal app users

R/module_teal_lockfile.R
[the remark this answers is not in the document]
m7pr: I meant the case where

`teal.lockfile.path` points to a file that does not exist AND file called `teal_app.lock` accidentally exists in the directory but it's not the file you want to be the lockfile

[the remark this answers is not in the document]
m7pr:
```r
library(teal)
# this is the file that accidentally existed in the directory
file.create('teal_app.lock') 

# this is the file you would like to share, but you made a typo and this file does not exist
options(teal.lockfile.path = "file_not_exists.lock")

app <- init(
  modules = example_module()
)

shinyApp(app$ui, app$server)

```

image

tests/testthat/setup-options.R
gogonzo: @pawelru I've updated [a comment](https://github.com/insightsengineering/teal/issues/1276#issuecomment-2288684906). I think if we are about to disable lockfile generation in `callr` process then we have to do it always (not conditional on flag set in the parent process). If this is what you expect then we can continue

R/module_teal_lockfile.R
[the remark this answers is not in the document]
gogonzo: > teal.lockfile.path points to a file that does not exist AND file called teal_app.lock accidentally exists in the directory but it's not the file you want to be the lockfile

`teal_app.lock` is created in the internal `teal` process. I've removed one occurrence from the documentation so nobody should teoreticaly create this file manually. To provide `lockfile` we have an option `teal.lockfile.path` only.

tests/testthat/setup-options.R
m7pr: If we want to disable lockfile generation for `callr`, `testthat::tests` and `R CMD CHECK` no matter what's the parent flag, then this needs to be added to `zzz.R -> onLoad`

```r
  lockfile_flag <-
    !(
      identical(Sys.getenv("CALLR_IS_RUNNING"), "true") || # inside callr process
      identical(Sys.getenv("TESTTHAT"), "true") || # inside devtools::test
      (
        ("CheckExEnv" %in% search()) || any(c("_R_CHECK_TIMINGS_", "_R_CHECK_LICENSE_") %in% names(Sys.getenv()))
      ) # inside R CMD CHECK
    )

  teal_default_options <- list(
    teal.show_js_log = FALSE,
    teal.lockfile.enable = lockfile_flag,
    shiny.sanitize.errors = FALSE
  )
```

I made a PR for this in here https://github.com/insightsengineering/teal/pull/1346

If we want to pay attention to the parent flag, then below code 
```r
  lockfile_flag <-
    !(
      identical(getOption("teal.lockfile.enable", ""), "") && # option is not set
      (
        identical(Sys.getenv("CALLR_IS_RUNNING"), "true") || # inside callr process
        identical(Sys.getenv("TESTTHAT"), "true") || # inside devtools::test
        (
          ("CheckExEnv" %in% search()) || any(c("_R_CHECK_TIMINGS_", "_R_CHECK_LICENSE_") %in% names(Sys.getenv()))
        ) # inside R CMD CHECK
      )
    )


  teal_default_options <- list(
    teal.show_js_log = FALSE,
    teal.lockfile.enable = lockfile_flag,
    shiny.sanitize.errors = FALSE
  )
```

DESCRIPTION
m7pr: We introduced

```r
.is_lockfile_deps_installed <- function() {
  requireNamespace("mirai", quietly = TRUE) && requireNamespace("renv", quietly = TRUE)
}
```

closing this now

tests/testthat/setup-options.R
m7pr: @pawelru this should be satisfied in this PR https://github.com/insightsengineering/teal/pull/1263/commits/761dd68e10426c451ba9eb1c0b8d2f40b0ead583

gogonzo:
@pawelru this is ready ⭐

vignettes/teal-options.Rmd
```diff
@@ -84,10 +84,19 @@ This indicates whether to print the `JavaScript` console logs to the `R` console
 
 Default: `FALSE`.
 
-### `teal.renv.lockfile` (`character`)
+### `teal.lockfile.enable` (`logical`)
 
-The path to the pre-computed `renv` lockfile that will be shared through teal app. To read more about lockfile usage creation check `?teal::teal_lockfile`.
+This enables app users to download `renv` lockfile by clicking "download lockfile" in the footer. 
+If `teal.lockfile.path` is not specified then lockfile is created automatically by `teal`.
```
pawelru, line 90: to me, this sentence belongs more into the `teal.lockfile.path` option description rather than `teal.lockfile.enable`

R/module_teal_lockfile.R
```diff
@@ -0,0 +1,212 @@
...
+    !(
+      identical(Sys.getenv("CALLR_IS_RUNNING"), "true") || # inside callr process
+        identical(Sys.getenv("TESTTHAT"), "true") || # inside devtools::test
+        !identical(Sys.getenv("QUARTO_PROJECT_ROOT"), "") || # inside Quarto process
+        (
+          ("CheckExEnv" %in% search()) || any(c("_R_CHECK_TIMINGS_", "_R_CHECK_LICENSE_") %in% names(Sys.getenv()))
+        ) # inside R CMD CHECK
+    )
+  }
+}
```
pawelru, line 212: I have some concerns on how the `teal.lockfile.enable` option is used. 
- This is not set by us and it would probably be never set by the end-users.
- If I set it to FALSE then (assuming all conditions are not met) I still might end up with lockfile created

How about below: three possible option values: true, auto, false and on load we set it to auto.

vignettes/teal-options.Rmd
gogonzo: This is difficult as lockfile-process is joint dependent on both options. Therefore I suggest to have a `teal.lockfile` section with both options enumerated within the section.

R/module_teal_lockfile.R
gogonzo: @pawelru good point

```r
if (<is set>) {
  getOption("teal.lockfile.enable")
} else {
  ...
}
```

pawelru: Some additional thoughts:

- `"enabled"` option name typically denotes boolean with true/false choices so from that perspective it would make sense to _merge_ the "true" and "auto" value

- if we want to continue with this we can have the following values: auto, user, false. That option should be named more as a "mode" instead

- why do we need the true value? I think we are more interested in false instead. If that's for testing then the answer is to temporarily mock that `.is_lockfile_enabled()` function (and not reverse engineer options values)

vignettes/teal-options.Rmd
gogonzo: Yes, I had this dilemma where to put what. You see, lockfile-process i joint-dependent on both options and it make sense to make `teal.lockfile` section instead and enumerate `teal.lockfile.path` and `teal.lockfile.enable` within this section. I decided not to make the separate section as the whole vignette is constructed in a way that each option has own section with `<option> <value class>` for example `teal.lockfile.path character`

R/module_teal_lockfile.R
gogonzo: It make sense to 
```
if (<option is set>)
  getOption("teal.lockfile.enable")
} else {
```


Just saying (and it is a general comment not specifically about this solution) that setting option in the parent process doesn't always guarantee that all children will preserve an option. It depends on a child process (e.g. `run_examples(fresh = TRUE)` function `build_vignettes()`) passes options further.

pawelru: So this option is definitely needed to allow for explicit user disabling if one does not want it. So we need the "false" value. 
I just can't imagine a user (!) scenario where the "true" value would be used. This value would indicate more or less the following: ignore all of those checks that we came up with and create lockfile no matter what. These checks suppose to eliminate scenarios that occurs only within package development. A "standard" usage should not be covered by these. On the other hand - we can make mistakes in our checks so this mechanism opens the doors for overwrittes. So maybe it's not that bad.
I can image a scenario where "true" is needed - our tests. But like I said earlier - if we have that logic niceley encapsulated in one function `.is_lockfile_enabled()` - we should mock that fun instead. 

Re interdependent options - I started to like `user` option value. Currently we have a following logic: if user provides a file path and such file does not exist (or inaccessible) then fallback to another option value checks. This is probably a little bit too convoluted and also difficult to implement and document. A "user" option value (alongside another path option value) makes it clear that I want user provided file (and don't have fallback mechanism). We can easily return from [here](https://github.com/insightsengineering/teal/blob/f76a30ff4403e45e5ff18fcc9ccea05b84d00d72/R/module_teal_lockfile.R#L104-L106). 
I believe that this would make it more explicit and simpler. Especially for the user.

m7pr: > Re interdependent options - I started to like user option value. Currently we have a following logic: if user provides a file path and such file does not exist (or inaccessible) then fallback to another option value checks. This is probably a little bit too convoluted and also difficult to implement and document. A "user" option value (alongside another path option value) makes it clear that I want user provided file (and don't have fallback mechanism). We can easily return from [here](https://github.com/insightsengineering/teal/blob/f76a30ff4403e45e5ff18fcc9ccea05b84d00d72/R/module_teal_lockfile.R#L104-L106).
I believe that this would make it more explicit and simpler. Especially for the user.

Totally agree. We had a similar discussion in here https://github.com/insightsengineering/teal/pull/1263#discussion_r1760862414

So I guess it should/could be
- `teal.lockfile.path` 
    - when provided, it's superior to `teal.lockfile.enable` and if the file exists then it's possible to download the pre-computed file
    - when provided, but file does not exist, we `return(NULL)` from [here](https://github.com/insightsengineering/teal/blob/f76a30ff4403e45e5ff18fcc9ccea05b84d00d72/R/module_teal_lockfile.R#L104-L106) and we do not compute lockfile
    - when not provided, then regular flow for lockfile creation is followed
    
Regular flow, when `teal.lockfile.path` not provided
- check `teal.lockfile.emabled`
    - when set to TRUE, compute anyway (case for tests in our scenario, and unpredicted user scenarios)
    - when set to FALSE, do not compute
    - when not set - check IF_QUATRO, IF_TESTING, IF_R_CMD_CHECK, IS_CALLR 
        - if TRUE, omit
        - if FALSE, proceed
        
        
For the documentation - let's make this behavior clear in `options-vignette`. Let's link to `options-vignette` in the manual page of `module_teal_lockfile`


How does that sound @pawelru  :) ?

tests/testthat/setup-options.R
m7pr: resolving this and continuing the discussion in here https://github.com/insightsengineering/teal/pull/1263#discussion_r1766426661

R/module_teal_lockfile.R
pawelru: I am moving towards replacing `.enabled` option with a new one `.mode` which is checked first.

The logic would be as follows:
```mermaid
flowchart TD
    A{teal.lockfile.mode}
    A -->|true| yes[yes]
    A -->|auto| B{check conditions}
    B -->|true| yes
    B -->|false| B_no[logger]
    B_no --> no[no]
    A -->|user| C{teal.lockfile.path}
    A -->|false| A_false[logger]
    A_false --> no
    C -->|empty| C_empty[logger]
    C_empty --> no
    C -->|non-empty| F[use what's provided]
```

The biggest difference is to eliminate the "if smth is provided then its superior to smth" and "if condition is not met then fallback to regular". These are the weakest part IMHO.

m7pr: Oh my, this graph doesn't make me think anything got simplified : P

m7pr: So `teal.lockfile.mode` can be `true`, `false`, `auto` and `user`. And then if `teal.lockfile.mode = 'user'` we also check if `teal.lockfile.path` is empty, non-empty and exists, non-empty and does not exist.

I think it's harder to understand with 4 `options` for `mode` and 3 `scenarios` for `path` and harder to document. 

> The biggest difference is to eliminate the "if smth is provided then its superior to smth"

But it also exists in your flow, if `teal.lockfile.mode` it is provided it is superior to `teal.lockfile.path`.
I think with two options we will always need to prioritize one of it.

Other idea
1 option `teal.lockfile.mode` with values: `true`, `false`, `auto` and `actual_path`.

pawelru: > > The biggest difference is to eliminate the "if smth is provided then its superior to smth"
> 
> But it also exists in your flow, if `teal.lockfile.mode` it is provided it is superior to `teal.lockfile.path`. I think with two options we will always need to prioritize one of it.

There is a slight difference. What I am suggesting is to have one option to be always (!) superior than the other one. I want to eliminate auto-magic: if one is provided then it's superior.
The usage of `.path` is not conditional on its existence but rather based on another `.mode` value.

The second biggest difference to your (previous) suggestion is that if we got the "user" mode turned on then we go hard with that mode and throw if cannot be done - in particular: do not fallback to the auto mode.
I do consider "user" mode together with a incorrect path as a wrong input and I would like to notify user about that.

> Other idea
> 1 option teal.lockfile.mode with values: true, false, auto and actual_path.

What if my lockfile path is `true` or `false` or `auto`? I think it's a mistake to merge responsibilities of two vars into one. One responsibility is "mode" choice - one out of fixed scope of available choices. The other one is a path. The latter is used only if the former has some specific value.

m7pr: I would assume your lockfile path has `.lock` at the end, so it would rather not be `true`, `false` nor `auto` but rather `true.lock`, `false.lock`, or `auto.lock`. If `true.lock` or `auto.lock` are way to popular names and someone can name their lockfiles without extensions and use `true` or `auto`, then we can change mode to use `disabled` and `enabled`.

My point is that if `teal.lockfile.mode` is already a character, then it could take the path. And maybe we don't need 2 options. We can put a condition so that path needs to be specified with a prefixed tag: `option(teal.lockfile.mode = "path:custom_user_path.lock")`. Then you have a clear distinction on whether this is an actual path or type of the mode. 

```r
grepl('^path:', getOption('teal.lockfile.mode')) 
```

Let's you know if this is user provided lockfile, or all other possibilities

m7pr: > This is an interesting idea. I already have seen this in pkgdepends to install packages from different sources so you are giving the reference and the category of that reference which is a string, e.g. git::some_url or github::some_url or gitlab::some_url or local::path. So it's a way to classify / categorise string type of data which is an obligatory input.

Yes, I've seen it in `pkgdepends` :)

>  I think it's easier to have this as a separate being.

Alrighty, then let's keep it as you proposed. I'm on board!

- Check `teal.lockfile.mode` and if:
    - `auto` set on the package startup, does not look at `teal.lockfile.path`, goes into SCENARIO conditions
    - `true`, need to be set manually, does not look at `teal.lockfile.path`, does not goes into SCENARIO conditions, just computes
    - `false`, need to be set manually, does not look at `teal.lockfile.path`, does not goes into SCENARIO conditions, does not compute
    - `user`, need to be set manually, does look at `teal.lockfile.path` - stops when does not exist or empty, computes when not-empty and exists
    
Each path that ends up with lockfile not being computed shows logger trace.

Sounds good

pawelru: 👍

tests/testthat/test-module_teal.R
```diff
@@ -65,6 +65,59 @@ transform_list <<- list(
...
+        testthat::expect_false(file.exists(renv_filename))
+      }
+    )
+  })
+  testthat::it("is copied from option-teal.lockfile.path and then removed from an app directory", {
+    temp_lockfile_mock <- tempfile()
+    writeLines("test", temp_lockfile_mock)
+    withr::with_options(
+      list(
+        teal.lockfile.mode = "auto",
```
m7pr:
```suggestion
        teal.lockfile.mode = "user",
```
```suggestion
        teal.lockfile.mode = "auto",
```
```suggestion
        teal.lockfile.mode = "auto",
```

R/module_teal_lockfile.R
```diff
@@ -0,0 +1,219 @@
...
+        logger::log_debug("Removing lockfile after shutting down the app")
+        file.remove(lockfile_path)
+      }
+    })
+
+    lockfile_path <- "teal_app.lock"
+    mode <- getOption("teal.lockfile.mode", default = "")
+    user_lockfile_path <- getOption("teal.lockfile.path", default = "")
+
+    if (!(mode %in% c("auto", "true", "false", "user"))) {
```
gogonzo, line 80: `%in%` has precendence before `!`. 
Also mode should be problably "true" -> "enable", "false" -> "disable" as "true" and "false" are boolean values and having something else is weird in this context.
```suggestion
    if (!mode %in% c("auto", "true", "false", "user")) {
```

R/module_teal_lockfile.R
```diff
@@ -0,0 +1,219 @@
...
+      return(NULL)
+    }
+
+    if (mode %in% c("auto", "true") && !.is_lockfile_deps_installed()) {
+      warning("Automatic lockfile creation disabled. `mirai` and `renv` packages must be installed.")
+      shinyjs::hide("lockFileLink")
+      return(NULL)
+    }
+
+    if (mode == "auto" && .is_disabled_lockfile_scenario()) {
```
gogonzo, line 102: this `if` should be placed before previous `if`. Imagine that it is auto-disabled and renv isn't installed - then warning "mirai and renv packages must be installed" but the true reason is that this process would be disabled anyway.

R/module_teal_lockfile.R
```diff
@@ -0,0 +1,219 @@
...
+          file.copy(lockfile_path, file)
+          file
+        },
+        contentType = "application/json"
+      )
+    }
+    disable_lockfile_download <- function() {
+      warning("Lockfile creation failed.", call. = FALSE)
+      shinyjs::html("lockFileStatus", "Lockfile creation failed.")
+      shinyjs::disable("lockFileLink")
```
gogonzo, line 66: I'm still not convinced why we keep download button but disabled (forever). I think warning + notification and hide is nicer for an app-user.
```suggestion
      shinyjs::hide("lockFileLink")
```

gogonzo (commented):
minor comments

R/module_teal_lockfile.R
pawelru: how about "on"/"off"?

R/module_teal_lockfile.R
```diff
@@ -0,0 +1,219 @@
...
+      return(NULL)
+    }
+
+    if (file.exists(lockfile_path)) {
+      logger::log_debug("Lockfile has already been created for this app - skipping automatic creation.")
+      enable_lockfile_download()
+      return(NULL)
+    }
+
+    if (mode %in% c("auto", "true") && !.is_lockfile_deps_installed()) {
```
pawelru, line 96: if you exit from user and false earlier - you can remove the condition on the mode

R/module_teal_lockfile.R
m7pr: @gogonzo I know that

```r
!x %in% y
```

is the same as

```r
!(x %in% y)
```

but it always makes me confused when there are no parenthesis as it's not the same for `&` operator. So I better put parenthesis so the order is clear

R/module_teal_lockfile.R
m7pr: totally, thanks!

R/module_teal_lockfile.R
m7pr: Totally, thanks!

R/module_teal_lockfile.R
m7pr: When it comes to the name, I'd vote for `enabled/disabled` or `enable/disable`

R/module_teal_lockfile.R
m7pr: https://github.com/insightsengineering/teal/pull/1263/commits/5bdd79ca3139b9328ddfbc4c0b3faf78bc224841

R/module_teal_lockfile.R
```diff
@@ -0,0 +1,219 @@
...
+      if (file.exists(user_lockfile_path)) {
+        file.copy(user_lockfile_path, lockfile_path)
+        logger::log_debug('Lockfile set using option "teal.lockfile.path" - skipping automatic creation.')
+        enable_lockfile_download()
+      } else {
+        warning("Lockfile provided through options('teal.lockfile.path') does not exist.", call. = FALSE)
+        shinyjs::hide("lockFileLink")
+      }
+      return(NULL)
+    }
```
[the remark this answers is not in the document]
m7pr: https://github.com/insightsengineering/teal/pull/1263/commits/5bdd79ca3139b9328ddfbc4c0b3faf78bc224841

R/module_teal_lockfile.R
m7pr: https://github.com/insightsengineering/teal/pull/1263/commits/5bdd79ca3139b9328ddfbc4c0b3faf78bc224841

R/module_teal_lockfile.R
m7pr: https://github.com/insightsengineering/teal/pull/1263/commits/5bdd79ca3139b9328ddfbc4c0b3faf78bc224841

R/module_teal_lockfile.R
```diff
@@ -0,0 +1,219 @@
...
+      }
+    })
+
+    lockfile_path <- "teal_app.lock"
+    mode <- getOption("teal.lockfile.mode", default = "")
+
+    if (!(mode %in% c("auto", "true", "false", "user"))) {
+      stop("'teal.lockfile.mode' option can only be one of \"auto\", \"true\", \"false\", or \"user\". ")
+    }
+
```
gogonzo, line 82:
```suggestion
  if (mode != "user" && getOption("teal.lockfile.path", default = "") != "") {
    warning("teal.lockfile.path has been specified but teal.lockfile.mode is not 'user'. Ignoring user path as mode has precedence.")
  }
```

m7pr: It won't hurt to have this check : )

R/module_teal_lockfile.R
gogonzo: I don't like `teal.lockfile.mode` option to have 4 levels and `teal.lockfile.path` being considered only when `.mode == "user"`. I'm not a fan of specifying two options in order to provide pre-calculated file. I like the idea of `teal.lockfile.path` to automatically enable lockfile. Maybe the problem is in options naming? I think these options names are more relevant.

- `teal.lockfile.user` - If specified, lockfile is read from the given location. Has a precendence over `teal.lockfile.auto`
- `teal.lockfile.auto` (or `teal.lockfile.system`) - when set to `TRUE` (conditoinal default) then lockfile is calculated automatically using home directory.

image

Of course we need a initial warning if `teal.lockfile.auto = TRUE` and `teal.lockfile.user = "<not empty>` saying "teal.lockfile.user has a precedence over teal.lockfile.auto. Reading a lockfile from the user specified location"

m7pr: Just the summary of the discussion we had today - we remove `user` option, hence `option('teal.lockfile.path')` as it is risky to allow to display custom renv.lockfile that might be different than the actual lockfile of the project (potentially outdated). We will proceed with just 3 options: `disable`, `enable`, `auto`

m7pr (approved):
@pawelru since @gogonzo is the author, he can not approve the PR. I implemented all the changes we discussed and I am stating this is solid. Any final words?

R/module_teal_lockfile.R
chlebowa: > it always makes me confused when there are no parenthesis

I recommend highly `is.element` 😉

vignettes/teal-options.Rmd
```diff
@@ -84,9 +84,19 @@ This indicates whether to print the `JavaScript` console logs to the `R` console
 
 Default: `FALSE`.
 
-### `teal.renv.lockfile` (`character`)
+### `teal.lockfile.mode` (`character`)
 
-The path to the pre-computed `renv` lockfile that will be shared through teal app. To read more about lockfile usage creation check `?teal::teal_lockfile`.
+This enables app users to download `renv` lockfile by clicking `"download lockfile"` in the footer.
```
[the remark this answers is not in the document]
m7pr: https://github.com/insightsengineering/teal/pull/1263/commits/3ffc3de1b4f774c28fdc6edc2f0488014e4eaa37

pawelru (approved):
Looks good 👍 it took us a while to get there :)

m7pr:
@gogonzo you can do the honors of merging this in : ) great colab team!

PR #1273, opened 2024-07-23
Data module as tab@669 insert UI@main

gogonzo:
""


R/module_nested_tabs.R
```diff
@@ -68,16 +68,24 @@ ui_teal_module.teal_modules <- function(id, modules, depth = 0L) {
...
-          module_label <- modules$children[[module_id]]$label
-          tabPanel(
-            title = module_label,
-            value = module_id, # when clicked this tab value changes input$<tabset panel id>
-            ui_teal_module(
-              id = ns(module_id),
-              modules = modules$children[[module_id]],
-              depth = depth + 1L
+          if (module_id == "teal_data_module") {
+            tabPanel(
```
gogonzo, line 72: @vedhav 
Looks for me like `ui_teal_module.shiny.tag` instead of making this if trick here. Or make pseudo `teal_module` in `ui_teal` so that it can be handled without changes in `ui_nested_tabs`

R/module_nested_tabs.R
```diff
@@ -57,33 +57,45 @@ ui_teal_module.default <- function(id, modules, depth = 0L) {
...
-          module_label <- modules$children[[module_id]]$label
-          tabPanel(
-            title = module_label,
-            value = module_id, # when clicked this tab value changes input$<tabset panel id>
-            ui_teal_module(
-              id = ns(module_id),
-              modules = modules$children[[module_id]],
-              depth = depth + 1L
+  tags$div(
+    class = "teal-body",
```
gogonzo, line 61: Maybe this `div(class = "teal-body, ...)` should be in `ui_teal`. Otherwise this div will be added as many times as depth-levels. `modules(label = "tab1", modules(label = "tab1.1, modules(tab = "1.1.1", ...`

vedhav: True, good catch. Let me move it.

R/module_bookmark_manager.R
```diff
@@ -159,6 +159,7 @@ get_bookmarking_option <- function() {
   if (is.null(bookmark_option) && identical(getOption("shiny.bookmarkStore"), "server")) {
     bookmark_option <- getOption("shiny.bookmarkStore")
     # option alone doesn't activate bookmarking - we need to set shinyOptions
+    shinyOptions(bookmarkStore = bookmark_option)
```
gogonzo, line 162:
```suggestion
```

vedhav: Thanks! It is resolved now.

R/teal_data_module.R
```diff
@@ -22,6 +22,10 @@
 #'  `shiny` module server function; must only take `id` argument;
 #'  must return reactive expression containing `teal_data` object
 #' @param label (`character(1)`) Label of the module.
+#' @param once (`logical(1)`) Applicable only when used in the `data` argument of `init`.
+#'  If `TRUE`, the data module can only be consumed once. After the successful consumption,
+#'  the data module tab will disappear and the user will not be able to interact with the UI of the module.
```
gogonzo, line 27: Please also elaborate somewhere in a documentation about data-reloading. Separate section in this topic (`teal_data_module`) makes sense.

```suggestion
#'  If `TRUE`, the data module will be shown only once and will disappear after successful data loading. 
#' App user will no longer be able to interact with this module anymore. 
#' If `FALSE` ...
```

In the same time please get ready for a discussion (after merge) about `teal_data_module` and `teal_transform_module`. IMO we can't have a single constructors for both as assertions on their `server` function differ and now also `once` is dedicated to `teal_data_module` only.

R/1.0_module_data.R
```diff
@@ -39,9 +39,10 @@ NULL
 ui_data <- function(id, data, title, header, footer) {
   ns <- shiny::NS(id)
   shiny::div(
+    id = ns("teal_data_body"),
```
gogonzo, line 42: can be just `body` or preferably `content` as `ui_data` is already called with namespace `teal-data`, so `teal-data-content` is shorter than `teal-data-teal_data_body`.
```suggestion
    id = ns("content"),
```

R/module_teal.R
```diff
@@ -197,5 +204,9 @@ srv_teal <- function(id, data, modules, filter = teal_slices()) {
     mapping_table <- srv_filter_manager_panel("filter_manager_panel", slices_global = slices_global)
     snapshots <- srv_snapshot_manager_panel("snapshot_manager_panel", slices_global = slices_global)
     srv_bookmark_panel("bookmark_manager", modules)
+
+    if (inherits(data, "teal_data_module")) {
+      setBookmarkExclude(c("teal_modules-active_tab"))
+    }
```
gogonzo, line 210: Wrong namespace. I think exclusion should be set in `srv_teal_module.teal_modules`

```suggestion
```

------

Besides, I think we can do something like this. WDYT? If you agree it is a good idea please add todo comment in `srv_teal_module.teal_modules` (this is where active tab exists

```r
# pseudocode
onBookmark({
  if (input$active_tab == "teal_data_module") {
    input_list[[-active_tab]]
  }
})

```

R/modules.R
```diff
@@ -439,6 +439,9 @@ module_labels <- function(modules) {
 #' @keywords internal
 modules_bookmarkable <- function(modules) {
   checkmate::assert_multi_class(modules, c("teal_modules", "teal_module"))
+  if (inherits(modules$children[[1]], "shiny.tag")) {
+    modules$children[[1]] <- NULL
+  }
```
gogonzo, line 444: `modules_bookmarkable` is used directly only on the `server` side. It is `ui_bookmark_panel` which should receive `modules` without a `shiny.tag` 



```suggestion
```

gogonzo (commented):
@vedhav

R/modules.R
vedhav: I don't get this. The `ui_bookmark_panel` also calls the `need_bookmarking` which calls `modules_bookmarkable`. Which was causing this issue. Perhaps the solution would be to pass the data module as an attribute instead of `modules$children`.

gogonzo: What I meant was that `modules` should be passed to `ui_filter_manager` without `shiny.tag` element.

R/module_nested_tabs.R
```diff
@@ -160,6 +169,12 @@ srv_teal_module.teal_modules <- function(id,
   moduleServer(id = id, module = function(input, output, session) {
     logger::log_debug("srv_teal_module.teal_modules initializing the module { deparse1(modules$label) }.")
 
+    observeEvent(input$active_tab, once = TRUE, {
+      if (input$active_tab == "teal_data_module") {
+        setBookmarkExclude("active_tab")
+      }
+    })
```
gogonzo, line 176: sorry I think I had a bad suggestion. I was thinking more about following, but this would be bad if there was a `teal_data_module`. I think previous implementation (one you commented out in module_teal.R) made more sense.

```
onBookmark({
  if (input$active_tab == "teal_data_module") {
        setBookmarkExclude("active_tab")
   }
})
```

inst/js/sidebar.js
```diff
@@ -3,38 +3,33 @@
...
     .addClass("col-sm-9");
-  $(".teal_secondary_col").fadeIn(650);
+  $(`#${tabpanel_wrapper} .teal_secondary_col`).fadeIn(650);
 };
-const toggleFilterPanel = () => {
-  if (filter_open && !$(".teal_secondary_col").is(':visible')) {
-    showSidebar();
+const toggleFilterPanel = (tabpanel_wrapper) => {
+  console.log(`#${tabpanel_wrapper} .teal_secondary_col`);
+  console.log(filter_open[tabpanel_wrapper] === undefined);
```
gogonzo, line 21:
```suggestion
```

PR #1274, opened 2024-07-23
4 tests for `teal_modules`

m7pr:


tests/testthat/test-module_teal.R
```diff
@@ -437,13 +437,98 @@ testthat::describe("srv_teal teal_modules", {
...
+      expr = {
+        session$setInputs(`teal_modules-active_tab` = "module_1")
+        testthat::expect_identical(modules_output$module_1()()[['iris_raw']], iris)
+      }
+    )
+  })
+
+  testthat::test_that("doesn't receive extra data added in a transform", {
+    testthat::skip("Looks like it does receive extra data?")
+    # Added swiss in a transformer and this is added to data()
```
m7pr, line 481: @gogonzo looks like data created in transformer is added to the `reactive(teal_data())` that gets out of the transformer which is expected. Should we test if the extra data IS received?

tests/testthat/test-module_teal.R
```diff
@@ -437,13 +437,98 @@ testthat::describe("srv_teal teal_modules", {
     )
   })
 
-  testthat::test_that("receives data with datasets == module$datanames")
+  testthat::test_that("receives data with datasets == module$datanames", {
+    shiny::testServer(
+      app = srv_teal,
+      args = list(
+        id = "test",
+        data = reactive(teal_data(iris = iris, mtcars = mtcars)),
```
vedhav, line 445: Do we need to check that the datanames is populated for different data formats we support (`teal_data`, `reactive(teal_data)`, `reactiveVal(teal_data)`, `teal_data_module`)?

Additionally, I think datanames should be checked for three cases:
1. `datanames = "all"` the default case
2. `datanames = c("iris", "mtcars")` When datasets are explicitly mentioned
3. `datanames = c("ADTTE")` When parents should be included using join keys. In this case even `ADSL` should be included if we provide join_keys.

tests/testthat/test-module_teal.R
gogonzo: Yup, this is something we need to fix I guess.

tests/testthat/test-module_teal.R
```diff
@@ -437,23 +437,198 @@ testthat::describe("srv_teal teal_modules", {
...
+        data = teal_data(iris = iris, mtcars = mtcars),
+        modules = modules(
+          module("module_1", server = function(id, ...) {
+            testthat::expect_null(list(...)$data)
+          })
+        )
+      ),
+      expr = {
+        session$setInputs(`teal_modules-active_tab` = "module_1")
+        session$flushReact()
```
gogonzo, line 548: For readibility

```suggestion
            list(...)$data
          })
        )
      ),
      expr = {
        session$setInputs(`teal_modules-active_tab` = "module_1")
        session$flushReact()
        testthat::expect_null(modules_output$module_1())
```

tests/testthat/test-module_teal.R
gogonzo: @m7pr could you add a test to check the third scenario mentioned by @vedhav

gogonzo (commented):
`testthat::testthat` -> `testthat::it` when it is enclosed within `testthat::describe`

tests/testthat/test-module_teal.R
m7pr: Yup, will do @gogonzo

m7pr:
`testthat::test_that -> testthat::it inside testthat::describe` https://github.com/insightsengineering/teal/pull/1274/commits/6dba1e8b0c66fdf30283c22c373d968979a56021

tests/testthat/test-module_teal.R
m7pr: https://github.com/insightsengineering/teal/pull/1274/commits/781feeda6e973840314bfa0786a45f3b478d7625

PR #1275, opened 2024-07-26
Fixes lingering/ghost slices in the filter panel after DDL reload

averissimo:
# Pull Request

#### Fixes
- teal.slices - after removal of the slice make sure that observers of FilterState are destroyed. To test - take an app with DDL. Add filter and reload data several times. After this try to remove filter with remove button. What is interesting that single FIlterState card remove button needs to be pushed several times (it removes old FIlterStates which are not rendered anymore) while remove-all seems to not add observers one over another (it is enough to click once).

#### Needs

Branch `teal_slice@669_insertUI@main` on `{teal.slices}`

#### Changes description

- Keeps previous FilteredDataset to safely clean the filter panel

##### How to test?

1. Run app below
2. Reload the dataset using the "`data` tab"
3. Remove individual filter from filter panel
4. Check if filter has been removed with a single click

Repeat test without `module_specific = TRUE`

```r
# teal.slices -
# after removal of the slice make sure that observers of FilterState are destroyed.
#
# To test - take an app with DDL.
# - Add filter
# - reload data several times.
#
# After this try to remove filter with remove button.
# What is interesting that single FIlterState card remove button needs to be
# pushed several times (it removes old FIlterStates which are not rendered anymore)
# while remove-all seems to not add observers one over another (it is enough to
# click once). @averissimo

# Possible bug: reloading data will clear default values of the initial filter state

options(
  teal.log_level = "DEBUG",
  teal.show_js_log = TRUE,
)

pkgload::load_all("../teal")

data <- teal::teal_data_module(
  ui = function(id) {
    ns <- shiny::NS(id)
    shiny::tagList(
      shiny::textInput(ns("username"), label = "Username"),
      shiny::passwordInput(ns("password"), label = "Password"),
      shiny::actionButton(ns("submit"), label = "Submit")
    )
  },
  server = function(id, ...) {
      shiny::eventReactive(input$submit, {
          within(
            {
              logger::log_debug("Loading data...")
              ADSL <- teal.data::rADSL
              ADTTE <- teal.data::rADTTE
              iris <- iris

              CO2 <- CO2
              factors <- names(Filter(isTRUE, vapply(CO2, is.factor, logical(1L))))
              CO2[factors] <- lapply(CO2[factors], as.character)
            },
            username = input$username,
            password = input$password
          )
        join_keys(data) <- default_cdisc_join_keys[c("ADSL", "ADTTE")]
        teal.data::datanames(data) <- c("ADSL", "ADTTE", "iris", "CO2")
        data
      })
    })
  }
)

teal::init(
  data = data,
  modules = teal::modules(
    teal::modules(
      label = "Group",
      teal::example_module()
    )
  ),
  filter = teal::teal_slices(
    teal.slice::teal_slice("ADSL", "AGE", selected = c(18L, 65L)),
    module_specific = TRUE,
    teal.slice::teal_slice("ADSL", "SEX")
  ),
  title = "yada"
) |>
  shiny::runApp()
```


gogonzo (commented):
IMO cleaning previous `FilteredData` object should happen where they are re-created so we don't need to cache an old object.  Can `datasets_rv` be a `reactiveVal` instead of `eventReactive`, and does it make sense to call 

```r
datasets_rv <- reactiveVal()
eventReactive(data_rv(), {
  datasets_rv()$clear_filter_states()
  datasets_rv(teal_data_to_filtered_data(data_rv()))
})
```

averissimo:
> IMO cleaning previous FilteredData object should happen where they are re-created so we don't need to cache an old object. Can datasets_rv be a reactiveVal instead of eventReactive, and does it make sense to call

This was my first implementation, but I didn't like:

- the duplicate code we need to add in `module_teal` and `module_nested_tabs`
    - due to `module_specific` attribute
- logic change around using `reactiveVal()` as we still need to pass on `NULL` or some other flag down to `module_nested_tab`

On the plus side, `module_filter_manager` manages internals and lifecycle of `FilterData`, which seems fitting.

I'm very open to revert back, WDYT?

gogonzo (approved):
Hurray!

PR #1253, opened 2024-06-27
Feature branch PR

gogonzo:
Closes #669 #860 https://github.com/insightsengineering/teal.slice/issues/62 #1228

- needs `teal.slice` branch `669_insertUI@main`

to check:
- [ ] check teal.gallery apps
  - [x] RNA-seq @vedhav
  - [x] basic-teal  @m7pr 
  - [x] early-dev @vedhav
  - [x] efficacy @m7pr
      - [x] Filter Panel Should not be visible for `tm_front_page` (similar to `reporter_previewer`) https://github.com/insightsengineering/teal/pull/1292
      - [x] https://github.com/insightsengineering/teal.slice/issues/607
      - [x] https://github.com/insightsengineering/teal.slice/issues/608
  - [x] exploratory @m7pr 
      - [ ] https://github.com/insightsengineering/teal.modules.general/issues/768
      - [x] https://github.com/insightsengineering/teal.modules.general/issues/769
      - [x] https://github.com/insightsengineering/teal/issues/1287
      - [ ] https://github.com/insightsengineering/teal/issues/1286 https://github.com/insightsengineering/teal.data/pull/322
  - [x] longitudinal @averissimo 
  - [x] patient-profile @m7pr 
      - [x] https://github.com/insightsengineering/teal/issues/1290 https://github.com/insightsengineering/teal.reporter/pull/282
  - [X] safety @m7pr 
  - [x] python @vedhav
- [x] check our teal.connector.xyz if they are supported - possibly not because modal can't open modal. 
  - [x] `teal.connector.harbour` @m7pr (also tested by @vedhav on week 30)
  - [x] `teal.connector.cdse` @averissimo 
  - [x] `teal.connector.rice` @averissimo (tested by @vedhav on week 30)
  - [x] `teal.connector.ssh` @averissimo https://github.com/insightsengineering/teal.connectors.ssh/pull/8
- [x] Can app developer provide simple js code to hide all elements of certain class (for example reporter buttons)
  - [x] [this](https://github.com/insightsengineering/teal/issues/1220) should be easy to fix now @gogonzo https://github.com/insightsengineering/teal/pull/1253#issuecomment-2270393344
  - [x] [this](https://github.com/insightsengineering/teal/issues/1114) should be easy now @gogonzo https://github.com/insightsengineering/teal/pull/1253#issuecomment-2270393344
- [x] teal as a module - use `ui_teal` and `srv_teal` in any application (see teal as a module example below)
[...]

PR #1280, opened 2024-07-30
use `topological_sort` on `datanames()`

m7pr:
Part of https://github.com/insightsengineering/teal/pull/1253


R/utils.R
```diff
@@ -191,7 +191,19 @@ check_filter_datanames <- function(filters, datanames) {
 teal_data_datanames <- function(data) {
   checkmate::assert_class(data, "teal_data")
   if (length(teal.data::datanames(data))) {
-    teal.data::datanames(data)
+    datanames <- teal.data::datanames(data)
+    join_keys <- teal.data::join_keys(data)
```
m7pr, line 195:
```suggestion
    joinkeys <- teal.data::join_keys(data)
```

R/utils.R
```diff
@@ -191,7 +191,19 @@ check_filter_datanames <- function(filters, datanames) {
...
+    datanames <- teal.data::datanames(data)
+    joinkeys <- teal.data::join_keys(data)
+    if (length(join_keys()) > 0) {
+      child_parent <- sapply(
+        datanames,
+        function(name) teal.data::parent(joinkeys, name),
+        USE.NAMES = TRUE,
+        simplify = FALSE
+      )
+      intersect(topological_sort(child_parent), datanames)
```
gogonzo, line 203: I guess you wanted a `union`?
```suggestion
      union(topological_sort(child_parent), datanames)
```

m7pr: I followed the implementation in `teal.slice` https://github.com/insightsengineering/teal.slice/blob/4366cf9ad01a97210d03e4f776215a758e0b69ff/R/FilteredData.R#L126

This part for not clear, but I guess `topological_sort` might return more names than there are in `datanames`, that's why we limit the result with `intersect`.

gogonzo: if `topological_sort(child_parent)` is always a superset of datanames then ok. But union is much safer, because we wan't `datanames + parents` in order where parents preceed their child.

m7pr: Okey, make sense. In case there is a parent in join_keys but not in datanames with union it will appear before the children

gogonzo: Skärmavbild 2024-07-31 kl  08 26 17

From the last tab in the above example

m7pr: Got it @gogonzo . `union` was included in the code

m7pr: Also added a test to https://github.com/insightsengineering/teal/pull/1280/commits/a31ebfbb5187a25c01fbbc23ce8761e34864f27e

gogonzo: @m7pr example app I've provided still shows wrong order of datanames. Please fix

m7pr: Does it? The last tab has `c("ANL", "iris", "ADTTE")` specified as datanames. There are only `join_keys` for `"ADSL"` and `"ADTTE"` in `data`. So `ADSL` is added because it is a parent of `ADTTE`. So we end up with what's on your print screen. 

Ahhhh, but I get we would like to see `ADTTE` just after `ADSL` and not far behind at the end? I don't think topological sort does that - it only ensures `ADSL` is somewhere before `ADTTE` - but I will investigate.

m7pr: Ah yeah, I see - `ADSL` should be directly before `ADTTE` and there was a stupid typo https://github.com/insightsengineering/teal/pull/1280/commits/5e7717ae216b6c94a1975b1f76e0c745ad0c8391

m7pr: After 2 fix commits the result is still the same as on your print.screen. Give me some more time before I call you for a review! Stay tuned!

m7pr: Ok, got this sorted out
image


We needed to use `teal_data_datanames` instead of `teal.data::datanames` in `summary-table` https://github.com/insightsengineering/teal/pull/1280/commits/be270a8cb1389f7c3a4b0c1fa53e1e5d60682136

R/utils.R
```diff
@@ -191,7 +191,19 @@ check_filter_datanames <- function(filters, datanames) {
...
+    datanames <- teal.data::datanames(data)
+    joinkeys <- teal.data::join_keys(data)
+    if (length(join_keys) > 0) {
+      child_parent <- sapply(
+        datanames,
+        function(name) teal.data::parent(joinkeys, name),
+        USE.NAMES = TRUE,
+        simplify = FALSE
+      )
+      union(unlist(topological_sort(child_parent)), datanames)
```
gogonzo, line 203: looks like this should be in `teal.data::datanames` **getter**. I think `topological_sort@669_insertUI@main` in `teal.data` is a better way to go - only tests would apply here

m7pr: If we want to include that in `teal.data` then I think we should adjust the code in all the pieces where
- datanames get changed
- join_keys get changed

As both of them influence the `topological_sort()`.

So I created a PR where we edited `teal_data` consturctor, `"datanames<-"` method and `"join_keys<-"` method. https://github.com/insightsengineering/teal.data/pull/318
However I will need to consult with Andre, if we need to edit `names(join_keys) <-` and `parents(join_keys) <- ` as well

m7pr: PR in `teal.data` https://github.com/insightsengineering/teal.data/pull/319 - will decrease this PR only to have tests

R/1.0_module_data_summary.R
```diff
@@ -129,7 +129,7 @@ srv_data_summary <- function(id, teal_data) {
 #' @rdname module_data_summary
 #' @keywords internal
 get_filter_overview <- function(teal_data) {
-  datanames <- teal.data::datanames(teal_data())
+  datanames <- teal_data_datanames(teal_data())
```
gogonzo, line 132:
```suggestion
  datanames <- teal.data::datanames(teal_data())
```

gogonzo (approved):
Thanks!

PR #1281, opened 2024-07-30
Adjust `TealAppDriver` methods for `filter_panel` `shinytest2` tests

m7pr:
Part of https://github.com/insightsengineering/teal/pull/1253


m7pr:
This fixes `modules`, `filter-panel` and `teal_slices` tests, as they use `$get_active_filter_vars()` and `$navigate_tab()` methods

tests/testthat/test-shinytest2-module_bookmark_manager.R
```diff
@@ -3,7 +3,7 @@ testthat::test_that("bookmark_manager_button is not rendered by default", {
   app <- TealAppDriver$new(
     data = simple_teal_data(),
     modules = example_module(label = "Example Module"),
-    options = options()
+    options = list()
```
m7pr, line 6: @vedhav could you explain this change?

vedhav: It's cause the machine running the test can have different options which will make the tests perform differently.
For example in my machine by default, I set the testing depth to 5 in my global .Rprofile.
Sorry, I should have commented about this.

m7pr: No worries. All clear

vedhav (approved):
LGTM! Let's check the `e2e: filter panel is not displayed when datanames is NULL` after the datanames issue is resolved.

PR #1283, opened 2024-08-01
Have different constructor for "Data" module and "Transform" module

vedhav:
Adds `teal_transform_module()` to be used to create data modules for transformation.

Example usage

```r
library(teal)
data <- teal_data() %>%
  within({
    iris <- iris
    mtcars <- mtcars
  })
my_transformers <- list(
  teal_transform_module(
    label = "Custom transform for iris",
    ui = function(id) {
      ns <- NS(id)
      tags$div(
        numericInput(ns("n_rows"), "Number of rows to subset", value = 6, min = 1, max = 150, step = 1)
      )
    },
    server = function(id, data) {
        reactive({
          data() %>%
            within(
              {
                iris <- head(iris, num_rows)
              },
              num_rows = input$n_rows
            )
        })
      })
    }
  )
)
app <- init(
  data = data,
  modules = example_module(transformers = my_transformers)
)
if (interactive()) {
  shinyApp(app$ui, app$server)
}
```


R/teal_data_module.R
```diff
@@ -61,15 +61,71 @@
...
+#'       ns <- NS(id)
+#'       tags$div(
+#'         numericInput(ns("n_rows"), "Number of rows to subset", value = 6, min = 1, max = 150, step = 1)
+#'       )
+#'     },
+#'     server = function(id, data) {
+#'       moduleServer(id, function(input, output, session) {
+#'         reactive({
+#'           data() %>%
+#'             within(
```
gogonzo, line 107: No `magrittr` please
```suggestion
#'             within(data(),
```

R/teal_data_module.R
```diff
@@ -61,15 +61,71 @@
...
     once = once
   )
 }
+
+#' Data module for `teal` transformers.
+#'
+#' @description
+#' `r lifecycle::badge("experimental")`
+#'
+#' Create a `teal_data_module` object for custom transformation of data for pre-processing before passing the data into the module.
```
gogonzo, line 78:
```suggestion
#' Create a `teal_data_module` object for custom transformation of data for pre-processing before passing the data into the `server` of the [teal_module()].
```

R/teal_data_module.R
```diff
@@ -61,15 +61,71 @@
...
+#'
+#' @details
+#' `teal_transform_module` creates a `teal_data_module` object to transform the data in a `teal` application
+#' after it has been passed through the filter panel and before it is passed to the modules which uses the transform module.
```
gogonzo, line 82: "modules" is too ambiguous, please use "server instance of a [teal_module()]".
```suggestion
#' `teal_transform_module` creates a `teal_data_module` object to transform a filtered data
#' before handing over to the server of a [teal_module()].
```

gogonzo (commented):
Please link `[teal_transform_module()]` in description `transformers` argument in docs of `teal_module` function. Please review current `teal_module` and `teal_transform_module` documentation and write comprehensive description aimed for app developer - app developer needs to know everything about how `teal_transform_module` and `teal_module` works, but don't need to know `FilteredData`. What I mean by this is that you did right writing "passed through filter panel" but writing "modules" should not occur in this case. App developer needs to understand that `teal_data` object is sequenced through all `teal_transform_modules`' servers  (it is a list) and eventually it is handed over to the server of `teal_module`

Roxygen docs should be more technical, while vignette should be focused on usage

Would be nice if @donyunardi looked at this content

R/teal_data_module.R
```diff
@@ -61,15 +61,71 @@
 #' @export
 teal_data_module <- function(ui, server, label = "data module", once = TRUE) {
   checkmate::assert_function(ui, args = "id", nargs = 1)
-  checkmate::assert(
-    checkmate::check_function(server, args = "id", nargs = 1),
-    # todo: allow for teal_data_module$server to have 'data' argument or break this in teal_transformer_module
-    checkmate::check_function(server, args = c("id", "data"), nargs = 2)
-  )
+  checkmate::check_function(server, args = "id", nargs = 1)
```
gogonzo, line 64: Please make tests and fix failing ones.
```suggestion
  checkmate::assert_function(server, args = "id", nargs = 1)
```

vignettes/data-transform-as-shiny-module.Rmd
```diff
@@ -0,0 +1,172 @@
...
+
+## Introduction
+
+The input data supplied by the app developer in the `data` argument of the `init` function is passed through the filter panel and the filtered data is supplied to the all modules.
+It is important to note that the filter panel only performs one kind of computation on the datasets, it filters the data based on the user interaction.
+However, there are use cases where we need to provide a more customized transformation options for the app user.
+Historically, customised data transformations was achieved in the teal apps using the encoding panel, where the module developer provides custom UI and server logic for the data transformation.
+But, with the new `teal_transform_module` function, it is now possible to provide a more dynamic transformation capability which can be customised by the app developer and not just the module developer.
+
+The `teal_transform_module()` function can be used to create a `teal_data_module` object that is basically a shiny module where the module server function takes in the data from the filter panel and returns the data after applying the custom transformation logic.
```
donyunardi, line 21:
```suggestion
When data transformations need to occur inside the teal app, they typically happen within the teal module development by the teal module developer. This results in the transformation code being tightly coupled with the teal module. Any changes to the data transformation logic require updating the teal module, which can impact the teal app development until the module is updated.

With the new `teal_transform_module()` function, it is now possible to provide a more dynamic data transformation capability that can be customized by the teal app developer, not just the teal module developer.

The `teal_transform_module()` function is essentially a Shiny module that takes `ui` and `server` arguments and can be provided to the `transformer` argument in a `module()`. When provided, teal will execute the data transformation logic for the specified module when it is loaded. 

In addition, this approach makes it easier for teal app developers to handle additional data transformations while using the teal modules R packages like `teal.modules.general` and `teal.modules.clinical`.

Let’s explore this new way to manage custom data transformations in teal apps.
```

Here's an alternative of how I think the introduction could be.

vignettes/data-transform-as-shiny-module.Rmd
```diff
@@ -0,0 +1,172 @@
...
+data <- teal_data() %>%
+  within({
+    iris <- iris
+    mtcars <- mtcars
+  })
+
+
+app <- init(
+  data = data,
+  modules = example_module()
```
donyunardi, line 41: Should we use a very simple `module()` so that it's clear that `transformer` is passed as an argument to this function?

R/teal_data_module.R
donyunardi:
```suggestion
#' `teal_transform_module` creates a `teal_data_module` object to transform data in a `teal` application.
#' This transformation happens after the data has passed through the filtering activity in teal. The transformed data is then sent to the server of the [teal_module()].
```
Another alternative

donyunardi (commented):
Provided some alternative to the content.

vignettes/data-transform-as-shiny-module.Rmd
```diff
@@ -0,0 +1,172 @@
...
+
+```{r}
+library(teal)
+data <- teal_data() %>%
+  within({
+    iris <- iris
+    mtcars <- mtcars
+  })
+
+my_transformers <- list(
```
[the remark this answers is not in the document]
gogonzo: I guess it makes sense, but let me think about this. Currently we have:
- `module` is wrapped by `modules` 
- `teal_slice` is wrapped `teal_slices`
- `teal_data_module` is not wrapped by anything
- `teal_transform_module` is wrapped by `list()`

vignettes/data-transform-as-shiny-module.Rmd
gogonzo: Yes, I think vignette should elaborate more about creating `module` which consumes transformers.

I've updated description here:

PR #1284, opened 2024-08-01
Add landing popup argument in `init` and soft deprecate calling it as a normal module

vedhav:
Changes:

1. Soft deprecates calling the `landing_popup_module` inside `modules` argument of `init`
2. Add a new `landing_popup` argument to `init` to accept this module.
3. Update the `TealAppDriver` to accept this new argument.


R/init.R
```diff
@@ -159,10 +163,15 @@ init <- function(data,
...
-  landing_module <- NULL
   if (length(landing) == 1L) {
-    landing_module <- landing[[1L]]
+    landing_popup <- landing[[1L]]
     modules <- drop_module(modules, "teal_module_landing")
+    # TODO: verify the version before release.
+    lifecycle::deprecate_soft(
+      when = "0.16",
+      what = "landing_popup_module()",
+      details = "The `landing_popup_module` is supposed to be called as the `landing_popup` argument of the `init` instead of passing it as a module inside the `modules` argument of `init`"
```
gogonzo, line 173:
```suggestion
      details = "Pass `landing_popup_module` to the `landing_popup` argument of the `init` instead of wrapping it into `modules()` and passing to the `modules` argument"
```

or

```r
Pass `landing_popup_module` to the `init(landing_popup)` argument 
instead of wrapping it into `modules()` and passing to the `init(modules)`"
```

R/init.R
```diff
@@ -159,10 +163,15 @@ init <- function(data,
   # argument transformations
   ## `modules` - landing module
   landing <- extract_module(modules, "teal_module_landing")
-  landing_module <- NULL
   if (length(landing) == 1L) {
-    landing_module <- landing[[1L]]
+    landing_popup <- landing[[1L]]
```
m7pr, line 167:
```suggestion
    landing_popup <- landing_popup
```

R/init.R
```diff
@@ -159,10 +163,15 @@ init <- function(data,
   # argument transformations
   ## `modules` - landing module
   landing <- extract_module(modules, "teal_module_landing")
```
m7pr, line 165:
```suggestion

```

R/init.R
```diff
@@ -159,10 +163,15 @@ init <- function(data,
...
-  landing_module <- NULL
   if (length(landing) == 1L) {
```
m7pr, line 166:
```suggestion
  if (length(landing_popup) == 1L) {
```

R/init.R
m7pr: Or remove this line complitely. Previously landing_popup was extracted from modules, now it is a parameter

m7pr:
@vedhav I added my comments, but those are relevant if we want to support only the parameter approach. If we want to support both: parameter and the possibility to pass as module (even for the soft deprecation period) then my comments are not releveant

PR #1285, opened 2024-08-01
Workflow propagations

walkowif:
Propagation of:
* https://github.com/insightsengineering/idr-tasks/issues/781
* https://github.com/insightsengineering/idr-tasks/issues/785
* https://github.com/insightsengineering/idr-tasks/issues/793
* https://github.com/insightsengineering/idr-tasks/issues/800
* https://github.com/insightsengineering/idr-tasks/issues/802


m7pr (approved):

m7pr:
Hey @walkowif hey @pawelru I see that r-hub workflow on Scheduled Action failed for the last action https://github.com/insightsengineering/teal/actions/runs/10251244595/job/28358867629
Is that intended?

m7pr:
I see it fails for

>  :heavy_multiplication_x: No lookup refs provided. Skipping.

https://github.com/insightsengineering/r.pkg.template/blob/main/.github/workflows/rhub.yaml#L84

So I guess this is the place that needs a fix

m7pr:
If inputs.lookup-refs is empty - should we remove from the top of the inputs, and from the parameters of the run?

walkowif:
I'm not sure if each and every package will need the non-empty `lookup-refs` input.
If not, we might have to adjust the assertion in `setup-r-dependencies`.
If yes, we'll simply have to provide a proper `lookup-refs` input to each package, but the correct value for that input will have to be checked package by package.

m7pr:
This was fixed by Pawel in here https://github.com/insightsengineering/teal/pull/1291#pullrequestreview-2220543014

PR #1291, opened 2024-08-06
add refs to rhub workflow

pawelru:
https://github.com/insightsengineering/teal/actions/runs/10262416165


m7pr (approved):
Fixed runs - as you can see in here https://github.com/insightsengineering/teal/actions/runs/10262416165/job/28392056413

PR #1296, opened 2024-08-07
Always position the teal util icons in the same place

vedhav:
Have a fixed position for the icons in the top right which do not change place when the number of teal tab grows.

#### Before
Screenshot 2024-08-07 at 7 10 13 PM

#### After the changes in the PR
Screenshot 2024-08-07 at 7 09 45 PM

### DIFF
--- a/R/module_teal.R
+++ b/R/module_teal.R
@@ -108,7 +108,7 @@ ui_teal <- function(id,
     ),
     tags$div(
       id = ns("options_buttons"),
-      style = "margin-left: auto;",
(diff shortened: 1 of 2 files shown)


m7pr:
@vedhav I think this is great, but I would ask @kumamiao for the opinion if this is acceptable change

kumamiao:
Looks good to me, thanks @vedhav and @m7pr for checking!

m7pr (approved):

PR #1301, opened 2024-08-08
Extend teal transform vignette

vedhav:
Changes:

1. Adds the section on how to place the transform UI in the custom position inside the module.
2. Formats the `_pkgdown.yml` file (No change made to the file, just styling)
3. Extend the `example_module()` to handle the selection when datasets change (This can happen when using DDL to change datasets completely, such an example will be added to `teal.gallery`)


vignettes/data-transform-as-shiny-module.Rmd
```diff
@@ -160,6 +160,70 @@ if (interactive()) {
 When a custom transformation is used, the UI for the transformation is placed below the filter panel.
 However, there is a way to customize the placement of the UI inside the module content.
 
-This can only be done when the module allocates a dedicated space for the transform UI.
+In order to place the transformation UI inside the module there are few things one has to do:
+1. Create a custom module wrapper function.
+2. In the module wrapper function, Call the module and store it in a variable.
```
vedhav, line 165:
```suggestion
2. Call the desired module in the module wrapper function and store it in a variable so it can be modified.
```

vignettes/data-transform-as-shiny-module.Rmd
```diff
@@ -160,6 +160,70 @@ if (interactive()) {
 When a custom transformation is used, the UI for the transformation is placed below the filter panel.
 However, there is a way to customize the placement of the UI inside the module content.
 
-This can only be done when the module allocates a dedicated space for the transform UI.
+In order to place the transformation UI inside the module there are few things one has to do:
+1. Create a custom module wrapper function.
+2. Call the desired module in the module wrapper function and store it in a variable so it can be modified.
+3. Modify the UI of the module with the transform UI at the desired location by calling the `ui_transform_data`. Note that in order for the transform to work you need to change the name space of the id by passing `NS(gsub("-module$", "", id), "data_transform")`.
```
vedhav, line 166: @averissimo The namespace change is the part that is not great. Is there another way to do this?
@m7pr suggested to have some kind of a wrapper to achieve this. Either for the name space or the whole module customization itself.

averissimo: I think there needs to be a trick somewhere as we're injecting the `ui_transform` under different namespace locations, so we have 2 options:

1. As is with no namespace trick when injecting the `ui_transform_data` in most cases and trick on manual
2. The other way round

From a design point of view, it seems that the trick should be applied on the exception/rare case. But this makes it module-developer visible.

Feel more than free to propose a change :-D ... I'm not keen on it as it's a lesser of evils kind of implementation.

I've tried getting to the `SessionProxy` parent, but the public API seems to not allow for it, `override` of `SessionProxy` is lost as scope is created (only accessing the original session)

vignettes/data-transform-as-shiny-module.Rmd
```diff
@@ -160,6 +160,70 @@ if (interactive()) {
...
+  mod$ui <- function(id) {
+    ns <- NS(id)
+    teal.widgets::standard_layout(
+      output = verbatimTextOutput(ns("text")),
+      encoding = tags$div(
+        ui_transform_data(NS(gsub("-module$", "", id), "data_transform"), transformers),
+        selectInput(ns("dataname"), "Choose a dataset", choices = NULL),
+        teal.widgets::verbatim_popup_ui(ns("rcode"), "Show R code")
+      )
+    )
```
m7pr, line 184: Maybe we can invent an example that is not based on `example_module`?
Maybe you can create a simple `module` that uses `ui` and `server` and you can show an application on whether it uses the default placement, and a custom placement?

This way it might be a bit more clear, instead of showing an example where we use pre-defined module and then we adjust it's UI. Could be simpler to have simple module with ui and server and then version adjusted with ui_transform_data added

vignettes/data-transform-as-shiny-module.Rmd
```diff
@@ -160,6 +160,70 @@ if (interactive()) {
...
+  }
+  attr(mod$transformers, "custom_ui") <- TRUE
```
m7pr, line 186: Trying to figure out if this can be added automatically

vignettes/data-transform-as-shiny-module.Rmd
```diff
@@ -160,6 +160,70 @@ if (interactive()) {
 When a custom transformation is used, the UI for the transformation is placed below the filter panel.
 However, there is a way to customize the placement of the UI inside the module content.
 
-This can only be done when the module allocates a dedicated space for the transform UI.
+In order to place the transformation UI inside the module there are few things one has to do:
+1. Create a custom module wrapper function.
+2. Call the desired module in the module wrapper function and store it in a variable so it can be modified.
```
m7pr, line 165:
```suggestion
2. Call the desired module in the module wrapper function and store it in a variable so it's UI can be modified.
```

vignettes/data-transform-as-shiny-module.Rmd
m7pr, line 166:
```suggestion
3. Modify the UI of the module with the transform UI at the desired location by calling the `ui_transform_data`. Note that in order for the transform to work you need to change the namespace of the `id` by passing `NS(gsub("-module$", "", id), "data_transform")`.
```

m7pr: Would this help a bit, if we expose/create such a function in this example?
```
transform_id <- function(id) {
   NS(gsub("-module$", "", id), "data_transform")
}
```

vedhav: `transform_id` does simplify getting the namespace with the added advantage that we can change the namespaces in teal and still not break apps using this feature because they use our wrapper.
Are we okay to export a function that caters to a small audience? Advanced users should be okay with this kind of complexity, moreover, we will have teal transform wrappers and this could be a topic to discuss during that time.
And this feature could be like a hidden feature that we don't advertise.

vignettes/data-transform-as-shiny-module.Rmd
vedhav: I wanted to do it as you mentioned by creating a custom module and then injecting this.
However, The typical use case for this would be to inject the custom transform in a module from a package like tmc. And, the vignette would become longer with custom module code which might distract the users from the actual important code needed to position the transform.
So, I decided to just place the transform to the custom location from the example mentioned above.

m7pr: hmmm, I'll think about it

m7pr:
@vedhav check my proposition for the wrapper that appends the `encoding` argument of `module$ui` part with the function needed to show the transformer, for custom position 145f1d95900d9dcee4f805df0e4fd01000c0a7cb

Maybe you'd like to polish the wrapper a bit? I had issues using `append()` function, that's why I used the for loop. But maybe you will be more successful

```r
mod$ui
function(id) {
      ns <- NS(id)
        encoding = tags$div(
        )
      )
    }
<bytecode: 0x000001e87d53f730>
<environment: 0x000001e800f38ad0>
> custom_module_position(mod, position = 1)$ui
function (id) 
{
    ns <- NS(id)
    teal.widgets::standard_layout(output = verbatimTextOutput(ns("text")), 
        encoding = tags$div(ui_transform_data(transform_id(id), 
            transformers), selectInput(ns("dataname"), "Choose a dataset", 
            choices = NULL), teal.widgets::verbatim_popup_ui(ns("rcode"), 
            "Show R code")))
}
<environment: 0x000001e800f38ad0>
> custom_module_position(mod, position = 2)$ui
function (id) 
{
    ns <- NS(id)
    teal.widgets::standard_layout(output = verbatimTextOutput(ns("text")), 
        encoding = tags$div(selectInput(ns("dataname"), "Choose a dataset", 
            choices = NULL), ui_transform_data(transform_id(id), 
            transformers), teal.widgets::verbatim_popup_ui(ns("rcode"), 
            "Show R code")))
}
<environment: 0x000001e800f38ad0>
> custom_module_position(mod, position = 3)$ui
function (id) 
{
    ns <- NS(id)
    teal.widgets::standard_layout(output = verbatimTextOutput(ns("text")), 
        encoding = tags$div(selectInput(ns("dataname"), "Choose a dataset", 
            choices = NULL), teal.widgets::verbatim_popup_ui(ns("rcode"), 
            "Show R code"), ui_transform_data(transform_id(id), 
            transformers)))
}
<environment: 0x000001e800f38ad0>
```

m7pr:
Hey @vedhav it was hard to propose a suggestion on the text that was not edited in the PR so I allowed myself to make a commit. https://github.com/insightsengineering/teal/pull/1301/commits/799740b53ad30de1bd89cf6cb9b59f579a62665e with few propositions for changes. We can comment under those changes in the GitHub UI now. 

I had an impression that there is too much poetic and unclear statements in the introduction. I just made it more developer-like-documentation specific. Let me know your thoughts on the changes. No hard feelings if you still think we can improve this.

vignettes/data-transform-as-shiny-module.Rmd
m7pr: I think it is fine to use `example_module` if we got with the wrapper that I proposed

vedhav:
@m7pr I think we will have a hard time maintaining this wrapper function:
1. It only works when the `teal.widgets::standard_layout` is used with `encoding`. So, this does not work on `teal.modules.general::tm_variable_browser()`
2. Even if we handle all the UI cases, we create a reverse dependency of managing the `custom_module_position()` function when some UI structure of the downstream packages.
3. If there are custom modules created by people, there is a chance that our wrapper might not work so it's hard to recommend it to people.
4. With all these cons, I do not like the fact that we have an exported wrapper when there is a way to position UI in any module if they know what to do.

I think it will be a good to circle back on this topic when we implement the `teal.transform` wrappers for data extracts. It might become useful to create some utility around that time. I think now we are early.

m7pr:
@vedhav totally agree, especially that the solution is based on `encoding = ` name of the argument. Name does not need to be specified :) However this is a good starting point - we at least had a chance to figure out how to manage functions with `body` and `as.call` so that `ui` changes in the future might be more flexible. We can delete the wrapper for now.

Any comments to the content changes of the vignette?

vedhav:
I like the content changes :)

m7pr (approved):

m7pr:
Hey I do approve but I'd like @donyunardi to review again, as he said he reviewed it previously. Also @donyunardi it is a chance for you to get up to date with the current state of the solution

vignettes/data-transform-as-shiny-module.Rmd
```diff
@@ -1,26 +1,30 @@
...
-With the new `teal_transform_module()` function, it is possible to provide a more dynamic data transformation capability that can be customized by the teal app developer, not just the teal module developer.
-The `teal_transform_module()` function is essentially a Shiny module that takes `ui` and `server` arguments and can be provided to the `transformer` argument in a `module()`. When provided, teal will execute the data transformation logic for the specified module when it is loaded. 
-In addition, this approach makes it easier for teal app developers to handle additional data transformations while using the teal modules R packages like `teal.modules.general` and `teal.modules.clinical`.
-Let’s explore this new way to manage custom data transformations in teal apps.
+`teal` ver `0.15.x` (TO BE UDPATED) introduced new argument in `teal::module` called `transformers`. 
+This argument allows to pass a `list` of `teal_data_module` objects that are created either with `teal_transform_module()` or `teal_data_module()` functions.
```
donyunardi, line 16: I'm unsure if it's necessary to mention using the `teal_data_module()` function to create the object and pass it to the `transformers` argument. This may make sense from a developer's perspective since `teal_transform_module()` also returns a `teal_data_module` class, but I'm not sure if it's useful information for the user.

In all the examples below, we only use `teal_transform_module()` to create the object passed to `transformers`.

If we decide to keep the mention of `teal_data_module()`, we would need to include examples of how to pass the object to `transformers` using that function. I expect that using `teal_data_module()` would involve additional steps, which might not be ideal since `teal_transform_module()` already handles this.

At the moment, I prefer to focus on `teal_transform_module()`, but I'd like to hear your thoughts on this.

```suggestion
This argument allows to pass a `list` of `teal_data_module` objects that are created using `teal_transform_module()` function.
```

vignettes/data-transform-as-shiny-module.Rmd
```diff
@@ -1,26 +1,30 @@
...
+
+This vignette presents the way on how to manage custom data transformations in `teal` apps.
 
 ## Creating your first custom data transformation module
 
-We initialize a simple teal app where we pass `iris` and `mtcars` as the input datasets.
+We initialize a simple `teal` app where we pass `iris` and `mtcars` as the input datasets.
 
 ```{r, message = FALSE, warning = FALSE}
 library(teal)
```
donyunardi, line 30: Since the `teal` library is already loaded in an earlier section, there's no need to repeat `library(teal)` in this chunk or in any subsequent chunks. WDYT?

vignettes/data-transform-as-shiny-module.Rmd
m7pr: Agree. I also had those thoughts when I wrote this sentence. Will accept the change

m7pr: Actually @vedhav can you accept during your last review? This all is ready to be accepted and merged

vedhav: Good catch Dony, It's actually only important to talk about the object `teal_data_module` in the context of the transform module to make sure that we communicate that both the DDL and transform use the same object.
But, right now we already have different features between the object created with `teal_transform_module()` and `teal_data_module()` that it does not make sense to use the later for transforming.

vignettes/data-transform-as-shiny-module.Rmd
vedhav: Thanks, removed instances of redundant library calls.

PR #1314, opened 2024-08-13
docs: correct get_client_timezone return value

pawelru:


R/teal_data_utils.R
```diff
@@ -1,6 +1,6 @@
 #' `teal_data` utils
 #'
-#' In `teal` we need to recreate the `teal_data` object due to two operations:integer.max
```
pawelru, line 3: probably weird auto-complete

donyunardi (approved):

PR #1318, opened 2024-08-14
get_teal_bs_theme(): use checkmate for throw and check

pawelru:
one step closer towards consistent error / warning message when checking classes

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -25,14 +25,22 @@ get_client_timezone <- function(ns) {
 #' @keywords internal
 get_teal_bs_theme <- function() {
   bs_theme <- getOption("teal.bs_theme")
+
   if (is.null(bs_theme)) {
-    NULL
-  } else if (!inherits(bs_theme, "bs_theme")) {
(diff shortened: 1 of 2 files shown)


R/utils.R
```diff
@@ -25,14 +25,22 @@ get_client_timezone <- function(ns) {
...
+    return(NULL)
+  }
+
+  if (!checkmate::test_class(bs_theme, "bs_theme")) {
+    msg <- paste0(
+      "Assertion on 'teal.bs_theme' option value failed: ",
+      checkmate::check_class(bs_theme, "bs_theme"),
+      ". The default shiny bootstrap will be used."
+    )
+    warning(msg)
```
gogonzo, line 39: `...` elements in `warning` are being concatenated
```suggestion
    warning(
      "Assertion on 'teal.bs_theme' option value failed: ",
      checkmate::check_class(bs_theme, "bs_theme"),
      ". The default shiny bootstrap will be used."
     )
```

gogonzo (approved):
Thanks for improving message. I liked previous `if-else` style, but it is not a blocker

PR #1324, opened 2024-08-14
check reactive using custom checkmate extension

pawelru:
- create custom validation functions that fits the checkmate 
  - for now, to limit the changes, these are private (not-exported) functions but happy to discuss and export if needed
- apply it wherever applicable


R/module_teal_data.R
pawelru, line 1: FYI: the whole journey starts from here. The current way is not consistent when it comes to parsing the error message. With this change it's more consistent not only within this file but also with the whole package.

vignettes/adding-support-for-reporting.Rmd
```diff
@@ -35,7 +35,7 @@ my_module <- function(label = "example teal module") {
   module(
     label = label,
     server = function(id, data) {
-      checkmate::assert_class(data, "reactive")
+      assert_reactive(data)
```
gogonzo, line 38: Nobody is required to assert on `data`, we guarantee that this will be a `reactive`
```suggestion
```

vignettes/creating-custom-modules.Rmd
```diff
@@ -113,7 +113,7 @@ ui_histogram_example <- function(id) {
 # displays a histogram of the selected variable
 # displays code to reproduce the histogram
 srv_histogram_example <- function(id, data) {
-  checkmate::assert_class(data, "reactive")
+  assert_reactive(data)
   checkmate::assert_class(isolate(data()), "teal_data")
```
gogonzo, line 117:
```suggestion
```

vignettes/creating-custom-modules.Rmd
```diff
@@ -62,7 +62,7 @@ my_module <- function(label = "example teal module") {
   module(
     label = label,
     server = function(id, data) {
-      checkmate::assert_class(data, "reactive")
+      assert_reactive(data)
       checkmate::assert_class(isolate(data()), "teal_data")
```
gogonzo, line 67:
```suggestion
```

vignettes/adding-support-for-reporting.Rmd
```diff
@@ -35,7 +35,7 @@ my_module <- function(label = "example teal module") {
...
       checkmate::assert_class(isolate(data()), "teal_data")
```
gogonzo, line 39:
```suggestion
```

R/module_teal_data.R
```diff
@@ -96,16 +96,17 @@ srv_validate_reactive_teal_data <- function(id, # nolint: object_length
                                             data,
                                             modules = NULL,
                                             validate_shiny_silent_error = FALSE) {
-  moduleServer(id, function(input, output, session) {
-    if (!is.reactive(data)) {
-      stop("The `teal_data_module` passed to `data` must return a reactive expression.", call. = FALSE)
-    }
+  checkmate::assert_string(id)
+  assert_reactiveExpr(data)
```
gogonzo, line 100: Imagine you're app developer and you get an error message:

```
Error in srv_validate_reactive_teal_data(id = "validate", data = data_out,  : 
  Assertion on 'data' failed: Must be a reactive expression (i.e. inherit from 'reactiveExpr' class) but has class 'character'.
```

While on main it is:

```
The `teal_data_module` passed to `data` must return a reactive expression.
```

----

The message should be directed to app developer not to us. App developer should not know anything about `srv_validate_teal_data`, this is why `call. = FALSE` was needed.

R/teal_data_module-eval_code.R
```diff
@@ -35,9 +35,7 @@ setMethod("eval_code", signature = c("teal_data_module", "character"), function(
       moduleServer(id, function(input, output, session) {
         teal_data_rv <- object$server("mutate_inner")
 
-        if (!is.reactive(teal_data_rv)) {
-          stop("The `teal_data_module` must return a reactive expression.", call. = FALSE)
-        }
+        assert_reactiveExpr(teal_data_rv, .var.name = "teal_data_module")
```
gogonzo, line 38: Disagree with the change - eval_code will return error message which gives no clue to app developer what is wrong

gogonzo (changes requested):
I admit that there is a room for improvement on main branch but:
- Error messages should help app developer to figure out what went wrong with `teal_data_module` or `teal_transform_module` that one provided. I think this PR worsens the quality of these messages. If PR suppose to make errors more consistent then it should at least keep the quality of them at the same level. This is why `if (...) stop()` for `teal_data_module` output is necessary and shouldn't be asserted in the same way as in other places where we expect some reactives - this message should be directed to app developer, not to us.

- I don't see any reason of replacing `checkmate::check_class(teal_data, "reactive")` with `assert_reactive(teal_data)`. `checkmate::check_class(teal_data, "reactive")` is a one liner and straightforward.

----

>  happy to discuss and export if needed

To put it mildly this is not a value added at all and it is not responsibility of `teal` to provide such utilities. I have an opinion that all `validate_` functions shouldn't be exported from `teal` neither, but for historical reasons I've not yet suggested to move or remove them. I think NAMESPACE of teal should contain only:

- init
- modules, module
- ui/srv_teal
- teal_data_module
- teal_transform_module
- teal_slices

and probably (but not necessary):
- landing_page_module
- reporter_previewer_module

Adding anything to this list pollutes documentation in my opinion.

pawelru:
I talked with @gogonzo about this and here are the outcomes:
- we agreed that this should remain as a private functions not to extend `teal` NAMESPACE
- we are going to keep standard error message using checkmate flavour and also we are going to add contextual information by a new decorator function `decorate_err_message(msg: <str>)` so that at the end it will look like this: `assert_xyz() |> decorate_err_message("perhaps you mean blablabla")`.
- move most (if not all?) validation logic into the teal-data-module (and also: teal-transform) constructor function. By doing this we can have access to module label, function name etc. that should allow us to build more informative error message. This also mean that inside teal we would have _correct_ objects and don't have to worry whether these has been validated or not. On this Dawid agreed to join me for development.

gogonzo (approved):
@pawelru I've implemented: 
- validate if `teal_data_module`/`teal_transform_module` returns a `reactive` in the constructor
- applied `decorate_err_msg` so it indicates which function needs to be fixed
- removed unused functions `assert_reactiveExpr` and `assert_reactiveVal`, when they will be needed then we can bring them back
- removed unnecessary `checkmate::assert_class(data, "reactive")` from `teal_module` examples as they are unnecessary. Users don't need to bother if `teal` provides a correct object - it will always.

R/teal_data_module.R
```diff
@@ -64,7 +64,16 @@ teal_data_module <- function(ui, server, label = "data module", once = TRUE) {
...
   structure(
-    list(ui = ui, server = server),
+    list(
+      ui = ui,
+      server = function(id) {
+        data_out <- server(id)
+        decorate_err_msg(
+          prefix = "teal_data_module()",
+          assert_reactive(data_out, .var.name = "server")
+        )
```
pawelru, line 74: Hey @gogonzo. Copied from your message:

> applied decorate_err_msg so it indicates which function needs to be fixed (needs to return a reactive). We have access also to the label argument which can indicate which teal_data_module produced a failure. But I'm not sure how the decoration should look like. Maybe "teal_data_module(label = 'label...') or teal_data_module labeled as "label ..."

image

I have a few comments on this:
- teal_data_module() needs to be code-formatted to become `teal_data_module()`. In checkmate it is done with `'` ticks
- just `teal_data_module()` might not be enough. This is quite tricky as this is not about this function inputs but actually what's returned from the module constructed with this function. 
- `server` is also misleading - this is about what server function returns. I think I would keep the current variable name so that devs can debug and rely on decoration for app/module developer.

All in all, I would propose the following: 
```suggestion
        decorate_err_msg(
          prefix = sprintf("From: 'teal_data_module()':\nA 'teal_data_module' with \"%s\" label:
", label),
          suffix = "Please make sure that this module returns a 'reactive` object containing 'teal_data' class of object.",
          assert_reactive(data_out)
        )
```

So that the final message would look like this:
```
Error:
From: 'teal_data_module()':
A 'teal_data_module' with "module label" label:
Assertion on 'data_out' failed: ...
Please make sure that this module returns a 'reactive` object containing 'teal_data' class of object.
```

I wanted to discuss this with you instead of pushing directly.

Other:
- we must introduce assert on `label` before using this so this must be added beforehand
- similar (same?) thing needs to be added to the `teal_transform_module()`

gogonzo: @pawelru I agree with all above. One tiny thing, `prefix` and `suffix` is not precise as it refers to part of the word instead of part of the sentence. I suggest `pre`, `post`

pawelru: Pushed the changes. Please have a look. Closing this

R/module_teal_data.R
```diff
@@ -116,9 +115,10 @@ srv_validate_reactive_teal_data <- function(id, # nolint: object_length
             need(
               FALSE,
               paste(
-                strip_style(data_out$message),
+                "Shiny error when executing the `data` module:",
+                strip_style(paste(data_out$message, collapse = "\n")),
```
gogonzo, line 119: Not needed as `data_out$message == ""`  always - because of the condition: `if (inherits(data_out, "shiny.silent.error") && identical(data_out$message, ""))`

```suggestion
                "Shiny error when executing the `data` module",
```

gogonzo:
✅ Approve. Thanks @pawelru 💪

PR #1325, opened 2024-08-16
reorg module_teal.R

pawelru:
reorg inside `ui_teal()` and `srv_teal()` so that we have:
- better readability
- better separation on concerns
- easier extendability
- other

Initially I started with three sections within a single function but splitting this into multiple functions has some additional benefits.

This will be probably anyway significantly changed soon during UX reorg but it gives a good foundation for the implementation.


R/module_teal.R
```diff
@@ -180,38 +134,130 @@ srv_teal <- function(id, data, modules, filter = teal_slices()) {
...
       id = "teal_modules",
       data_rv = data_rv,
       datasets = datasets_rv,
       modules = modules,
       slices_global = slices_global
     )
-    mapping_table <- srv_filter_manager_panel("filter_manager_panel", slices_global = slices_global)
-    snapshots <- srv_snapshot_manager_panel("snapshot_manager_panel", slices_global = slices_global)
+    srv_filter_manager_panel("filter_manager_panel", slices_global = slices_global)
+    srv_snapshot_manager_panel("snapshot_manager_panel", slices_global = slices_global)
```
pawelru, line 260: FYI: variables assigned but not used - removed the assignment

pawelru: my mistakes - this is actually used in the shinytest2
will revert this soon

pawelru: done

tests/testthat/test-module_teal.R
```diff
@@ -171,7 +171,7 @@ testthat::describe("srv_teal arguments", {
 testthat::describe("srv_teal teal_modules", {
   testthat::it("are not called by default", {
     shiny::testServer(
-      app = srv_teal,
+      app = teal:::srv_teal_body,
```
gogonzo, line 174: Just saying - we don't test private functions. We test their effects only through public interface

pawelru: That's what I initially had but apparently the test are failing. See: https://github.com/insightsengineering/teal/actions/runs/10419973058/job/28859103688

The public `srv_teal` does not have access to variables defined in its child and private modules.

R/module_teal.R
```diff
@@ -180,38 +134,135 @@ srv_teal <- function(id, data, modules, filter = teal_slices()) {
       once = TRUE,
       handlerExpr = {
         session$userData$timezone <- input$timezone
-        logger::log_debug("srv_teal@1 Timezone set to client's timezone: { input$timezone }.")
+        logger::log_debug("srv_teal_header@1 Timezone set to client's timezone: { input$timezone }.")
```
gogonzo, line 137: Handling a timezone being classified as a "header" process is subjective.

tests/testthat/test-module_teal.R
gogonzo: Testing private function `teal:::srv_teal_body` needs better reasoning that public function didn't have access to some objects. PR won't be accepted until good reason will be provided and test files reorganized. My idea is that everything should be tested through public functions as testing private doesn't guarantee the precision (private functions in tests might be run in a different context that they are executed in the framework, thus their test result are not 100% reliable)

gogonzo (changes requested):

gogonzo:
@pawelru this branch is desynchronised and has failing tests. There was pretty significant changes in `teal` in the mean time and with "decorate" there will be another. I'm closing this PR as it isn't a good moment for code refurnishing.

PR #1334, opened 2024-08-29
datanames slot in `teal_transform_module` (ignore `@datanames`)

gogonzo:
Alternative solution to https://github.com/insightsengineering/teal/pull/1327

```
#' @param datanames (`character`) A vector with `datanames` that are relevant for the item. The
#'   filter panel will automatically update the shown filters to include only
#'   filters in the listed datasets. `NULL` will hide the filter panel,
#'   and the keyword `"all"` will show filters of all datasets. `datanames` also determines
#'   a subset of datasets which are appended to the `data` argument in server function.
```

If you look at the `?module` documentation of `datanames` argument, there is nothing about "if datanames is 'all' then all datasets from @datanames will be taken". It is very hard to explain as `datanames` is not only a `module` property, which makes it difficult to describe in a documentation. However, in the vignette we have it described in details [see here](https://insightsengineering.github.io/teal/latest-tag/articles/including-data-in-teal-applications.html#teal_data-properties)

Here we have a following rules:
1. Module's server get's always `module$datanames` if specified. 
    - When transform argument  is specified in the `module(transform)` then module appends `transform$datanames` to `module$datanames` to always keep `module$datanames` as a single point of truth. `transform` is always added as an argument of `module()` function, so resolving datanames there makes sense and is the most optimal. 
2. When `module$datanames = "all"` then all objects from transformed `data` is passed to the module.
    - When data contains datasets which shouldn't be displayed for a given module then one should `module |> set_datanames()` to limit displayed datasets
    

    
 **Advantage:**
- single point of truth (`module$datanames`) 
- don't need to take care of `data@datanames`
- can re-set `datanames` for the whole group of modules if needed 
 
 **Disadvantages:**
 - If re-set is applied for all modules `modules(mod1, mod2, mod3) |> set_datanames()` then each module's datanames will be re-set even if their `datanames` have been set to what is exactly needed. Possible solution is to `append_datanames()` instead of `set_datanames()`, WDYT?
 
 <details>
 <summary>app</summary>
 
 ```r
pkgload::load_all("teal")

trans <- list(
  anl_w_datanames = teal_transform_module(
    label = "ANL with datanames",
    ui = function(id) {
      ns <- NS(id)
      tagList(
        div("This transformer adds ANL based on specified ADSL, ADTTE"),
        numericInput(ns("obs"), "Number of subjects", value = 400, min = 0, max = 400)
      )
    },
    server = function(id, data) {
      moduleServer(id, function(input, output, session) {
        reactive({
          within(data(),
[...]

PR #1335, opened 2024-08-30
datanames slot in `teal_transform` (enhanced `teal.data::datanames()`)

gogonzo:
Done for https://github.com/insightsengineering/teal.data/pull/330 where I enhanced `teal.data::datanames()` function to return `ls(data@env)` when `@datanames` is not specified. In the long term I suggest `@datanames` field to be discontinued.

### DIFF
--- a/R/dummy_functions.R
+++ b/R/dummy_functions.R
@@ -21,7 +21,7 @@ example_module <- function(label = "example teal module", datanames = "all", tra
       checkmate::assert_class(isolate(data()), "teal_data")
       moduleServer(id, function(input, output, session) {
         datanames_rv <- reactive({
-          .teal_data_ls(req(data()))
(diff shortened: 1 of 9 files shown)


m7pr:
@gogonzo I agree that maintaining `teal_data@datanames`, `modules$datanames` and `ls(get_env(teal_data))` became problematic and it's confusing on how many ways are there of specyfing this. If we can get rid of `teal_data@datanames` and stick to  `ls(get_env(teal_data))` then at least it brings some clarity to the ecosystem. We could go even one step further and rename `teal.data::datanames()` to `teal.data::ls_env()` as it's gonna be even more clear

gogonzo:
Closing this PR in favour of https://github.com/insightsengineering/teal/pull/1334

Issue #1303, opened 2024-08-08
[Bug]: Document the different behaviour of active tab and hidden tab when using Data module


vedhav:
### What happened?

There is a different behaviour of data flow in modules when using transform during init.
The transform module server seems to be called for every module. But, shiny can trigger `reactive` values differently based on the shown/hidden state of the module. This can causing different behaviour for the active tab and hidden tab. It would be best to showcase this in the vignette about teal data module creation.

In the following example you can see that we have two `teal_modules` that are identical. However, we get a validation error in second module which is hidden and the first module works fine.

```r
devtools::load_all("teal")

my_transformers <- list(
  teal_transform_module(
    label = "Keep first 6 from IRIS",
    ui = function(id) {
      ns <- NS(id)
      div(
        checkboxInput(ns("check"), label = "Toggle `head(iris)`"),
      )
    },
    server = function(id, data) {
      moduleServer(id, function(input, output, session) {
        eventReactive(input$check, {
          print("Check triggered")
          req(data())
          if (input$check) {
            within(data(), iris <- head(iris, 6))
          } else {
            data()
          }
        })
      })
    }
  )
)

data <- teal_data(iris = iris)
app <- init(
  data = data,
  modules = modules(
    example_module("Module with transformations 1", transformers = my_transformers),
    example_module("Module with transformations 2", transformers = my_transformers)
  )
)

shinyApp(app$ui, app$server)
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [X] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [X] I agree to follow this project's Security Policy.

averissimo:
@vedhav can you confirm that this was fixed in #1253 ?

vedhav:
This issue persists. When I run the app mentioned in this issue I get the error message in the second tab that could have been avoided.

<img width="1468" alt="Screenshot 2024-08-13 at 2 56 27 PM" src="https://github.com/user-attachments/assets/cb543e22-6b1b-4244-b0f3-d08efb0181c9">
<img width="1470" alt="Screenshot 2024-08-13 at 2 56 33 PM" src="https://github.com/user-attachments/assets/8e58d79d-f35e-4ea1-b6f5-2f2908cc16cd">

averissimo:
_TL;DR_: This is a consequence of our transform design that penalizes transform modules that are not reactive to `data()`

Action: Do you think we can detect or mitigate it?

[...]

PR #1330, opened 2024-08-28
Fix errors while handling different input data types in teal

vedhav:
Closes #1304, #1307, and #1308

PR #1336, opened 2024-08-30
1303 add a note to `teal_data_module` constructor's vignette

m7pr:
As discussed, the note would be helpful in the `teal_data_module` constructor's vignette https://github.com/insightsengineering/teal/issues/1303
@donyunardi assigning you for the review as you said yesterday you might have a look at this one.


donyunardi (changes requested):
I understand that `teal_transform_module` is an instance of `teal_data_module`, which is probably why we use the term `teal_data_module` here. However, since the conversations leading up to this PR have been more focused on `teal_transform_module`, it would be better to keep these terms separate. `teal_data_module` should refer to defining delayed data, while `teal_transform_module` is for transforming data within the module. 

Additionally, I think it would be better to include this information in the `data-transform-as-shiny-module.Rmd` vignette, as it relates more directly to the transformation process.

We should also make it clear on what user need to do if they choose to use `eventReactive`. I'm fine with directing user to the discussion link, but we should summarize it clearly for users too.

m7pr:
Hey, I have

- moved the note from `data-as-shiny-module.Rmd` to `data-transform-as-shiny-module.Rmd`
- changed the name from `teal_data_module` into `teal_transform_module`
- specify that `eventReactive` needs to be reactive on `data()` in the `server` part

donyunardi (approved):
At first, I thought we should show some code if users were to use `eventReactive`. But after thinking it over, if we're discouraging users from doing this, maybe it's best not to mention it at all.

LGTM!

gogonzo:
Issue might be outdated when https://github.com/insightsengineering/teal/pull/1330 goes in

gogonzo (changes requested):
I don't think this is a good way. Please see the comment:

https://github.com/insightsengineering/teal/issues/1303#issuecomment-2355909715

m7pr:
I think we should revisit, once we are done with https://github.com/insightsengineering/teal/pull/1341

donyunardi:
@gogonzo @m7pr can you please assess if we still need this PR?
If not, please close.

gogonzo:
> @gogonzo @m7pr can you please assess if we still need this PR? If not, please close.

[I raised my opinion in the issue](https://github.com/insightsengineering/teal/issues/1303#issuecomment-2355909715), I think more effort on the `teal`'s reactivity is needed before giving up and writing a comment in a vignette.

Issue #1311, opened 2024-08-13
[Bug]: `._raw_` datasets should not be shown in the `Show R Code`


m7pr:
`._raw_` datasets got into the `Show R Code` somehow - https://genentech.shinyapps.io/nest_basic-teal_dev/

````r
library(dplyr)
IRIS <- iris %>% mutate(Species1 = Species, Species2 = Species, Species3 = Species, Species4 = Species, Species5 = Species)
MTCARS <- mtcars
stopifnot(rlang::hash(IRIS) == "1caadedc699b329d8d427daee0696be1")
stopifnot(rlang::hash(MTCARS) == "d0487363db4e6cc64fdb740cb6617fc0")
IRIS._raw_ <- IRIS
MTCARS._raw_ <- MTCARS
IRIS <- dplyr::filter(IRIS, Sepal.Width > 2.5 & Petal.Length > 1.5)
MTCARS <- dplyr::filter(MTCARS, gear == 3)
````

![image](https://github.com/user-attachments/assets/92e46f6d-e39d-4309-91cc-9bb4405f98d1)

m7pr:
It's a question on whether we use
````
teal.data::get_code(data, datanames = dataname_corrected_with_raw)
````
or
````
teal.data::get_code(data, datanames = dataname_corrected)
````
https://github.com/insightsengineering/teal/pull/1253/files#diff-4f959897d64b51a6ac57c9016430bdd92163a17fa62b5a61707a4b3bc7dd6325R42-R60

m7pr:
As @gogonzo mentioned

> this is intentional. data sent to module contain ._raw_ datasets. And everything which is in the qenv needs to have a code.

gogonzo:
I think we can close it or we can discuss is not too ugly and unique enough to make sure that user won't create this oneself (if app developer creates `_raw`, then it will be hard to detect which _raw has been created by teal and which by app developer). WDYT @m7pr  @vedhav @averissimo

vedhav:
I am okay with having this on the reproducible code chunk. It provides a checkpoint where data is not filtered/transformed and this can be helpful. I'm also okay with `._raw_`.

averissimo:
Agree. We can close this for now.

Issue #969, opened 2023-11-21
Filter mapping by ID instead of label


pawelru:
Conceptually, I find it quite odd to map filter to module _labels_. It should be a module _identifier_ instead. A label should be a free text but we tend to use it as identifier which is probably not the right thing to do.

chlebowa:
Aren't module ids created from the labels behind the scenes and never made available to the user?

pawelru:
Yes and I don't feel it's correct. Like I said earlier, a label should be a free text, in particular: an empty character or non-unique values between modules. These all should be allowed but it's not because we treat labels as identifier.

chlebowa:
That is a fair point, however, juggling labels AND ids is additional burden on the user.

pawelru:
Yes. We can probably came up with some convenience features but I agree that managing two attributes is more or less twice as difficult as managing just one. 

This doesn't change the point that IMHO right now labels are misused as ids. Recent rollout of mapping feature just exposed such flaw. Until this, the concept of module id was just useless for the user. I think we need to rethink the concept of module ID. Until a new not-yet-anticipated feature / extension that requires an ID.

lcd2yyz:
I remember we discussed this briefly during the design of the module-specific mechanism, but decided to stick with the labels for the sake of moving on, because there weren't clear/better alternatives at the time. I concur @pawelru's thoughts, thanks for raising the topic again.

gogonzo:
We are in dead end with this one.

1. Issue with mapping argument - even if we decide to do following

```r
modules(
  tm_xy("mod1", ,..., filters = teal_slices(teal_slice_age, teal_slice_sex),
  modules(
    label = "tab",
    tm_xy("mod1", ...., filters = teal_slices(teal_slice_age)
  )
)
``` 

We would be able to map modules properly together even if they are nested. 


2. Problem with displaying mapping in the app - here situation is different because we need to have a visual depresentation of these modules. For above example we would display (currently it is not possible because we don't allow duplicated labels when module-specific):

| filter | mod1 | mod 1 |
| ---- | ---- | ---- |
| AGE |  x | x |
| SEX | x | o |


And we rather should display this information in following way:

| filter | mod1 | tab > mod 1 |
| ---- | ---- | ---- |
| AGE |  x | x |
| SEX | x | o |

gogonzo:
Another solution is:

- teal to force `label` change when duplicated.
- remove nested modules

pawelru:
[...]

PR #1342, opened 2024-09-05
Change name of the raw (unfiltered object)

gogonzo:
From `<dataname>._raw_` to `.<dataname>_raw`. This means that `teal.data::datanames()` is not really needed. `ls(data@env)` returns all object names from environment except prefixed by `.` (`all.names = FALSE` is a default).

This adds clarity to the handling of a datanames in teal application: 
- "all" means, all object from the environment except those prefixed by `.`
- One can exclude datanames globally, for example by changing `con -> .con` and `ADSL_temp -> .ADSL_temp`


NEWS.md
```diff
@@ -22,6 +22,7 @@
 * Easier way of to call `javascript` events by setting `$(document).ready(function() { ... })`. #1114
 * Provided progress bar for modules loading and data filtering during teal app startup.
 * Filter mapping display has a separate icon in the tab.
+* Environment of the `data` passed to `teal_module`'s server contains unfiltered datasets names `.<dataname>_raw`
```
m7pr, line 25:
```suggestion
* Environment of the `data` passed to `teal_module`'s server contains unfiltered datasets names with the following convention: `.<dataname>_raw` 
```

R/teal_data_utils.R
```diff
@@ -42,8 +42,14 @@ NULL
...
   checkmate::assert_class(data, "teal_data")
   checkmate::assert_class(datanames, "character")
-  datanames_corrected <- intersect(datanames, ls(data@env))
-  dataname_corrected_with_raw <- intersect(c(datanames, sprintf("%s._raw_", datanames)), ls(data@env))
+  datanames_corrected <- intersect(datanames, .teal_data_ls(data))
+  dataname_corrected_with_raw <- c(
+    datanames_corrected,
+    intersect(
+      sprintf(".%s_raw", datanames_corrected),
+      ls(data@env, all.names = TRUE)
```
m7pr, line 50:
```suggestion
      ls(teal.code::get_env(data), all.names = TRUE)
```

R/teal_data_utils.R
```diff
@@ -42,8 +42,14 @@ NULL
...
+    )
+  )
```
m7pr, line 52: I am trying to figure out why in `.subset_teal_data` we once use `.teal_data_ls(data)` and in the next line we use `ls(data@env, all.names = TRUE)`.

Currently the only difference between `.teal_data_ls` and `ls`, is that the first uses `include_parent_datanames` which extracts the names of the `parents` for datasets based on `join_keys`.

Is the point of `include_parent_datanames` in `.teal_data_ls` to extend the list of names by `parents` or just to reorder the names? Based on the below code, where `iris` is a parent of `iris_join` I don't get `iris` in `@env` as a result of `teal:::.subset_teal_data(data, "iris_join"` but I do see that `join_keys` are created for that case.

I am asking because I am a bit confused. I see the same behavior on main but without the need to have `include_parent_datanames` in `.teal_data_ls`

# This branch

```r
iris_join <- data.frame(Species = unique(iris$Species), Letter = letters[1:3])

data <- teal.data::teal_data(
  iris = iris, .iris_raw = iris, 
  iris_join = iris_join, .iris_join_raw = iris_join, 
  mtcars = mtcars, .mtcars_raw = mtcars, 
  x = swiss,
  factor = factor(c('a', 'b'))
)
teal.data::datanames(data) <- c('iris', 'iris_join', 'mtcars')


teal.data::join_keys(data) <- teal.data::join_keys(
  teal.data::join_key("iris", "iris_join", keys = c("Species"))
)


teal:::.subset_teal_data(data, 'iris')
#> Registered S3 method overwritten by 'teal':
#>   method        from      
#>   c.teal_slices teal.slice
#> ✖ unverified teal_data object
#> <environment: 0x0000029ab3151490> [L]
#> Parent: <environment: package:stats>
#> Bindings:
#> • iris: <df[,5]> [L]
#> • .iris_raw: <df[,5]> [L]
teal:::.subset_teal_data(data, 'iris_join')
#> ✖ unverified teal_data object
#> <environment: 0x0000029ab5a54680> [L]
#> Parent: <environment: package:stats>
#> Bindings:
#> • iris_join: <df[,2]> [L]
#> • .iris_join_raw: <df[,2]> [L]
teal.data::join_keys(teal:::.subset_teal_data(data, 'iris'))
#> An empty join_keys object.
teal.data::join_keys(teal:::.subset_teal_data(data, 'iris_join'))
#> A join_keys object containing foreign keys between 2 datasets:
#> iris: [no primary keys]
#>   <-- iris_join: [Species]
#> iris_join: [no primary keys]
#>   --> iris: [Species]


teal:::include_parent_datanames('iris', teal.data::join_keys(data))
#> [1] "iris"
teal:::include_parent_datanames('iris_join', teal.data::join_keys(data))
#> [1] "iris"      "iris_join"
```


# Main branch

```r
iris_join <- data.frame(Species = unique(iris$Species), Letter = letters[1:3])

data <- teal.data::teal_data(
  iris = iris, iris._raw_ = iris,
  iris_join = iris_join, iris_join._raw_ = iris_join,
  mtcars = mtcars, mtcars._raw_ = mtcars,
  x = swiss,
  factor = factor(c('a', 'b'))
)

teal.data::datanames(data) <- c('iris', 'iris_join', 'mtcars')


teal.data::join_keys(data) <- teal.data::join_keys(
  teal.data::join_key("iris", "iris_join", keys = c("Species"))
)


teal:::.subset_teal_data(data, 'iris')
#> Registered S3 method overwritten by 'teal':
#>   method        from      
#>   c.teal_slices teal.slice
#> ✖ unverified teal_data object
#> <environment: 0x0000022e5af49b20> [L]
#> Parent: <environment: package:stats>
#> Bindings:
#> • iris._raw_: <df[,5]> [L]
#> • iris: <df[,5]> [L]
teal:::.subset_teal_data(data, 'iris_join')
#> ✖ unverified teal_data object
#> <environment: 0x0000022e5d7c12c8> [L]
#> Parent: <environment: package:stats>
#> Bindings:
#> • iris_join: <df[,2]> [L]
#> • iris_join._raw_: <df[,2]> [L]
teal.data::join_keys(teal:::.subset_teal_data(data, 'iris'))
#> An empty join_keys object.
teal.data::join_keys(teal:::.subset_teal_data(data, 'iris_join'))
#> A join_keys object containing foreign keys between 2 datasets:
#> iris: [no primary keys]
#>   <-- iris_join: [Species]
#> iris_join: [no primary keys]
#>   --> iris: [Species]
```

m7pr:
Hey @gogonzo this looks really good. The behavior is the same as on main. I left one comment about the usage of `teal:::include_parent_datanames` as I would like to understand it's purpose in `.teal_data_ls()`

pawelru:
I haven't looked into the implementation but I like the change based on the description.

I have one idea though. Have you considered having `.raw_data` list instead? So this would be `.raw_data$<dataname>` instead of `.ADSL_raw`. Maybe this will be more elegant to use (and exclude where needed)?
Actually when I'm thinking now about this - `.raw_env` as an environment type of object would be even better.

gogonzo:
> I haven't looked into the implementation but I like the change based on the description.
> 
> I have one idea though. Have you considered having `.raw_data` list instead? So this would be `.raw_data$<dataname>` instead of `.ADSL_raw`. Maybe this will be more elegant to use (and exclude where needed)? Actually when I'm thinking now about this - `.raw_env` as an environment type of object would be even better.

We haven't consider but it is an interesting idea. I have nothing against and not super hyped about this. Would you like to halt this PR and continue discussion?

> Actually when I'm thinking now about this - .raw_env as an environment type of object would be even better.

Why would it be better?

pawelru:
> > I haven't looked into the implementation but I like the change based on the description.
> > I have one idea though. Have you considered having `.raw_data` list instead? So this would be `.raw_data$<dataname>` instead of `.ADSL_raw`. Maybe this will be more elegant to use (and exclude where needed)? Actually when I'm thinking now about this - `.raw_env` as an environment type of object would be even better.
> 
> We haven't consider but it is an interesting idea. I have nothing against and not super hyped about this. Would you like to halt this PR and continue discussion?

I'm not insisting on this particular but it would be good to make a call which way to go so we can close this for good and not leave leftovers.
I see a few benefits and honestly speaking - none of them is a game changer. It's a small enhancements that would make interacting with this a little bit easier and clearer.
- easier to read / write / acccess _raw_ set(s) - no more `paste()` or `sprintf()`
- easier to count objects
- (naming) conventions change occasionally (even here we used to have "ABC_FILTERED" some time in the past) and it would be easier to manage a single collection rather than specific naming convention
- possibly more...

I'm all for this idea but these are not game changers. If it's easy to implement I would encourage to go for it but I won't push on this by any means. 

> > Actually when I'm thinking now about this - .raw_env as an environment type of object would be even better.
> 
> Why would it be better?

I consider environments as an unordered list accessible only by keys (and not index). Conceptually, our collection of raw datasets do not have order so this seems to be more suitable.

gogonzo:
> > > I haven't looked into the implementation but I like the change based on the description.
> > > I have one idea though. Have you considered having `.raw_data` list instead? So this would be `.raw_data$<dataname>` instead of `.ADSL_raw`. Maybe this will be more elegant to use (and exclude where needed)? Actually when I'm thinking now about this - `.raw_env` as an environment type of object would be even better.
> > 
> > 
> > We haven't consider but it is an interesting idea. I have nothing against and not super hyped about this. Would you like to halt this PR and continue discussion?
> 
> I'm not insisting on this particular but it would be good to make a call which way to go so we can close this for good and not leave leftovers. I see a few benefits and honestly speaking - none of them is a game changer. It's a small enhancements that would make interacting with this a little bit easier and clearer.
> 
> * easier to read / write / acccess _raw_ set(s) - no more `paste()` or `sprintf()`
> * easier to count objects
> * (naming) conventions change occasionally (even here we used to have "ABC_FILTERED" some time in the past) and it would be easier to manage a single collection rather than specific naming convention
> * possibly more...
> 
> I'm all for this idea but these are not game changers. If it's easy to implement I would encourage to go for it but I won't push on this by any means.
> 
> > > Actually when I'm thinking now about this - .raw_env as an environment type of object would be even better.
> > 
> > 
> > Why would it be better?
> 
> I consider environments as an unordered list accessible only by keys (and not index). Conceptually, our collection of raw datasets do not have order so this seems to be more suitable.

@pawelru makes sense, I like this suggestion. Environment needs to be locked as it is shared by all `qenv`. Show R code would look like this:

```r
.raw_data <- list2env(list(ADSL = ADSL, ADTTE = ADTTE, iris = iris))
lockEnvironment(.raw_data)
```

instead of 

```r
.ADSL_raw <- ADSL
.ADTTE_raw <- ADSL
.iris_raw <- iris
```

R/teal_data_utils.R
m7pr: Ok, thanks, all clear!

pawelru:
Oh I haven't thought about. That makes it even better!

gogonzo:
@pawelru I'll change `.<dataname>_raw <- <dataname>` to the `environment`. I'll commit Tomorrow or on Thursday

m7pr:
Just played around with it and it looks solid. IMHO cleaner than what we had before in Show R Code (I didn't like the previous behavior where this `._raw` data was spread within multiple lines of Show R Code https://github.com/insightsengineering/teal/issues/1311).

Tested below cases

ADSL and ADTTE as input, ADTTE not in datanames - Show R Code returned only for ADSL

image

ADSL and ADTTE as input, ADTTE not in datanames - Show R Code returned for both

image

gogonzo:
> ADSL and ADTTE as input, ADTTE not in datanames - Show R Code returned only for ADSL
> 
> image

Do you mean this:
- `data` contains `ADSL` and `ADTTE`, and 
- `module$datanames = "ADSL"`

If yes, then it is expected. Code should show only what is related to `module$datanames`

R/module_filter_data.R
```diff
@@ -52,7 +52,15 @@ srv_filter_data <- function(id, datasets, active_datanames, data_rv, is_active)
...
-  data <- eval_code(data, sprintf("%1$s._raw_ <- %1$s", datanames))
+  data <- eval_code(
+    data,
+    paste0(
+      ".raw_data <- list2env(list(",
+      toString(sprintf("%1$s = %1$s", datanames)),
+      "))\n",
+      "lockEnvironment(.raw_data) #@linksto .raw_data" # this is environment and it is shared by qenvs. CAN'T MODIFY!
+    )
+  )
```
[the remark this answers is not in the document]
gogonzo: > can we work on expressions please instead of strings - that does not look pro

We can't because `#@linksto` comment will be discarded when turned into expression. 
Also creating an expression `.raw_data <- list2env(list(<datanames> = <datanames>))` where datanames has `length > 1` will probably involve some `str2lang` or `parse(text = )`

> can we make a single expr lockEnvir(list2env(...)) - just to limit the technical clutter in SRC from two lines into a single one

we need to assign `.raw_data` in this env. If you wish it, I can make `lockEnvironment(.raw_data <- ...)`. But this is  even clutter-ous.

P.S. What I think instead we can give up keeping this code at all. But this is a danger zone as we provide object which is not "documented" by the code. Theoretically, one can use object `.raw_data` in the module to build output. I was thinking that maybe `get_var.qenv` and `[[.qenv` shouldn't allow extracting hidden objects prefixed by `.`. Then the whole concept of not documenting `.<hidden object name>` would make sense. But this would involve also changing `validate(<qenv>)` function, which checks `@code => @env`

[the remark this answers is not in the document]
pawelru: I'm sorry for my naive question - why do we care about `#@linksto` comment?

> where datanames has length > 1 will probably involve some str2lang or parse(text = )

This is exactly the use case where rlang shines but I do understand that you want to stick to the base. ok let's leave it

> But this is even clutter-ous.

Agree - let's leave it.

> P.S.

I see your point here but on the other hand I do like how simple qenv is. For me `get_var.qenv` is the same as `base::get()`. That _hidding_ dot vars is happening only in `ls()` but not in `get()`.
Not to mention changes in validate.
Let's keep it simple and eliminate workarounds.
If we are thinking that far - I would be going in the direction of storing these raw data outside of qenv. AFAIK these are needed for filtering so it would make sense to keep it in the dataset-filter instance.
Also, I don't think we need _complete_ datasets with raw values. We might only need colnames, distribution data, etc. But this is a complitely separate topic.

m7pr:
> If yes, then it is expected. Code should show only what is related to module$datanames

Yes, was just trying to prove that what is expected is actually displayed :)

R/module_filter_data.R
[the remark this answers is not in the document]
m7pr: > I'm sorry for my naive question - why do we care about #@linksto comment?

So that this is line 

```
lockEnvironment(.raw_data)
````

is extracted when you ask for the [`teal.data::get_code(data, datanames =`](https://github.com/insightsengineering/teal/pull/1342/files#diff-4f959897d64b51a6ac57c9016430bdd92163a17fa62b5a61707a4b3bc7dd6325R59) for raw objects.

When you specify datanames, then all the code needed to reproduce this object is returned. If it's hard to determine lines that have an affect on the object, because this is the case of a line that only make a side-effect on the object, then you need to mark this line with `#@linksto` tag. Otherwise side-effects made on those objects will not be returned with the code.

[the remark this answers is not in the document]
Comment (pawelru) [reply R/module_filter_data.R:63]: Thanks. It's all connected to my very old issue [here](https://github.com/insightsengineering/teal.code/issues/181). which reflect my thinking that the code is a code but it's a string instead. I look into the [docs](https://insightsengineering.github.io/teal.data/latest-tag/reference/get_code.html#extracting-dataset-specific-code) and it's a functionality built on the top of it. 

Honestly, I'm getting more and more concerned as there are more and more functionalities built on the top of that questionable implementation but this is definitely going beyond that PR. All good here.

**Comment by pawelru** [comment]
Thanks 👍  Conceptually it is looking good. It's an soft approve as a comment from my side as I haven't check the details of the implementation

File: R/module_data_summary.R [hunk 5]
```diff
@@ -141,16 +140,13 @@ srv_data_summary <- function(id, teal_data) {
...
     datanames,
     function(name) teal.code::get_env(teal_data())[[name]],
     simplify = FALSE
   )
-  unfiltered_data_objs <- sapply(
-    datanames,
-    function(name) teal.code::get_env(teal_data())[[paste0(name, "._raw_")]],
-    simplify = FALSE
-  )
+  unfiltered_data_objs <- teal.code::get_env(teal_data())[[".raw_data"]]
```
pawelru, line 149: if we already using API from `teal.code`, maybe `teal.code::get_var()`?
not a big deal though

R/module_init_data.R
```diff
@@ -151,7 +154,7 @@ srv_init_data <- function(id, data, modules, filter = teal_slices()) {
       list(code = trimws(c(teal.code::get_code(data), hashes), which = "right")),
       list(join_keys = teal.data::join_keys(data)),
       sapply(
-        ls(teal.code::get_env(data)),
+        .teal_data_ls(data),
```
pawelru, line 157: why this new function?
If you are doing `ls()` this should return all the names - no need to explicitly append parents
you might want to reorder the objects but if we are in the environment then the order does not matter (which is a big advantage of the envir compared to list)
I think it should be fine to stick to `ls(..., all.names = FALSE)` and remove `.teal_data_ls()` at all

R/teal_data_utils.R
```diff
@@ -42,35 +42,37 @@ NULL
...
   checkmate::assert_class(data, "teal_data")
   checkmate::assert_class(datanames, "character")
-  datanames_corrected <- intersect(datanames, ls(data@env))
-  dataname_corrected_with_raw <- intersect(c(datanames, sprintf("%s._raw_", datanames)), ls(data@env))
-
+  datanames_corrected <- intersect(datanames, .teal_data_ls(data))
+  datanames_corrected_with_raw <- c(datanames_corrected, ".raw_data")
   if (!length(datanames)) {
     return(teal_data())
   }
```
pawelru, line 49: this should be run before `datanames_corrected`

R/teal_data_utils.R
```diff
@@ -42,35 +42,37 @@ NULL
...
-      mget(x = dataname_corrected_with_raw, envir = data@env),
+      mget(x = datanames_corrected_with_raw, envir = teal.code::get_env(data)),
       list(
         code = gsub(
           "warning('Code was not verified for reproducibility.')\n",
           "",
-          teal.data::get_code(data, datanames = dataname_corrected_with_raw),
+          teal.data::get_code(data, datanames = datanames_corrected_with_raw),
           fixed = TRUE
         ),
```
pawelru, line 61: a nasty coupling here
to be addressed outside of this PR

R/teal_data_utils.R
```diff
@@ -42,35 +42,37 @@ NULL
 .subset_teal_data <- function(data, datanames) {
   checkmate::assert_class(data, "teal_data")
   checkmate::assert_class(datanames, "character")
-  datanames_corrected <- intersect(datanames, ls(data@env))
-  dataname_corrected_with_raw <- intersect(c(datanames, sprintf("%s._raw_", datanames)), ls(data@env))
-
+  datanames_corrected <- intersect(datanames, .teal_data_ls(data))
+  datanames_corrected_with_raw <- c(datanames_corrected, ".raw_data")
```
pawelru, line 46: 1. I'm thinking that this function should be part of `teal_data` class API.

2. I'm thinking about possible future extensions of this.
Can we make this `ls(envir = teal.code::get_env(data), pattern = "\\.", all.names = TRUE)` (plus change the var name).
Combining this with 1), `teal` (or anyone else) would be able to internal stuff with `.xyz` and `teal.data` would require any changes

R/teal_data_utils.R
pawelru: could be avoided with:
- moving this to teal.data - see: https://github.com/insightsengineering/teal/pull/1342/files#r1772857752
- add argument to `teal.data::get_code` (in particular: not `teal.code::get_code()`

tests/testthat/test-module_teal.R
```diff
@@ -1,6 +1,6 @@
 # comment: srv_teal is exported so the tests here are extensive and cover srv_data as well.
 #          testing of srv_data is not needed.
-module_output_table <<- function(output, id) {
+module_summary_table <<- function(output, id) {
```
pawelru, line 3: you haven't changed this but out of curiosity: why double assignment here?

pawelru (commented):
Please find my comments.
Overall it's good and I very much like the change. This introduces much simpler and nicer convention which is much easier to extend in the future. Very nicely done.

I submitted a few small comments about implementation. I'm aware that some of them are to be addressed outside of this PR but still I decided to add them not to loose my thought on this. Please open a new issue or PR if you agree and want to do this separately.

R/module_init_data.R
gogonzo: > I think it should be fine to stick to ls(..., all.names = FALSE) and remove .teal_data_ls() at all

It is not enough. I've created `.teal_data_ls()` because it does topological sort - to display datasets in the filter panel.

tests/testthat/test-module_teal.R
gogonzo: `testthat::describe/testthat::it` evaluates in the own environment which is the sibling of the environment from `test-<filename>.R`. These testing functions are forcing you to test only public objects so even in tests it is required to somehow deliver helper function.

We choose to use `describe/it` instead of `test_that` to provide end-to-end testing and not to test private utilities (which are evaluated anyway within e2e test)

R/module_init_data.R
pawelru: understood 
I totally got that this functionality must to be there but this is not very clear to me that this is the place responsible for such reorder. Maybe we should move it somewhere else closer to the filter panel
the description for this function is that this is responsible for signature protection

gogonzo: Problem is that not only filter-panel displays datanames in topological order but also summary table (which is the part of teal now). I can move `.include_parent_datanames` away from `.teal_data_ls` but it will probably always go together. 

_`.include_parent_datanames` adds parents if missing and orders topologically_

R/teal_data_utils.R
gogonzo: > I'm thinking that this function should be part of teal_data class API.

I have an idea that `@datanames` shouldn't exist and we should read the datanames (names) straight from the environment (except prefixed by `.`). It means that we can have `teal_data_ls` method registered in `teal.code`, with extension method `teal_data_ls.teal_data` which can handle topological sort (because join keys are teal_data property).

See the issue https://github.com/insightsengineering/teal.data/issues/333

> I'm thinking about possible future extensions of this.

I don't understand it yet.

R/teal_data_utils.R
gogonzo: > moving this to teal.data - see: https://github.com/insightsengineering/teal/pull/1342/files#r1772857752
 
 I don't see the possibility to move this to `teal.data`. `datanames` argument is fed by `module$datanames` + `teal_transform_module@datanames`.

>  add argument to teal.data::get_code (in particular: not teal.code::get_code()

What argument in `get_code`? I don't think that `get_code` needs any argument. Theoretically, `get_code(<teal_data>, datanames = <names of the objects>)` accepts any name even those prefixed by `.`.

R/module_init_data.R
gogonzo: I've just have removed `.teal_data_ls` in favour of decoupled `ls()` + `.topologically_sort_datanames()`

R/module_data_summary.R
gogonzo: addressed

R/teal_data_utils.R
gogonzo: addressed

R/teal_data_utils.R
gogonzo: Please note also that dataname with `.` prefix is problematic. In `teal.slice` we have a `check_simple_name` which throws when dataname doesn't start with letter and doesn't contain only alphanumeric. [see here](https://github.com/insightsengineering/teal.slice/blob/aa05f7feb26bb32fbb5f069ec3e9e77d17450022/R/utils.R#L14)
This limitation is set because of modules (namespaces) are using datanames as an id. Dots will cause the problems with css selectors. Therefore I think dot-prefixed datanames should be avoided.

pawelru: > I have an idea that `@datanames` shouldn't exist

I'm all for it!

- no dataname slot - makes it simpler and reduce the clutter code made by end-user on the preprocessing part 
- topological sort functionality should live in the teal-data class

> It means that we can have teal_data_ls method registered in teal.code

One question tough. Can we avoid extending the base class in teal.code and apply extension only in teal.data? I would like to avoid edits in teal.code. It's good as is.

> > I'm thinking about possible future extensions of this.
> 
> I don't understand it yet.

I do imagine us having a need in the future to push yet another purely internal objects into the quenv. Right now we are excluding `.raw_data` value only. If we would exclude everything with dot upfront then this functionality would not require any edits. We would nicely follow open-close principle - open for extensions (i.e. "in order to create internal vars please do this with dot-name") but close for modifications (no need to further edits to support that).

> Please note also that dataname with . prefix is problematic.

Yes I am aware. It very connected to my old issue [here](https://github.com/insightsengineering/teal/issues/969). I guess we won't solve it here. But I think we should be good on this. I think that we are on the same page that `get_datanames()` functionality (however its implemented) should exclude datanames starting with dot. That check function you referenced should be called after that filter.

R/teal_data_utils.R
pawelru: I was thinking about the following: `get_code(<teal_data>, datanames = ..., include_warnings = FALSE)`.

The warning code is added inside `teal.data` so `teal.data` should be also responsible for removing it. We should ask ourselves a question: what if the warning message got changed for some reason? Right now we would have to process this change in two repos. We should avoid such coupling. This is the type of the things that I would always trying to point out.
If you don't agree on extending `get_code()` - maybe the whole `subset_teal_data()` should live in teal.data?

gogonzo: Maybe warning shouldn't be produced at all in `qenv`? `teal` adds `stopifnot` anyway so it will throw an error anyway id datasets are not reproducible.

R/teal_data_utils.R
gogonzo: responded in the issue https://github.com/insightsengineering/teal.data/issues/333#issuecomment-2376056555
I'm resolving this conversation here

R/teal_data_utils.R
pawelru: Even better! qenv should return an analysis / datasets source code. This warning belongs to neither of them.

gogonzo: here I'm removing warning message:
https://github.com/insightsengineering/teal.data/pull/334

pawelru (approved):

PR #1343, opened 2024-09-10
vbump teal.logger

pawelru:
teal.logger had been already released so there is no point of depending on the old development version

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -52,7 +52,7 @@ Imports:
     shinyjs,
     stats,
     teal.code (>= 0.5.0),
-    teal.logger (>= 0.1.3.9013),
+    teal.logger (>= 0.2.0),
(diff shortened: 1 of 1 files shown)


donyunardi:
Hi @pawelru 
Now that we're done with [this teal.logger PR](https://github.com/insightsengineering/teal.logger/pull/86), let's release teal.logger first and then we'll update teal to have the released version of teal.logger.

I'll prep this today.

pawelru:
Hey @donyunardi - shall we wait maybe? I would propose to keep that references changes in devel state of teal.logger so that we would have some time to discover some edge cases (if exists) that hasn't been spotted in the review process. 
I have a mirai PR in my head where devs identified issues after merge to main.

donyunardi (approved):

Issue #1276, opened 2024-07-29
Possibility to disable lockfile creation


pawelru:
In the projects like TLG-Catalog, where we produce a lot of small apps for WebR
I want to be able to disable lockfile creation

m7pr:
@pawelru if this is urgent, as a quick fix for now you can provide a path to an empty file via `options('teal.renv.lockfile')`

m7pr:
To disable lockfile creation this function should return `NULL` https://github.com/insightsengineering/teal/pull/1263/files#diff-c389a151f2967297002c1253ee7819f95d9017f250147356c6e8d6d66165638cR59-R73
We could set this up with an R option via `options`. `options('teal.renv.create = TRUE)'` by default, that can be set to `FALSE`?

pawelru:
Yes. That's what I was thinking of. A new option would have to be documented properly of course.

pawelru:
I am observing more and more unusual times when `teal_app.lock` file is being generated:
- rendering quarto article with teal app code
- running examples with `devtools::run_examples()`
- my experimental roxygen roclet running examples

That all makes me think that we should be more restrictive on when to trigger creating this file.

m7pr:
@pawelru we can always switch to the case when this is run on demand by a parameter in init. `init(lockfile = FALSE)` 

```r
@param lockfile `(logical)` whether to generate the lockfile for the app. by default FALSE
```

This will be easier to turn it on, then detect places where it it problematic

pawelru:
I honestly have to admit that I am not super convinced to do this. I do really like how simple `init()` is. For me, all the arguments needs to be of the same _importance_ level. Now let's compare that new argument against `data` or `modules` to see what I am concerned about.
I don't really know how to proceed. This might be indeed good to allow to control but don't want to disrupt current simplicity of init

m7pr:
So we will end up controlling everything with `options()` because we don't like arguments : ) and that ain't gonna be simple, because you need extra documentation for `options` used in teal, and that is not very popular way of modifying functions' behaviors (in comparison to arguments usage). We can introduce one meta argument, that takes a list or arguments as an input, and that's gonna be the place for the placement of all custom low-important parameters?
Still thinking out loud. Not necessarily pushing any idea

gogonzo:
Cases when we should include lockfile process:
- a running app (normal shiny app) on
- interactive session (for example by running testServer by hand) on
- r cmd check/tests off
- devtools::run examples off
[...]

PR #1346, opened 2024-09-17
disable lockfile generation in `callr`, `testthat::tests` and `R CMD CHECK`

m7pr:
Part of https://github.com/insightsengineering/teal/pull/1263 and closes https://github.com/insightsengineering/teal/issues/1276

Local results of `shinytest2` tests

```r
══ Results ══════════════════════════════════════════════════════════════════════════
Duration: 540.8 s

── Skipped tests (6) ────────────────────────────────────────────────────────────────
• need a fix in a .slicesGlobal (1): test-module_teal.R:1139:11
• todo (5): test-module_teal.R:1404:7, test-module_teal.R:1411:5,
  test-module_teal.R:1414:5, test-module_teal.R:1673:5, test-module_teal.R:1731:5

[ FAIL 0 | WARN 0 | SKIP 6 | PASS 515 ]
```

### DIFF
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -5,6 +5,7 @@
     !(
       identical(Sys.getenv("CALLR_IS_RUNNING"), "true") || # inside callr process
       identical(Sys.getenv("TESTTHAT"), "true") || # inside devtools::test
+      !identical(Sys.getenv("QUARTO_PROJECT_ROOT"), "") || # inside Quarto process
       (
         ("CheckExEnv" %in% search()) || any(c("_R_CHECK_TIMINGS_", "_R_CHECK_LICENSE_") %in% names(Sys.getenv()))
       ) # inside R CMD CHECK
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -1,8 +1,18 @@
 .onLoad <- function(libname, pkgname) {
   # adapted from https://github.com/r-lib/devtools/blob/master/R/zzz.R
+
+  lockfile_flag <-
+    !(
+      identical(Sys.getenv("CALLR_IS_RUNNING"), "true") || # inside callr process
+      identical(Sys.getenv("TESTTHAT"), "true") || # inside devtools::test
+      (
+        ("CheckExEnv" %in% search()) || any(c("_R_CHECK_TIMINGS_", "_R_CHECK_LICENSE_") %in% names(Sys.getenv()))
+      ) # inside R CMD CHECK
+    )
+
   teal_default_options <- list(
     teal.show_js_log = FALSE,
-    teal.lockfile.enable = TRUE,
+    teal.lockfile.enable = lockfile_flag,
     shiny.sanitize.errors = FALSE
   )
 
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -1,19 +1,8 @@
 .onLoad <- function(libname, pkgname) {
   # adapted from https://github.com/r-lib/devtools/blob/master/R/zzz.R
 
-  lockfile_flag <-
-    !(
-      identical(Sys.getenv("CALLR_IS_RUNNING"), "true") || # inside callr process
-      identical(Sys.getenv("TESTTHAT"), "true") || # inside devtools::test
-      !identical(Sys.getenv("QUARTO_PROJECT_ROOT"), "") || # inside Quarto process
-      (
-        ("CheckExEnv" %in% search()) || any(c("_R_CHECK_TIMINGS_", "_R_CHECK_LICENSE_") %in% names(Sys.getenv()))
-      ) # inside R CMD CHECK
-    )
-
   teal_default_options <- list(
     teal.show_js_log = FALSE,
-    teal.lockfile.enable = lockfile_flag,
     shiny.sanitize.errors = FALSE
   )
 
--- a/R/module_teal_lockfile.R
+++ b/R/module_teal_lockfile.R
@@ -75,8 +75,6 @@ srv_teal_lockfile <- function(id) {
 
     lockfile_path <- "teal_app.lock"
     is_lockfile_enabled <- .is_lockfile_enabled()
-    cat('\n is_lockfile_enabled', is_lockfile_enabled, '\n')
-    cat('\n identical(Sys.getenv("TESTTHAT"), "true")', identical(Sys.getenv("TESTTHAT"), "true"), '\n')
     user_lockfile_path <- getOption("teal.lockfile.path", default = "")
     is_user_lockfile_set <- !identical(user_lockfile_path, "")
 
--- a/R/module_teal_lockfile.R
+++ b/R/module_teal_lockfile.R
@@ -74,7 +74,9 @@ srv_teal_lockfile <- function(id) {
     })
 
     lockfile_path <- "teal_app.lock"
-    is_lockfile_enabled <- isTRUE(getOption("teal.lockfile.enable"))
+    is_lockfile_enabled <- .is_lockfile_enabled()
+    cat('\n is_lockfile_enabled', is_lockfile_enabled, '\n')
+    cat('\n identical(Sys.getenv("TESTTHAT"), "true")', identical(Sys.getenv("TESTTHAT"), "true"), '\n')
     user_lockfile_path <- getOption("teal.lockfile.path", default = "")
     is_user_lockfile_set <- !identical(user_lockfile_path, "")
 
@@ -193,4 +195,20 @@ utils::globalVariables(c("opts", "sysenv", "libpaths", "wd", "lockfilepath", "ru
 #' @rdname module_teal_lockfile
 .is_lockfile_deps_installed <- function() {
   requireNamespace("mirai", quietly = TRUE) && requireNamespace("renv", quietly = TRUE)
+}
+
+#' @rdname module_teal_lockfile
+.is_lockfile_enabled <- function() {
+  if (isTRUE(getOption("teal.lockfile.enable"))) {
+    return(TRUE)
+  } else {
+    !(
+      identical(Sys.getenv("CALLR_IS_RUNNING"), "true") || # inside callr process
+      identical(Sys.getenv("TESTTHAT"), "true") || # inside devtools::test
+      !identical(Sys.getenv("QUARTO_PROJECT_ROOT"), "") || # inside Quarto process
+      (
+        ("CheckExEnv" %in% search()) || any(c("_R_CHECK_TIMINGS_", "_R_CHECK_LICENSE_") %in% names(Sys.getenv()))
+      ) # inside R CMD CHECK
+    )
+  }
 }
--- a/tests/testthat/setup-options.R
+++ b/tests/testthat/setup-options.R
@@ -19,3 +19,7 @@ if (isFALSE(getFromNamespace("on_cran", "testthat")()) && requireNamespace("with
     .local_envir = testthat::teardown_env()
   )
 }
+
+cat('\nIS_TESTING', Sys.getenv("TESTTHAT"), '\n')
+cat('\n teal.lockfile.enable: ',
+    getOption('teal.lockfile.enable'), '\n')
--- a/tests/testthat/setup-options.R
+++ b/tests/testthat/setup-options.R
@@ -19,7 +19,3 @@ if (isFALSE(getFromNamespace("on_cran", "testthat")()) && requireNamespace("with
     .local_envir = testthat::teardown_env()
   )
 }
-
-cat('\nIS_TESTING', Sys.getenv("TESTTHAT"), '\n')
-cat('\n teal.lockfile.enable: ',
-    getOption('teal.lockfile.enable'), '\n')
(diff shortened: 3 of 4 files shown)


gogonzo:
It does seem to work. I've checked `devtools::run_examples` and `pkgdown::build_site` and I've not seen any lockfile created. Can you confirm that https://github.com/insightsengineering/tlg-catalog/ rendering doesn't create any lockfile?

m7pr:
> Can you confirm that https://github.com/insightsengineering/tlg-catalog/ rendering doesn't create any lockfile?

Will try to build in a sec


> It does seem to work. I've checked devtools::run_examples and pkgdown::build_site and I've not seen any lockfile created. 


Does `devtools::run_examples(fresh = TRUE)` not show any lockfiles because we do have this `if` statement around all our apps in examples?

```r
if (interactive()) {
  shinyApp(ui, server)
}
```

When I run `devtools::run_examples(fresh = FALSE)` it runs in my `interactive()` session and I do see `lockfile` being created

m7pr:
Sorry for the delay, I needed to install Quarto, `webshot2`, `shinylive`, `quatro` and render the book. Trying to render it locally

gogonzo:
> When I run `devtools::run_examples(fresh = FALSE)` it runs in my `interactive()` session and I do see `lockfile` being created

Then `options(teal.lockfine.enable)` is not set properly I guess. `devtools::run_examples(fresh = TRUE)` [runs callr::r](https://github.com/r-lib/devtools/blob/f0dccc8a069adffc8507fe82436a584338fc2552/R/run-examples.R#L47
) - theoretically lockfile shouldn't be triggered.

m7pr:
For `devtools::run_examples(fresh = TRUE)` I don't see lockfile being created (or maybe it happens too quickly and then it immediately gets deleted), even on `479_mirai_lockfile@main`. Let's talk about it tomorrow for a quick moment after the daily.

I hope I will get quarto book rendered sorted out during this time. For now I am experiencing an issue with the installation of GitHub packages.

m7pr:
Ok, for `devtools::run_examples(fresh = TRUE)` we had a discussion in which we agreed that if we run it, and have all examples covered in `if (interactive() )` then no teal session will be started, hence to teal lockfile generated.

m7pr:
Hey @pawelru and @cicdguy 

In this PR we try to disable one of the functionalities of teal, which is lockfile creation. We listed out scenarios in which we would like to disable the option https://github.com/insightsengineering/teal/issues/1276#issuecomment-2288684906

We are trying to figure out what elements are covered in tlg-catalog Quarto book creation so that at some parts this feature can be disabled as well. I am trying to figure out the flow of Quarto (that uses webR and uses shinylive) for tlg-catalog rendering. Would you be able to help me explain the process?

Based on the example:
- graphs -> efficacy -> fstg01
- url https://insightsengineering.github.io/tlg-catalog/stable/graphs/efficacy/fstg01.html
- and the corresponding file https://github.com/insightsengineering/tlg-catalog/blob/main/book/graphs/efficacy/fstg01.qmd
- that uses some `_utils` templates https://github.com/insightsengineering/tlg-catalog/tree/main/book/_utils

here is what I understand is happening in tlg-catalog.

Quarto is used to render the whole book/website.
Each page (in this case efficacy -> fstg01) contains of few tabs. 

There are tabs that just have data preparations and plot generation - those have a preview section and webR section.
Preview is responsible for showing the code and the output, webR is responsible for running R live in browser.
Here no teal apps are created.
image

There is also a tab called `teal app` - this have a preview section and shinylive section.
Preview is responsible for showing the code and the **screenshot** of the app.
shinylive section is responsible for running **teal** app in it.
teal app is not started before you hit shinylive section, however in the preview section there is already some screenshot of the app.

preview
image

shinylive
image

So for the app that is started in shinylive section, we would like to show teal lockfile. However for the moment when the Quarto page is generated/rendered and the screenshot for the preview section is made, we would not like to create a lockfile. I am trying to figure out how the screenshot is made, and whether we needed to run the teal app to create the screenshot. I understand that screenshot is made with `webshot2::webshot(app_url)` but someone needed to somehow somewhere run the app so that `app_url` is exposed. I assume teal app is run during Quarto rendering.

If the teal app is run inside Quarto process just to create the screenshot, we can easily detect that Quarto is running by checking `Sys.getenv("QUARTO_PROJECT_ROOT")` variable and disabling lockfile creation of this is set. I am just trying to verify if I understand things right.

m7pr:
Ok, I think I was able to understand the process based on a small test Quarto file with teal app.

<details><summary> Quarto file with teal app </summary></details>

---
title: Test
---

------------------------------------------------------------------------


```{r}
knitr::opts_template$set(
  app = list(
    webshot = "webshot2",
    screenshot.force = TRUE,
    screenshot.opts = list(delay = 5),
    dev = "png",
    fig.width = 15,
    cache = FALSE,
    message = FALSE,
    R.options = list(
      teal.log_level = "TRACE"
    )
  )
)
```


## `teal` App

```{r teal, opts.label = c("app")}
library(teal.modules.clinical)

## Data reproducible code
data <- teal_data()
data <- within(data, {
  ADSL <- random.cdisc.data::cadsl
  ADRS <- random.cdisc.data::cadrs
})
datanames <- c("ADSL", "ADRS")
datanames(data) <- datanames
join_keys(data) <- default_cdisc_join_keys[datanames]

arm_ref_comp <- list(
  ARM = list(
    ref = "B: Placebo",
    comp = c("A: Drug X", "C: Combination")
  ),
  ARMCD = list(
    ref = "ARM B",
    comp = c("ARM A", "ARM C")
  )
)

## Reusable Configuration For Modules
ADSL <- data[["ADSL"]]
ADRS <- data[["ADRS"]]

## Setup App
app <- init(
  data = data,
  modules = modules(
    tm_g_forest_rsp(
      label = "Forest Response",
      dataname = "ADRS",
      arm_var = choices_selected(
        variable_choices(ADSL, c("ARM", "ARMCD")),
        "ARMCD"
      ),
      arm_ref_comp = arm_ref_comp,
      paramcd = choices_selected(
        value_choices(ADRS, "PARAMCD", "PARAM"),
        "BESRSPI"
      ),
      subgroup_var = choices_selected(
        variable_choices(ADSL, names(ADSL)),
        c("BMRKR2", "SEX")
      ),
      strata_var = choices_selected(
        variable_choices(ADSL, c("STRATA1", "STRATA2")),
        "STRATA2"
      ),
      plot_height = c(600L, 200L, 2000L),
      plot_width = c(1100L, 200L, 2000L)
    )
  )
)

shinyApp(app$ui, app$server)
```

Quarto renders the file and starts teal app, once the chunk is finished it is possible to create a snapshot/webshot of the app that was run. You can see in the logs that the whole app was run.

So to omit teal lockfile creation during tlg-creation we need to disable it based on a Quarto running flag
`Sys.getenv("QUARTO_PROJECT_ROOT")` 

</details>

m7pr:
Yeah, for Quarto you can see it was possible to disable lockfile with the last commit https://github.com/insightsengineering/teal/pull/1346/commits/d54cb80db9180629e88654e45fc828326e18100c

# Without the flag it was running the lockfile

on_main


# With the flag it is not running the lockfile

on_feature_branch_based_on_quarto_flag

R/zzz.R
pawelru, line 1: Thanks. This is going in the right direction. But I think we also need to think _when_ to determine whether to enable/disable lockfile creation.

Right now it is on `teal` load. This might introduce some limitation. My challenge is: should it be run dynamically somewhere in `teal::init()`?

pawelru: after I while I think it's good
the whole logic is indeed inside the init() - here is just an option which an input to that logic. One might choose to modify this

m7pr: Well, currently it sets an `option` during `onLoad`. Then you can edit the option whenever you want. You know the option, before you run `teal::init()`  and you can always check it. In `teal::init()` the option is checked and the procedure is disabled/enabled.

I think this gives enough flexibility to disable/enable

```r
library(teal) # sets it for you depending on circumnstances so you don't need to set on your own in multiple places
options(teal.lockfile.enable = TRUE/FALSE) # if you really need to set it on your own, use one-liner
app <- teal::init() # does not yet create the lockfile
shinyApp(app$ui, app$server) # determines whether to create the lockfile or not
```

pawelru, line 1: A new question: what is happening if you run `pkgdown::build_site()`?

FYI: there is a `pkgdown::in_pkgdown()`

R/zzz.R
```diff
@@ -1,8 +1,19 @@
 .onLoad <- function(libname, pkgname) {
   # adapted from https://github.com/r-lib/devtools/blob/master/R/zzz.R
+
+  lockfile_flag <-
+    !(
+      identical(Sys.getenv("CALLR_IS_RUNNING"), "true") || # inside callr process
+      identical(Sys.getenv("TESTTHAT"), "true") || # inside devtools::test
```
pawelru, line 7: a question
what would happen if we do `load_all()` and then `test()`? 
Do I understand correctly that the option would be set to true and then lockfile created inside the tests?

m7pr: Let me check, good question

R/zzz.R
m7pr: I don't see any lockfile being created for `teal` because all our examples are wrapped inside

`if (interactive()) shinyApp(app$ui, app$server)`

in examples and vignettes. 

I can imagine this can different for some other packages, so I would include

```r
identical(Sys.getenv("IN_PKGDOWN"), "true")
```

in the check just in case.

WDYT?

R/zzz.R
m7pr: I actually think that `Sys.getenv("TESTTHAT")` flag is set after the package is loaded during tests, and the initial value of `getOption('teal.lockfile.enable')` is set to `TRUE` during tests. Not as I would expect it to work

I added this snippet to check the behavior of this flag during checks.

```r
cat('\n teal.lockfile.enable: ', getOption('teal.lockfile.enable'), '\n')
```

This made me think we actually need to disable the creation of the lockfile inside `srv_teal_lockfile`, if `Sys.getenv("TESTTHAT")` gets set later after the package is loaded

image

m7pr: `Sys.getenv("TESTTHAT")` is set at `testthat::test_that`, not at `devtools::test()`

image

R/zzz.R
pawelru: Can you please try this in https://github.com/insightsengineering/teal.modules.general/pull/775 (it would require `insightsengineering/roxy.shinylive@init`)

Re your suggestion: probably yes but let's try this first to confirm.

R/zzz.R
pawelru: > This made me think we actually need to disable the creation of the lockfile inside srv_teal_lockfile, if Sys.getenv("TESTTHAT") gets set later after the package is loaded

I got a similar idea. And I was thinking about this particular example in my previous comment [here](https://github.com/insightsengineering/teal/pull/1346#discussion_r1764903472)

R/zzz.R
m7pr: Yes, it did create `teal_app.lock` lockfile. So it's needed to include `identical(Sys.getenv("IN_PKGDOWN"), "true")`

R/zzz.R
m7pr: With this commit https://github.com/insightsengineering/teal/pull/1346/commits/18620432f1db0de5d99ca23922dd80a713615be8 I moved the functionality inside `srv_teal_lockfile` so the checks are made on lockfile execution.
If the option was set before the lockfile was called, then the option is superior to the circumstances check

m7pr:
Current state of test

```r
══ Results ══════════════════════════════════════════════════════════════════════════════
Duration: 564.7 s

── Skipped tests (6) ────────────────────────────────────────────────────────────────────
• need a fix in a .slicesGlobal (1): test-module_teal.R:1139:11
• todo (5): test-module_teal.R:1404:7, test-module_teal.R:1411:5,
  test-module_teal.R:1414:5, test-module_teal.R:1673:5, test-module_teal.R:1731:5

[ FAIL 0 | WARN 0 | SKIP 6 | PASS 515 ]
```

R/zzz.R
gogonzo: `devtools::build_vignettes()` also doesn't trigger lockfile creation

gogonzo:
I've tested using `Rscript -e "Sys.setenv(TESTING_DEPTH=5)"` and lockfile is not created in both mirai PR and this PR:


|                             | @mirai | @this |
|-----------------------------|--------|-------|
| devtools::run_examples()    | ✅      | ✅     |
| devtools::build_vignettes() | ✅      | ✅     |
| devtools::test()            | ✅      | ✅     |
| devtools::check()           | ✅      | ✅     |
| pkgdown::build_site()       | ✅      | ✅     |
| roxy.shinylive of tmg (install docs render and view)              | ✅      | ✅     |

I didn't check tlg-catalog. 

NOTE: In `roxy.shinylive` apps lockfile download is hidden, I assume `renv` is not available in shinylive (maybe is not compiled?).

m7pr:
I've tested manually Quarto and callr and it is improved with this PR

|                             | @mirai | @this |
|-----------------------------|--------|-------|
| quarto render     | ❌       | ✅     |
| callr | ❌       | ✅     |

Looks like we are good to go

gogonzo (approved):
Let's continue on the feature branch

m7pr:
Alrighty, merging into feature branch

PR #1355, opened 2024-09-27
increase timeout for ci

donyunardi:
Closes https://github.com/insightsengineering/coredev-tasks/issues/581

Adding maximum timeout to 500 seconds and change duration every 0.5 seconds when checking lockfile existence.


gogonzo (approved):
👍

PR #1363, opened 2024-10-04
Fix Dependency Test - max

### LINKED ISSUE #1362: Dependency Test - max failed
https://github.com/insightsengineering/teal/actions/runs/11172110749/job/31058830842

```
  ══ Failed tests ════════════════════════════════════════════════════════════════
  ── Failure ('test-module_teal.R:90:13'): creation process is invoked for teal.lockfile.mode = "enabled" and snapshot is copied to teal_app.lock and removed after session ended ──
  
  `actual`:   FALSE
  `expected`: TRUE 
```

This is similar fail with our daily integration test and we have already increased the timeout [here](https://github.com/insightsengineering/teal/pull/1355) to 500 seconds.

To test this, I'm going to see if increasing the timeout will help resolve this.

donyunardi:
Fixes #1362 

Increase iteration time to 1000 seconds by increasing sleep time

### DIFF
--- a/tests/testthat/test-module_teal.R
+++ b/tests/testthat/test-module_teal.R
@@ -83,7 +83,7 @@ testthat::describe("srv_teal lockfile", {
           ),
           expr = {
             iter <- 1
-            while (!file.exists(renv_filename) && iter <= 1000) {
+            while (!file.exists(renv_filename) && iter <= 2000) {
               Sys.sleep(1)
               iter <- iter + 1 # max wait time is 500 seconds
             }
--- a/tests/testthat/test-module_teal.R
+++ b/tests/testthat/test-module_teal.R
@@ -84,7 +84,7 @@ testthat::describe("srv_teal lockfile", {
           expr = {
             iter <- 1
             while (!file.exists(renv_filename) && iter <= 1000) {
-              Sys.sleep(0.5)
+              Sys.sleep(1)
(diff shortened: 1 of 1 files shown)


donyunardi:
Still fails:
https://github.com/insightsengineering/teal/actions/runs/11184035499/job/31093937769

One last time, update to 2000 seconds.

donyunardi:
Still fails after update to 2000 seconds
https://github.com/insightsengineering/teal/actions/runs/11185274030/job/31097779397

tests/testthat/test-module_teal.R
```diff
@@ -83,8 +83,8 @@ testthat::describe("srv_teal lockfile", {
           ),
           expr = {
             iter <- 1
-            while (!file.exists(renv_filename) && iter <= 1000) {
-              Sys.sleep(0.5)
+            while (!file.exists(renv_filename) && iter <= 2000) {
+              Sys.sleep(1)
```
m7pr, line 87: So you extended the total wait time from 0.5*1000 seconds (8.3 minutes) to 1*2000 seconds (33 minutes) and it still does not work : ) ?

If it does not work for 8 minutes, it will not work at all. lockfile should take about 10-60 seconds in this case.

donyunardi: Okay, let's investigate this more.

PR #1367, opened 2024-10-08
Fix the datanames bug causing teal app to crash when `teal_data` has non standard object names

### LINKED ISSUE #1366: [Bug]: Teal app crashes when the input teal_data has objects that fail `check_simple_name()`
### What happened?

Please use the `main` branch version of packages.

This issue started after we [made the change](https://github.com/insightsengineering/teal/pull/1334) to pass all the objects from the `teal_data` in certain scenarios.
The actual error happens in `teal.slice` where [all the object names are recieved](https://github.com/insightsengineering/teal.slice/blob/aa05f7feb26bb32fbb5f069ec3e9e77d17450022/R/FilteredData.R#L130). This R6 class object is initialized in `teal` using the helper function `teal_data_to_filtered_data, by [default this takes the datanames as a list of all the objects in the `teal_data` object.](https://github.com/insightsengineering/teal/blob/e6db5cafbab56d8ac9c4dd55b2c1200f00e7a968/R/utils.R#L77)

Example app to test:

```r
devtools::load_all("teal")

data <- teal_data() |>
  within({
    IRIS <- iris
    MTCARS <- mtcars
    `%keep_label%` <- function(lhv, rhv) {
      attributes(lhv) <- attributes(rhv)
      lhv
    }
  })

datanames(data) <- c("IRIS", "MTCARS")

init(
  data = data,
  modules <- modules(
    example_module()
  )
) |>
  runApp()
```

### sessionInfo()

_No response_

### Relevant log output

_No response_

### Code of Conduct

### Contribution Guidelines

### Security Policy

vedhav:
Closes #1366

### DIFF
--- a/R/module_nested_tabs.R
+++ b/R/module_nested_tabs.R
@@ -359,7 +359,9 @@ srv_teal_module.teal_module <- function(id,
 
 .resolve_module_datanames <- function(data, modules) {
   stopifnot("data_rv must be teal_data object." = inherits(data, "teal_data"))
-  if (is.null(modules$datanames) || identical(modules$datanames, "all")) {
+  if (length(datanames(data))) {
+    .topologically_sort_datanames(datanames(data), teal.data::join_keys(data))
+  } else if (is.null(modules$datanames) || identical(modules$datanames, "all")) {
     .topologically_sort_datanames(ls(teal.code::get_env(data)), teal.data::join_keys(data))
   } else {
     intersect(
--- a/R/utils.R
+++ b/R/utils.R
@@ -74,10 +74,13 @@ get_teal_bs_theme <- function() {
 #' @param datanames (`character`) vector of data set names to include; must be subset of `datanames(x)`
 #' @return A `FilteredData` object.
 #' @keywords internal
-teal_data_to_filtered_data <- function(x, datanames = ls(teal.code::get_env(x))) {
+teal_data_to_filtered_data <- function(x, datanames = teal.data::datanames(data)) {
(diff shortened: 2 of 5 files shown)


R/module_nested_tabs.R
```diff
@@ -359,7 +359,9 @@ srv_teal_module.teal_module <- function(id,
 
 .resolve_module_datanames <- function(data, modules) {
   stopifnot("data_rv must be teal_data object." = inherits(data, "teal_data"))
-  if (is.null(modules$datanames) || identical(modules$datanames, "all")) {
+  if (length(datanames(data))) {
```
gogonzo, line 362: I don't think this can be a first condition. If `datanames` are not null then other conditions will be ignored.

R/utils.R
```diff
@@ -74,10 +74,13 @@ get_teal_bs_theme <- function() {
 #' @param datanames (`character`) vector of data set names to include; must be subset of `datanames(x)`
 #' @return A `FilteredData` object.
 #' @keywords internal
-teal_data_to_filtered_data <- function(x, datanames = ls(teal.code::get_env(x))) {
+teal_data_to_filtered_data <- function(x, datanames = teal.data::datanames(data)) {
```
gogonzo, line 77: Maybe we can instead relax the assertion on datanames in teal.slice?

gogonzo (changes requested):
PR brings back `datanames() <-` setter which is going to be deprecated soon. This PR fixes issue by introducing other (postponed)

PR #1372, opened 2024-10-09
1362 increase minimal version of `renv`

### LINKED ISSUE #1362: Dependency Test - max failed

m7pr:
Fixes #1362 

Dependency Test in here https://github.com/insightsengineering/teal/actions/runs/11251792020

Max strategy is satisfied
image

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -50,6 +50,7 @@ jobs:
         insightsengineering/teal.logger
         insightsengineering/teal.reporter
         insightsengineering/teal.widgets
+        rstudio/renv
   r-cmd-non-cran:
     name: R CMD Check (non-CRAN) 🧬
     uses: insightsengineering/r.pkg.template/.github/workflows/build-check-install.yaml@main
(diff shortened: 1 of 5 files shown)


pawelru (changes requested):
this is good but would be tough to revert - too many places

could you please try with the Remotes instead? So vbump + Remotes. Sorry about that.

on a general note I feel that something is wrong with our CI configuration - you should have to modify 12 lines to use devel version of a package...

m7pr:
Yeah @pawelru this make sense. I just reverted `lookups_ref` and just left `Remotes` for renv.
There is so many installation strategies and installation packages that I got lost in the ocean of possibilities, even though I was deep down focused on those things recently.

pawelru (approved):

PR #1381, opened 2024-10-15
forbid "all" in teal_transform_module

### LINKED ISSUE #1347: [Question]: No warning when `module` datanames is overwritten by transformer datanames
### What is your question?

Issue after merging https://github.com/insightsengineering/teal/pull/1334

Simplification of the `module$datanames` field reduced complexity in `teal` internal code. Simplification means that if when calling `module()` function with (module) `datanames` and `transformers` (they also have datanames), all datanames are combined together so that `<teal_module>$dataname` contains `datanames` (needed for module to work) and transformer-datanames.

This is fine in most of the scenarios but problem arises when `module(datanames)` is a specified for example `data1`, `data2`. In this case when any transformer has `datanames = "all"` then `module$datanames` is changed from `"data1", "data2"` to `"all"`. This supress the warning for unsufficient datanames because if loaded data contains `"data3", "data4"`, they still satisfy "all" - so no warning presented.

### Code of Conduct

### Contribution Guidelines

### Security Policy

gogonzo:
Closes #1347 

Additionally, changed `transforms` to `transformers` in the `ui/srv_transform_data()` documentation as there was a param-name mismatch with `module()`


vignettes/data-transform-as-shiny-module.Rmd
```diff
@@ -164,76 +164,3 @@ if (interactive()) {
...
-3. Modify the UI of the module with the transform UI at the desired location by calling the `ui_transform_data`. Note that in order for the transform to work you need to change the namespace of the `id` by passing `NS(gsub("-module$", "", id), "data_transform")`.
-4. Set the `custom_ui` attribute of the `module$transformers` to `TRUE`.
-
-Now the custom module should embed the transformation UI inside the module content.
-
-Here is an example of a custom module wrapper function that modifies the `example_module` module.
-```{r}
-example_module_encoding <- function(label = "example module (on encoding)", datanames = "all", transformers = list()) {
-  mod <- example_module(label, datanames, transformers)
-  mod$ui <- function(id) {
```
gogonzo, line 185: This was super-hacky solution. Avoiding public api and overwriting module's ui.

R/teal_data_module.R
```diff
@@ -144,10 +143,17 @@ teal_data_module <- function(ui, server, label = "data module", once = TRUE) {
...
   checkmate::assert_function(ui, args = "id", nargs = 1)
   checkmate::assert_function(server, args = c("id", "data"), nargs = 2)
   checkmate::assert_string(label)
+  checkmate::assert_character(datanames)
+  if (identical(datanames, "all")) {
+    stop(
+      "teal_transform_module can't have datanames property equal to 'all'. Set `datanames = character(0)` instead.",
+      call. = FALSE
+    )
+  }
```
averissimo, line 156:
```suggestion
  stopifnot(
    "teal_transform_module can't have datanames property equal to 'all'. Set `datanames = character(0)` instead." = 
      !identical(datanames, "all")
  )
```

R/teal_data_module.R
```diff
@@ -144,10 +143,17 @@ teal_data_module <- function(ui, server, label = "data module", once = TRUE) {
 teal_transform_module <- function(ui = function(id) NULL,
                                   server = function(id, data) data,
                                   label = "transform module",
-                                  datanames = "all") {
+                                  datanames = character(0)) {
```
averissimo, line 146: [DISCUSSION] This begs the question, do we really need "all" or would `character(0)`/`NULL` serve the same purpose on `modules`?

Food for thought, not a blocker on this PR

tests/testthat/test-module_teal.R
```diff
@@ -65,55 +65,6 @@ transform_list <<- list(
   )
 )
 
-testthat::describe("srv_teal lockfile", {
```
averissimo, line 68: Why are you removing this on the PR?

If it's giving you problems locally, try installing `renv@1.0.11` that should have a fix

vignettes/data-transform-as-shiny-module.Rmd
averissimo: And it will be competely redundant with `decorator` work :+1:

averissimo (changes requested):
PR looks good, but I think you removed an unrelated test (snapshot)

Note: you are signing the commits with "Dependabot" user/email, please correct your git user for future commits.

R/teal_data_module.R
averissimo: Nevermind about this. It creates a problem with glue formatter. 

The code is very readable

R/teal_data_module.R
gogonzo: Good question. I don't want to start this discussion now. Just sharing my opinion. I think having `datanames = NULL` to hide sidebar panel is a bad api from us. I think we should not handle `module$datanames = NULL` in teal this way but rather module should have an attribute `show_sidebar = TRUE/FALSE`. 

If we change this show/hide sidebar workflow, then we would be able to properly design `datanames` to be:
- character(0) (unspecified) means all
- <specified> means <specified> 

But I don't create an issue for this. I'm not 100% sure if this is a right way to remove "all", I know for sure that NULL should be abandoned.

tests/testthat/test-module_teal.R
gogonzo: Thanks for this. I was sloppy here. Newer `renv` helps also! 👍

R/teal_data_module.R
gogonzo: Ignoring ;)

R/module_nested_tabs.R
```diff
@@ -125,8 +125,8 @@ ui_teal_module.teal_module <- function(id, modules, depth = 0L) {
             width = 3,
             ui_data_summary(ns("data_summary")),
             ui_filter_data(ns("filter_panel")),
-            if (length(modules$transformers) > 0 && !isTRUE(attr(modules$transformers, "custom_ui"))) {
-              ui_transform_data(ns("data_transform"), transforms = modules$transformers, class = "well")
+            if (length(modules$transformers) > 0) {
```
gogonzo, line 128: I think this should be in the `ui_transform_data`

R/module_transform_data.R
```diff
@@ -44,37 +45,34 @@ ui_transform_data <- function(id, transforms, class = "well") {
...
 
 #' @rdname module_transform_data
-srv_transform_data <- function(id, data, transforms, modules, is_transformer_failed = reactiveValues()) {
+srv_transform_data <- function(id, data, transformers = list(), modules, is_transformer_failed = reactiveValues()) {
   checkmate::assert_string(id)
   assert_reactive(data)
-  checkmate::assert_list(transforms, "teal_transform_module", null.ok = TRUE)
+  checkmate::assert_list(transformers, "teal_transform_module")
   checkmate::assert_class(modules, "teal_module")
-  if (length(transforms) == 0L) {
```
gogonzo, line 60: This wasn't necessary as 

```r
    Reduce(
      function(previous_result, name) {
        stop("I WOULD FAIL IF X HAS A LENGTH > 0")
      },
      x = character(0),
      init = 1
    )
 #> [1]  1
```

averissimo (approved):
Minor comments on documentation.

LGTM

R/modules.R
```diff
@@ -42,39 +42,39 @@ setOldClass("teal_modules")
...
 #' @param server_args (named `list`) with additional arguments passed on to the server function.
 #' @param ui_args (named `list`) with additional arguments passed on to the UI function.
 #' @param x (`teal_module` or `teal_modules`) Object to format/print.
 #' @param indent (`integer(1)`) Indention level; each nested element is indented one level more.
 #' @param transformers (`list` of `teal_data_module`) that will be applied to transform the data.
-#' Each transform module UI will appear in the `teal` application, unless the `custom_ui` attribute is set on the list.
-#' If so, the module developer is responsible to display the UI in the module itself. `datanames` of the `transformers`
-#' will be added to the `datanames`.
+#' Each transform module UI will appear in the `teal`'s sidebar panel.
+#' Transformers' `datanames` are added to the `datanames`. See [teal_transform_module()].
```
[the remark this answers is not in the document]
gogonzo: Suggesting is confusing. What is parent? I know that it is `module()` but we also have parent-datasets defined by join_keys etc.

R/modules.R
```diff
@@ -42,39 +42,39 @@ setOldClass("teal_modules")
 #'  - `...` (optional) When provided, `ui_args` elements will be passed to the module named argument
 #'    or to the `...`.
 #' @param filters (`character`) Deprecated. Use `datanames` instead.
-#' @param datanames (`character`) Names of the datasets that are relevant for the item.
-#' The keyword `"all"` provides all datasets available in `data` passed to `teal` application.
-#' `NULL` will hide the filter panel.
+#' @param datanames (`character`) Names of the datasets relevant to the item.
+#' - The keyword `"all"` includes all datasets available in the data passed to the teal application.
+#' - `NULL` hides the sidebar panel completely.
```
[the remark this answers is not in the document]
gogonzo: Added this information to the section `datanames` with `transformers`

R/modules.R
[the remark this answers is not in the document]
averissimo: Ok, best to avoid any further confusion

PR #1388, opened 2024-10-18
210 substitute `teal.data::get_code` with `teal.code::get_code`

m7pr:
Partner to 
- https://github.com/insightsengineering/teal.data/pull/343 
- https://github.com/insightsengineering/teal.code/pull/214

Replace `teal.data::get_code()` (that worked on `teal_data()`) with `teal.code::get_code()` (that works on `qenv`).

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -37,7 +37,7 @@ BugReports: https://github.com/insightsengineering/teal/issues
 Depends:
     R (>= 4.0),
     shiny (>= 1.8.1),
-    teal.data (> 0.6.0.9014),
+    teal.data (>= 0.6.0.9014),
     teal.slice (>= 0.5.1.9009)
 Imports:
     checkmate (>= 2.1.0),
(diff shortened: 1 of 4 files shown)


gogonzo (changes requested):
I'm sceptic about `class(data) <- "qenv"` to force usage of the `teal.code` method. What is the reason? Is it to avoid deprecation warning create `teal.data::get_code`?

m7pr:
@gogonzo `teal.code::get_code` does not have a method for `teal_data`. It has method only for `qenv`. So if you apply `teal.code::get_code` on `teal_data` then a generic is applied. Generic doesn't extract code. So you need to change the class of the obeject to `qenv`

gogonzo:
> @gogonzo teal.code::get_code does not have a method for teal_data. It has method only for qenv

Yes and not.
```r
teal_data() |> inherits("qenv")
# TRUE
```

> Generic doesn't extract code

No, it doesn't. Generic selects a method which matches the `x` class and calls this method - which technically it will pick the `teal.code::get_code()`

m7pr:
My bad! You are right. I got confused. Reverted the change.
Now it's only a change from `teal.data::get_code -> teal.code::get_code`

gogonzo (approved):
👍

gogonzo:
CI/CD pulls released versions. Probably after replacing staged.dependencies action. 

image

PR #1392, opened 2024-10-22
Use CRAN version of `renv`

m7pr:
Since `renv` was released on CRAN on Oct 12 https://cran.r-project.org/web/packages/renv/index.html with the newest fix for the bug that we reported in here https://github.com/rstudio/renv/issues/2007 we can just vbump `renv` and use CRAN version instead of GitHub version.


pawelru (approved):
LGTM 👍

PR #1393, opened 2024-10-22
Accept functions

### LINKED ISSUE #1352: [Bug]: functions in `data` raise warnings
### What happened?

When `data` contains a function, a warning is logged when running `init`:

```
[WARN] 2024-09-26 11:08:10.5918 pid:6800 token:[] teal In 'rlang::hash(list(data = data, modules = modules))': 'package:teal' may not be available when loading
```

A similar one is logged when the app is run:
```
[WARN] 2024-09-26 11:48:19.3963 pid:21388 token:[4e453302] teal In 'rlang::hash(data[[dataname]])': 'package:teal' may not be available when loading  
```

It looks like it's happening in `create_app_id`. I [modified the function](https://github.com/insightsengineering/teal/pull/1353) so that `data` is also passed through `defunction` and that removed the warning raised when running `init` but not the one raised when runninng the app.

<details> <summary> example app </summary>

```
devtools::load_all("teal")

options("teal.bs_theme" = bslib::bs_theme(version = "5"))

data <- teal_data() |>
  within(iris <- iris) |>
  within(foo <- function(x) cat("hello\n"))
modules <- modules(
  example_module()
)

app <- init(data = data, modules = modules)

runApp(app)
```

</details>

### sessionInfo()

_No response_

### Relevant log output

```R
> packageVersion("teal")
[1] '0.15.2.9064'
```

### Code of Conduct

### Contribution Guidelines

### Security Policy

gogonzo:
Closes #1352 

This PR enables including any data type in the `data` (`teal_data`) object. 
- unfilterable datasets (not data.frame nor MAE) are not included in the filter-panel, but they are preserved in the `data`
- unsupported data types are displayed in the data-summary-table but they are hidden by default
- if any unsupported dataset is in the data they data-summary displays "show/hide unsupported" to toggle rows containing unsupported
- functions are excluded from a hash calculation and this code is not included in SRC
<table>
<thead>
<th>hide unsupported</th>
<th>show unsupported</th>
</thead>
<tbody>
<tr>
<td>
image
</td>
<td>
image

</td>
</tr>
</tbody>
<table>

<details>
<summary>App example</summary>

```r
devtools::load_all("teal.slice")
devtools::load_all("teal")
options("teal.bs_theme" = bslib::bs_theme(version = "5"))

data <- teal_data() |>
  within({
    library(MultiAssayExperiment)
    data(miniACC, envir = environment())
    iris <- iris
    foo <- function(x) cat("hello\n")
    vector <- letters
  })

modules <- modules(
  example_module(
    transformers = teal_transform_module(server = function(id, data) {
        reactive({
          data() |> within({
            foo2 <- function() NULL
          })
        })
      })
    })
  ),
  example_module(datanames = "iris")
)

app <- init(data = data, modules = modules)

runApp(app)

```

</details>


pawelru:
this topic is very close to https://github.com/insightsengineering/teal.slice/issues/621 issue. Would it make sense to include it or you prefer to do this separately?

gogonzo:
> this topic is very close to [insightsengineering/teal.slice#621](https://github.com/insightsengineering/teal.slice/issues/621) issue. Would it make sense to include it or you prefer to do this separately?

Nice, I guess this solves the problem, is it?

pawelru:
> > this topic is very close to [insightsengineering/teal.slice#621](https://github.com/insightsengineering/teal.slice/issues/621) issue. Would it make sense to include it or you prefer to do this separately?
> 
> Nice, I guess this solves the problem, is it?

I haven't tried it myself but based on below it seems that yes!

> unsupported data types are displayed in the data-summary-table but they are hidden by default

R/module_data_summary.R
```diff
@@ -3,15 +3,17 @@
 #' Module and its utils to display the number of rows and subjects in the filtered and unfiltered data.
 #'
 #' @details Handling different data classes:
-#' `get_object_filter_overview()` is a pseudo S3 method which has variants for:
+#' `get_filter_overview()` is a pseudo S3 method which has variants for:
```
gogonzo, line 6: In the future it can be easily changed into S3 - so that app developers/teal-developers can make own methods to display summary for other datasets. _There is a vision for `teal.slice` that it could be extended in the future by S3 methods._

R/module_data_summary.R
```diff
@@ -95,31 +94,47 @@ srv_data_summary <- function(id, teal_data) {
...
+              } else {
+                elem
+              }
             }
           )
 
-          header_labels <- names(summary_table())
+          header_labels <- names(summary_table_out)
+          header_labels[header_labels == "dataname"] <- "Data Name"
+          substring(header_labels, 1, 1) <- toupper(substring(header_labels, 1, 1))
```
gogonzo, line 137: `get_filter_overview` returns lower-case column names now, they need to be "prettified" so that "Data name", "Subjects", "Obs" are displayed

m7pr: `toupper` changes all the letters to capital letters. maybe you need `tools::toTitleCase`

R/module_data_summary.R
```diff
@@ -163,83 +178,86 @@ get_filter_overview <- function(teal_data) {
...
+        initial_data = initial_data_objs[[dataname]],
         dataname = dataname,
         subject_keys = subject_keys
       )
     }
   )
 
-  unssuported_idx <- vapply(rows, function(x) all(is.na(x[-1])), logical(1)) # this is mainly for vectors
-  do.call(rbind, c(rows[!unssuported_idx], rows[unssuported_idx]))
+  do.call(.smart_rbind, out)
```
gogonzo, line 190: `.smart_rbind` because `get_filter_overview` output is not restricted. Previously every method have returned: 
- `dataname`, 
- `obs` and `obs_filtered`: number of filtered and unfiltered rows which were formatted into single column in `renderUI` to `Obs = <filtered>/<unfiltered>`
- `subjects` and `subjects_filtered`

Now methods can return any columns but formatting occurs in the method and not in the `renderUI`
- `dataname` 
- `obs`: formatted counts `"<filtered>/<unfiltered>"`
- `subjects` 

I've modified the way what function returns to allow easy extension in the future. `.rbind_smart` is used only to bind overviews which might be different for different object types.

R/module_data_summary.R
gogonzo: This is exactly what I need, thanks!

R/utils.R
```diff
@@ -396,3 +396,22 @@ build_datanames_error_message <- function(label = NULL,
...
   )
 }
+
+#' Smart `rbind`
+#'
+#' Combine `data.frame` objects which have different columns
+#'
+#' @param ... (`data.frame`)
+#' @keywords internal
+.smart_rbind <- function(...) {
```
averissimo, line 406: Awesome function :100:

R/module_data_summary.R
```diff
@@ -163,83 +177,86 @@ get_filter_overview <- function(teal_data) {
...
 
 #' @rdname module_data_summary
-get_object_filter_overview_MultiAssayExperiment <- function(filtered_data, # nolint: object_length, object_name.
-                                                            unfiltered_data,
-                                                            dataname) {
-  experiment_names <- names(unfiltered_data)
+get_filter_overview_MultiAssayExperiment <- function(current_data, # nolint: object_length, object_name.
+                                                     initial_data,
+                                                     dataname) {
+  experiment_names <- names(initial_data)
```
averissimo, line 243: Having `initial_data` here creates problem with MAE that's added during transforms.

```suggestion
  experiment_names <- names(current_data)
```

averissimo: Here's a test spec for this case

```r
testthat::test_that("summary table displays MAE dataset added in transforms", {
  data <- within(teal.data::teal_data(), {
    iris <- iris
    mtcars <- mtcars
    foo <- identity
  })
  shiny::testServer(
    app = srv_teal,
    args = list(
      id = "test",
      data = data,
      modules = modules(module("module_1", server = function(id, data) data, datanames = "all", transformers = list(
        teal_transform_module(
          server = function(id, data) {
            reactive({
              data() |> within({
                withr::with_package("MultiAssayExperiment", {
                  data("miniACC", package = "MultiAssayExperiment", envir = environment())
                })
              })
            })
          }
        )
      )))
    ),
    expr = {
      session$setInputs("teal_modules-active_tab" = "module_1")
      session$flushReact()
      
      testthat::expect_equal(
        module_summary_table(output, "module_1"),
        data.frame(
          "Data Name" = c(
            "foo", "iris", "miniACC", "- RNASeq2GeneNorm", "- gistict", 
            "- RPPAArray", "- Mutations", "- miRNASeqGene", "mtcars"
          ),
          Obs = c("", "150/150", "", "198", "198", "33", "97", "471", "32/32"),
          Subjects = c(NA_integer_, NA_integer_, 92, 79, 90, 46, 90, 80, NA_integer_),
          check.names = FALSE
        )
      )
    }
  )
})
```

R/module_data_summary.R
averissimo: Can we make it S3 without exporting the method?

averissimo (changes requested):
I really like the idea and the implementation makes sense, however, 🔴 the functionality is not working on my machine as the datanames are always showing (by default) and JS link does nothing

[Screencast from 2024-10-30 08-28-12.webm](https://github.com/user-attachments/assets/7e7bde32-7a6e-4576-ac75-12e247f355f3)

Can you confirm it works to see if I need to research locally why my laptop is acting up :sweat_smile: 

Note: there's a conflict with `main`, but I suppose it's a simple fix to include the `.smart_rbind` function definition

-----

I've added a few comments and found a bug in data summary (not this functionality, but I believe we can sneak the fix here). 

I have 1 suggestion about the feature:
- Group together unsupported data types on bottom

<details>

My sample app to test a bunch of unsupported types.

<summary>Example app</summary>

```r
pkgload::load_all("teal")

options(
  teal.log_level = "TRACE",
  teal.show_js_log = TRUE,
)

init(
  data = teal_data() |> within({
    my_fun <- function(x) x
    `%cbind%` <- iris
    my_expression <- expression(1+1)
    my_primitive <- c
    my_env <- new.env()
    my_env$a <- 1
    my_env$b <- 2
    my_lang <- quote(1+1)
    my_expr <- expression(1+1)
    my_quo <- rlang::quo(1+1)
    my_numeric1 <- 2

    my_order0 <- 1
    my_order1 <- iris
    my_order2 <- 1

    iris <- iris
    mae <- withr::with_package("MultiAssayExperiment", {
      data("miniACC", package = "MultiAssayExperiment", envir = environment())
      miniACC
    })
  }),
  modules = modules(
    example_module("Module (mae unsupported)", datanames = c("mae")),
    example_module("Module (small)", datanames = c("iris", "my_order0", "my_order1", "my_order2")),
    example_module("Module (all)"),
    example_module(
      "With Transformers",
      transformers = list(
        teal_transform_module(
          server = function(id, data) {
            reactive({
              data() |> within({
                t_small_iris <- head(iris)
                t_character <- "transformed"
                t_mae <- withr::with_package("MultiAssayExperiment", {
                  data("miniACC", package = "MultiAssayExperiment", envir = environment())
                  miniACC
                })
              })
            })
          }
        )
      )
    )
  )
) |> shiny::runApp()
```

</details>

R/module_data_summary.R
gogonzo: We can but I wouldn't do this now. This won't even change the code dramatically because we can't export `<method>.data.frame` or `<method>.MultiAssayExperiment` because it will complicate overriding in the future by the users. When trying to refactor `teal.slice` (you were also involved i remember), we noticed that overriding a S3 method is possible only through `.GlobalEnv` - if somebody creates `<method>.data.frame` in the package then `teal` will always prioritize `teal` method. This is why I'd avoid creating S3 for now, because the only method will be `get_filter_overview.default` with this "ugly-if--pseudo-dispatch"

R/module_data_summary.R
```diff
@@ -65,24 +71,17 @@ srv_data_summary <- function(id, teal_data) {
...
-        filter_overview$Subjects <- ifelse(
-          !is.na(filter_overview$subjects),
-          sprintf("%s/%s", filter_overview$subjects_filtered, filter_overview$subjects),
-          ""
-        )
-
-        filter_overview <- filter_overview[, colnames(filter_overview) %in% c("Data Name", "Obs", "Subjects")]
-        Filter(function(col) !all(col == ""), filter_overview)
+      is_unsupported_hidden <- reactive({
+        is_hidden <- isTRUE(input$toggle_unsupported %% 2 == 0)
```
pawelru, line 78: this is not very clear
did you consider toggle type of input instead of action link?
then the whole update of labels would not be needed

R/module_data_summary.R
```diff
@@ -95,31 +94,46 @@ srv_data_summary <- function(id, teal_data) {
         } else if (is.null(summary_table_out)) {
           "no datasets to show"
         } else {
+          any_unsupported <- any(apply(summary_table_out, 1, function(x) all(is.na(x[-1]))))
+          if (any_unsupported) {
+            shinyjs::show("toggle_unsupported")
+          } else {
+            shinyjs::hide("toggle_unsupported")
+          }
```
pawelru, line 102: How about moving this to the separate observe? 
The rationale here is to keep it responsible for a single thing - rendering the table.
On the other hand: this is not very clear that rendering the summary table is doing show/hide the link

gogonzo: Will do, thanks!

R/module_data_summary.R
gogonzo: What do you mean by type of input?

pawelru: I mean `shiny::actionLink` vs `shiny::checkboxInput` (or `shiny::input_switch`)

pawelru: Thanks for pushing this. It makes it a little bit simpler now

DESCRIPTION
pawelru, line 1: Thanks Dawid. This is very nicely implemented now and heading into being merged.

That always makes me step back and try again to look from a bigger perspective. I do understand the need of presenting the information which objects do not end up being filterable. Is this the right way to communicate this? I have some doubts. 
I'm having [this issue](https://github.com/insightsengineering/teal.slice/issues/621) in my mind which is actually coming from our examples (most likely in teal.gallery). Have a look how big is the empty space. And space on the filter panel is really precious. You have added a show/hide for this. Is this the best solution? Me, as a user, I just want to know that `a`, `b`, `c` objects are not supported. For me it's fine to put that as a table footer (or text below) and enumerate names by comma. Adding this to the table unnecessarily (!) consumes a lot (!) of space. 
I also have concerns about show/hide. This is a summary table and that suppose to be a simple information _provided_ in a simple way. To me, It feels like an unnecessary feature that distracts from the main focus (that should be putted elsewhere - filtering, encoding, main panel).

I'm sorry to put my comment that late but I want you to have a second thought about this before it got merged.

(even though it's a general comment I'm putting this a code-comment to better organise the discussion)

pawelru: Yes. That was my first thought. We just need to think about HTML formatting - proper font size. Also, what to do when the string would be super wide. Maybe horizontal scrolling?

Yet another idea (probably worse) could be to enumerate in the tooltip - something like "hoover on me to see unsupported obj names)

Speaking about tooltip - maybe an additional info why on the "?" icon.
"Unsupported: <vector>. (?)" and in the tooltip on "(?)" -> "These objects are of class that is currently not supported in the Filter Panel"

gogonzo: Me personally like the tooltip idea more. Same tooltip which is now for the unsupported in the summary table - warning sign, when moving cursor over extra info is shown

pawelru: Alright. Let's try with this then

averissimo: Do the names really matter that much to the user? If the list is long enough then it could take 2 or more lines

Alternatively, we we just have those in the tooltip

image

pawelru: That also looks good 👍

gogonzo: Unfortunately circle without filling is in font-awesome Pro. I've added following:

image

gogonzo: I'm open for suggestions here. I just wanted to highlight that unsupported could be too strong as these datasets just can't be filtered. They still can be transformed and used in the module as normal. 

----
My suggestion for future is to have these counts as S3 method which can be extended quickly by us or app developer.

pawelru: > I just wanted to highlight that unsupported could be too strong as these datasets just can't be filtered. 

I agree. Good move

From my side:
- "dataset(s)" -> "object(s)"
- icon to the right after the text
- "+ X objs" -> "And X more objs"
- smaller font / spacing?

R/module_data_summary.R
```diff
@@ -95,60 +80,74 @@ srv_data_summary <- function(id, teal_data) {
...
+          div(
+            table_html,
+            if (any(is_unsupported)) {
+              p(
+                class = c("pull-right", "float-right", "text-secondary"),
+                icon(
+                  name = "fas fa-circle-info",
+                  title = paste(
+                    sep = "",
+                    collapse = ", ",
```
vedhav, line 119:
```suggestion
                    collapse = "\n",
```

I'd prefer having a line break instead of comma separation.
Screenshot 2024-11-01 153522

R/module_data_summary.R
```diff
@@ -95,60 +80,74 @@ srv_data_summary <- function(id, teal_data) {
...
+                    shQuote(summary_table()[is_unsupported, "dataname"]),
+                    " (",
+                    vapply(summary_table()[is_unsupported, "dataname"], function(x) typeof(teal_data()[[x]]), character(1L)),
+                    ")"
+                  )
+                ),
+                sprintf("+%s unfilterable dataset(s)", sum(is_unsupported))
```
vedhav, line 126:
```suggestion
                sprintf("+%s unfilterable object(s)", sum(is_unsupported))
```

Since we include non-dataset objects here I liked the previous naming where we called it just object.

DESCRIPTION
vedhav: @pawelru does this look okay for you after this change?
Screenshot 2024-11-04 at 1 50 22 PM

pawelru: Yeah this looks great now!

R/module_data_summary.R
```diff
@@ -95,60 +80,78 @@ srv_data_summary <- function(id, teal_data) {
...
+                icon(
+                  name = "fas fa-circle-info",
+                  title = paste(
+                    sep = "",
+                    collapse = "\n",
+                    shQuote(summary_table()[is_unsupported, "dataname"]),
+                    " (",
+                    vapply(
+                      summary_table()[is_unsupported, "dataname"],
+                      function(x) typeof(teal_data()[[x]]), character(1L)
```
pawelru, line 126: why `typeof()` and not e.g. `class()[[1]]`?

e.g.:
```
r$> typeof(data.table())
[1] "list"

r$> class(data.table())
[1] "data.table" "data.frame"
```

pawelru: (... or we can collapse the vector returned by `class()`)

DESCRIPTION
pawelru: Then I'm closing that thread. Thanks guys!

R/module_data_summary.R
vedhav:
```suggestion
                      function(x) paste(class(teal_data()[[x]]), collapse = ", "), character(1L)
```

pawelru: Both are good. Given the fact that it is inside the tooltip and it's meant to be _small_ I guess we can go with just the first element

R/module_data_summary.R
```diff
@@ -95,60 +80,78 @@ srv_data_summary <- function(id, teal_data) {
...
             tags$thead(header_html),
             tags$tbody(body_html)
           )
-          table_html
+          div(
+            table_html,
+            if (any(is_unsupported)) {
+              p(
+                class = c("pull-right", "float-right", "text-secondary"),
+                style = "scale: 0.8;",
```
averissimo, line 115: Why not add this to CSS instead?

Why scale? it looks kinda weird

```suggestion
                style = "font-size: 0.8em;",
```

averissimo: Scale

image

vs. font-size

image

averissimo (approved):
It looks good! All comments are minor.

As soon as those conversations are resolved, you're ok to merge :-)

PR #1397, opened 2024-10-25
Unify `#@linksto` usage

m7pr:
This does not change any behavior of anything.
This is just consistency management with `#@linksto` tag that we have for `lockEnvironment(.raw_data)`

https://github.com/search?q=repo%3Ainsightsengineering%2Fteal+%23+%40linksto&type=code

image

### DIFF
--- a/R/module_init_data.R
+++ b/R/module_init_data.R
@@ -128,7 +128,7 @@ srv_init_data <- function(id, data) {
     function(dataname, datasets) {
       hash <- rlang::hash(data[[dataname]])
       sprintf(
-        "stopifnot(%s == %s) # @linksto %s",
+        "stopifnot(%s == %s) #@linksto %s",
         deparse1(bquote(rlang::hash(.(as.name(dataname))))),
         deparse1(hash),
         dataname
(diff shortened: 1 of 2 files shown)


averissimo (approved):

R/module_init_data.R
```diff
@@ -128,7 +128,7 @@ srv_init_data <- function(id, data) {
     function(dataname, datasets) {
       hash <- rlang::hash(data[[dataname]])
       sprintf(
-        "stopifnot(%s == %s) # @linksto %s",
+        "stopifnot(%s == %s) #@linksto %s",
```
averissimo, line 131: This is always inside strings, however the styler ruleset that we have in insightengineering enforces a space after the `#`

Would it be better to change the other instance?

m7pr: No strong opinion as it is inside a string. Without the space there is less characters :D

m7pr: I'm having second thoughts and I would actually write # comment (hasthag space somment) when I would type a comment, so let's go with your approach

averissimo: Sounds good either way.

m7pr:
Hey @averissimo I unified the comments for the `@linksto` tag the other way around!

averissimo (approved):
Awesome! :+1:

PR #1399, opened 2024-10-25
Fixes show r code reproducibility and CI

### LINKED ISSUE #1398: [Bug]: "Show R Code" doesn't create reproducible hashes for functions / environments
### Problem

Show R code is not reproducible when a function & environment is in the `qenv`.

Additionally, we've been having a reoccurring with the message: `package:teal' may not be available when loading` from `rlang::hash` 

### Cause

Objects in the `qenv` have attached environments that contain a parent environment that may not be available when `rlang::hash` is being called.

This occurs for (at least):
- Functions (non-primitive)
- Environments

### How to reproduce

This is a simplification of what is happening in `{teal}`, but it shows both problems.

``` r
library(teal.data)
#> Loading required package: teal.code
td <- teal.data::teal_data() |> within(
  fun <- function(x) x
)
first <- rlang::hash(td[["fun"]])
#> Warning in rlang::hash(td[["fun"]]): 'package:teal.data' may not be available
#> when loading
first
#> [1] "e6842bb03e26798955b9e90be0a6901c"

fun <- function(x) x
second <- rlang::hash(fun)
second
#> [1] "5454dd502596faed1fb1321a5f0f8b55"

stopifnot(identical(first, second))
#> Error: identical(first, second) is not TRUE
```

<sup>Created on 2024-10-25 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### sessionInfo()

_No response_

### Relevant log output

_No response_

### Code of Conduct

### Contribution Guidelines

### Security Policy

averissimo:
# Pull Request

Fixes #1398

#### Changes description

- [x] Fixes problem with reproducibility of functions inside `teal_data`
    - Should fix CI error in  https://github.com/r-universe/pharmaverse/actions/runs/11517110492/job/32061355783
- [x] Fixes `noSuggest` [failure in CI ](https://github.com/insightsengineering/teal/actions/runs/11423178886/job/31782067851)

### DIFF
--- a/R/module_init_data.R
+++ b/R/module_init_data.R
@@ -126,11 +126,18 @@ srv_init_data <- function(id, data) {
   vapply(
     datanames,
     function(dataname, datasets) {
-      hash <- rlang::hash(data[[dataname]])
+      x <- data[[dataname]]
+
+      code <- if (is.function(x) && !is.primitive(x)) {
+        x <- deparse1(x)
+        bquote(rlang::hash(deparse1(.(as.name(dataname)))))
+      } else {
+        bquote(rlang::hash(.(as.name(dataname))))
+      }
       sprintf(
         "stopifnot(%s == %s) # @linksto %s",
-        deparse1(bquote(rlang::hash(.(as.name(dataname))))),
-        deparse1(hash),
+        deparse1(code),
+        deparse1(rlang::hash(x)),
         dataname
       )
     },
--- a/tests/testthat/test-module_teal.R
+++ b/tests/testthat/test-module_teal.R
@@ -70,6 +70,8 @@ testthat::describe("srv_teal lockfile", {
     "creation process is invoked for teal.lockfile.mode = \"enabled\" ",
     "and snapshot is copied to teal_app.lock and removed after session ended"
   ), {
+    testthat::skip_if_not_installed("mirai")
+    testthat::skip_if_not_installed("renv")
     withr::with_options(
       list(teal.lockfile.mode = "enabled"),
       {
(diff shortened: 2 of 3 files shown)


R/module_init_data.R
```diff
@@ -126,11 +126,18 @@ srv_init_data <- function(id, data) {
   vapply(
     datanames,
     function(dataname, datasets) {
-      hash <- rlang::hash(data[[dataname]])
+      x <- data[[dataname]]
+
+      code <- if (is.function(x) && !is.primitive(x)) {
+        x <- deparse1(x)
```
m7pr, line 132: @averissimo why do you need this line? `x` is not used later on

averissimo: `x` is overwritten here (L132) and then it's reused on L140 (`deparse1(rlang::hash(x)),`)

Where the hash is being calculated for the `stopifnot` expression

averissimo:
RHub run: [Scheduled 🕰️](https://github.com/insightsengineering/teal/actions/workflows/scheduled.yaml)

(Failed on some missing libraries, but `noSuggests` is green)

averissimo:
All green now after re-running failed instances

R/module_init_data.R
m7pr: ah, yeah it is

m7pr, line 132: Ok, so for functions it `deparses` it and pastes into a string of length 1 right? Previously (without `deparse1`) it was possible that `bquote(rlang::hash(.(as.name(dataname))))` had length > 1 ?

I'm trying to understand the need to deparse1 functions

m7pr: Love it, thanks for the explanation!

m7pr (approved):
Great work @averissimo !
Thanks for fixing this.
Amazing we got rid of `can't load teal` error message from `rlang::hash`

PR #1401, opened 2024-11-04
Modify the print for modules to contain more information

### LINKED ISSUE #1378: Modify print of `modules/module` to contain more information
Following should print the modules labels and their properties:

```r
print(modules, what = c("datanames", "bookmarkable")
```

This is needed because printed `modules` gives no clue about where to set or exclude some datanames. It is hard to determine before running an app what will be displayed where

vedhav:
Closes #1378

Example code to test the format:
```r
devtools::load_all("teal")

custom_module <- function(label = "label", ui_args = NULL, server_args = NULL, datanames = "all", transformers = list(), bk = FALSE) {
  ans <- module(
    label,
    server = function(id, data, ...) {
        
      })
    },
    ui = function(id, ...) {
    },
    datanames = datanames,
    transformers = transformers,
    ui_args = ui_args,
    server_args = server_args
  )
  attr(ans, "teal_bookmarkable") <- bk
  ans
}

dummy_transformer <- teal_transform_module(
  label = "Dummy Transform",
  ui = function(id) div("(does nothing)"),
  server = function(id, data) {
    moduleServer(id, function(input, output, session) data)
  }
)

plot_transformer <- teal_transform_module(
  label = "Plot Settings",
  ui = function(id) div("(does nothing)"),
  server = function(id, data) {
    moduleServer(id, function(input, output, session) data)
  }
)

complete_modules <- modules(
  custom_module(
    label = "Data Overview",
    datanames = c("ADSL", "ADAE", "ADVS"),
    ui_args = list(
      view_type = "table",
      page_size = 10,
      filters = c("ARM", "SEX", "RACE")
    ),
    server_args = list(
      cache = TRUE,
      debounce = 1000
    ),
    transformers = list(dummy_transformer),
    bk = TRUE
  ),
  modules(
    label = "Nested 1",
    custom_module(
      label = "Interactive Plots",
      datanames = c("ADSL", "ADVS"),
      ui_args = list(
        plot_type = c("scatter", "box", "line"),
        height = 600,
        width = 800,
        color_scheme = "viridis"
      ),
      server_args = list(
        render_type = "svg",
        cache_plots = TRUE
      ),
      transformers = list(dummy_transformer, plot_transformer),
      bk = TRUE
    ),
    modules(
      label = "Nested 2",
      custom_module(
        label = "Summary Statistics",
        datanames = "ADSL",
        ui_args = list(
          stats = c("mean", "median", "sd", "range"),
          grouping = c("ARM", "SEX")
        )
      ),
      modules(
        label = "Labeled nested modules",
        custom_module(
          label = "Subgroup Analysis",
          datanames = c("ADSL", "ADAE"),
          ui_args = list(
            subgroups = c("AGE", "SEX", "RACE"),
            analysis_type = "stratified"
          ),
          bk = TRUE
        )
      ),
      modules(custom_module(label = "Subgroup Analysis in non-labled modules"))
    )
  ),
  custom_module("Non-nested module")
)

complete_modules
```

```r
> complete_modules
TEAL ROOT
  ├─ Data Overview
  │  ├─ Data Sources: ADSL, ADAE, ADVS
  │  ├─ Properties:
  │  │  ├─ Bookmarkable: Yes
  │  │  └─ Reportable: No
  │  ├─ UI Arguments: view_type, page_size, filters
  │  ├─ Server Arguments: cache, debounce
  │  └─ Transformers: Dummy Transform
  ├─ Nested 1
  │  ├─ Interactive Plots
  │  │  ├─ Data Sources: ADSL, ADVS
  │  │  ├─ Properties:
  │  │  │  ├─ Bookmarkable: Yes
  │  │  │  └─ Reportable: No
  │  │  ├─ UI Arguments: plot_type, height, width, color_scheme
  │  │  ├─ Server Arguments: render_type, cache_plots
  │  │  └─ Transformers: Dummy Transform, Plot Settings
  │  └─ Nested 2
  │     ├─ Summary Statistics
  │     │  ├─ Data Sources: ADSL
  │     │  ├─ Properties:
  │     │  │  ├─ Bookmarkable: No
  │     │  │  └─ Reportable: No
  │     │  ├─ UI Arguments: stats, grouping
  │     │  ├─ Server Arguments: ─ none
  │     │  └─ Transformers: ─ none
  │     ├─ Labeled nested modules
  │     │  └─ Subgroup Analysis
  │     │     ├─ Data Sources: ADSL, ADAE
  │     │     ├─ Properties:
  │     │     │  ├─ Bookmarkable: Yes
  │     │     │  └─ Reportable: No
  │     │     ├─ UI Arguments: subgroups, analysis_type
  │     │     ├─ Server Arguments: ─ none
  │     │     └─ Transformers: ─ none
  │     └─ root
  │        └─ Subgroup Analysis in non-labled modules
  │           ├─ Data Sources: all
  │           ├─ Properties:
  │           │  ├─ Bookmarkable: No
  │           │  └─ Reportable: No
  │           ├─ UI Arguments: ─ none
  │           ├─ Server Arguments: ─ none
  │           └─ Transformers: ─ none
  └─ Non-nested module
     ├─ Data Sources: all
     ├─ Properties:
     │  ├─ Bookmarkable: No
     │  └─ Reportable: No
     ├─ UI Arguments: ─ none
     ├─ Server Arguments: ─ none
     └─ Transformers: ─ none
```


R/modules.R
```diff
@@ -321,11 +321,124 @@ modules <- function(..., label = "root") {
 # printing methods ----
 
 #' @rdname teal_modules
+#' @param what character vector specifying which metadata to display. 
+#'   Possible values: "data_sources", "properties", "ui_args", "server_args", "transformers"
 #' @export
-format.teal_module <- function(x, indent = 0, ...) {
-  paste0(paste(rep(" ", indent), collapse = ""), "+ ", x$label, "\n", collapse = "")
+format.teal_module <- function(x, indent = 0, is_last = FALSE, parent_prefix = "", what = c("data_sources", "properties", "ui_args", "server_args", "transformers"), ...) {
```
m7pr, line 327: @vedhav are `is_last`, `parent_prefix` and `is_root` parameters documented?

DESCRIPTION
```diff
@@ -41,6 +41,7 @@ Depends:
     teal.slice (>= 0.5.1.9009)
 Imports:
     checkmate (>= 2.1.0),
+    crayon,
```
m7pr, line 44: Is this just for the colouring of the output? Do you think we could live without it? Not a big deal, but always an extra dependency

R/modules.R
```diff
@@ -321,11 +321,124 @@ modules <- function(..., label = "root") {
...
+  format_list <- function(lst, empty = empty_text) {
+    if (is.null(lst) || length(lst) == 0) {
+      empty
+    } else {
+      paste(names(lst), collapse = ", ")
+    }
+  }
+
+  bookmarkable <- ifelse(isTRUE(attr(x, "teal_bookmarkable")), "Yes", "No")
+  reportable <- ifelse("reporter" %in% names(formals(x$server)), "Yes", "No")
```
m7pr, line 342: What about `TRUE` and `FALSE` instead of `Yes` and `No`. So we stick to regular R logical values?

R/modules.R
```diff
@@ -321,11 +321,124 @@ modules <- function(..., label = "root") {
...
+      output,
+      content_prefix, "├─ ", crayon::green("Server Arguments:"), " ", format_list(x$server_args), "\n"
+    )
+  }
+  if ("transformers" %in% what) {
+    output <- paste0(
+      output,
+      content_prefix, "└─ ", crayon::magenta("Transformers:"), " ", transformers, "\n"
+    )
+  }
```
m7pr, line 383: Maybe this part can be grouped into a function call? And you could have 3 calls. `if` could be a part of the function.

R/modules.R
```diff
@@ -321,11 +321,124 @@ modules <- function(..., label = "root") {
...
+  output <- paste0(current_prefix, crayon::bold(x$label), "\n")
+  
+  if ("data_sources" %in% what) {
+    output <- paste0(
+      output,
+      content_prefix, "├─ ", crayon::blue("Data Sources:"), " ", paste(x$datanames, collapse = ", "), "\n"
+    )
+  }
+  if ("properties" %in% what) {
+    output <- paste0(
```
m7pr, line 359: Can you keep `output` as a list, so you don't need to keep pasting. And then at the end of the function, can you unlist everything and paste once?

R/modules.R
vedhav: Sure, makes sense. I got carried away with the customization :)

vedhav:
```suggestion
  bookmarkable <- isTRUE(attr(x, "teal_bookmarkable"))
  reportable <- "reporter" %in% names(formals(x$server))
```

R/modules.R
vedhav: No, Good catch! Let me add them.

DESCRIPTION
m7pr: Alrighty, just checkin

R/modules.R
```diff
@@ -321,11 +321,124 @@ modules <- function(..., label = "root") {
 # printing methods ----
 
 #' @rdname teal_modules
+#' @param what character vector specifying which metadata to display. 
+#'   Possible values: "data_sources", "properties", "ui_args", "server_args", "transformers"
 #' @export
-format.teal_module <- function(x, indent = 0, ...) {
-  paste0(paste(rep(" ", indent), collapse = ""), "+ ", x$label, "\n", collapse = "")
+format.teal_module <- function(x, indent = 0, is_last = FALSE, parent_prefix = "", what = c("data_sources", "properties", "ui_args", "server_args", "transformers"), ...) {
+  empty_text <- "─ none"
```
m7pr, line 328: Maybe emoji for `- none` or NULL? or leave empty? I bet this will probably be the most popular part of the output

R/modules.R
```diff
@@ -321,11 +321,124 @@ modules <- function(..., label = "root") {
...
+  }
+
+  bookmarkable <- isTRUE(attr(x, "teal_bookmarkable"))
+  reportable <- "reporter" %in% names(formals(x$server))
+
+  transformers <- if (length(x$transformers) > 0) {
+    paste(sapply(x$transformers, function(t) attr(t, "label")), collapse = ", ")
+  } else {
+    empty_text
+  }
```
m7pr, line 348: Any way we can go level deeper and provide deeper print for transformers? They also have datanames and ui/server args?

R/modules.R
```diff
@@ -321,11 +321,124 @@ modules <- function(..., label = "root") {
...
+
+  output <- paste0(current_prefix, crayon::bold(x$label), "\n")
+  
+  if ("data_sources" %in% what) {
+    output <- paste0(
+      output,
+      content_prefix, "├─ ", crayon::blue("Data Sources:"), " ", paste(x$datanames, collapse = ", "), "\n"
```
m7pr, line 355:
```suggestion
      content_prefix, "├─ ", crayon::blue("Datasets:"), " ", paste(x$datanames, collapse = ", "), "\n"
```

R/modules.R
```diff
@@ -321,11 +321,124 @@ modules <- function(..., label = "root") {
...
+  if ("ui_args" %in% what) {
+    output <- paste0(
+      output,
+      content_prefix, "├─ ", crayon::green("UI Arguments:"), " ", format_list(x$ui_args), "\n"
+    )
+  }
+  if ("server_args" %in% what) {
+    output <- paste0(
+      output,
+      content_prefix, "├─ ", crayon::green("Server Arguments:"), " ", format_list(x$server_args), "\n"
```
m7pr, line 375: If we can make `UI Arguments` and `Server Arguments` to have the same amount of characters, then this output would be a  bit more readable:



image

m7pr: Or maybe we can adjust `:` so they are in the same column, the same way as it is in `print.teal_slice`

image

R/modules.R
```diff
@@ -321,11 +321,124 @@ modules <- function(..., label = "root") {
...
+      output,
+      content_prefix, "├─ ", crayon::yellow("Properties:"), "\n",
+      content_prefix, "│  ├─ Bookmarkable: ", bookmarkable, "\n",
+      content_prefix, "│  └─ Reportable: ", reportable, "\n"
+    )
+  }
+  if ("ui_args" %in% what) {
+    output <- paste0(
+      output,
+      content_prefix, "├─ ", crayon::green("UI Arguments:"), " ", format_list(x$ui_args), "\n"
```
m7pr, line 369: Maybe also list the type of the argument in parenthesis? So it's `subgroups (character), analysis_type (integer)`

R/modules.R
vedhav: I think we have to implement format for `teal_transform_module` and just use it over here.
But I'm skeptical a little. Perhaps it should not be printed by default and only when explicitly asked by the user.

R/modules.R
vedhav: Good point, I think it also holds true for the sub divisions of the Properties.

R/modules.R
vedhav: I like this idea.

R/modules.R
```diff
@@ -321,11 +321,124 @@ modules <- function(..., label = "root") {
...
+#' @param what character vector specifying which metadata to display. 
+#'   Possible values: "data_sources", "properties", "ui_args", "server_args", "transformers"
 #' @export
-format.teal_module <- function(x, indent = 0, ...) {
-  paste0(paste(rep(" ", indent), collapse = ""), "+ ", x$label, "\n", collapse = "")
+format.teal_module <- function(x, indent = 0, is_last = FALSE, parent_prefix = "", what = c("data_sources", "properties", "ui_args", "server_args", "transformers"), ...) {
+  empty_text <- "─ none"
+  branch <- if (is_last) "└─" else "├─"
+  current_prefix <- paste0(parent_prefix, branch, " ")
+  content_prefix <- paste0(parent_prefix, if (is_last) "   " else "│  ")
```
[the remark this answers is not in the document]
vedhav: Yes, it does.

R/modules.R
vedhav: I did not like that idea. It will be a function that serves a very small use case and someone working on this file for the first time will have a good time understanding this over switching files and finding that function.
But, I think it's just personal preference. I'm 60-40 right now in favor of just having this if statements.

R/modules.R
```diff
@@ -321,11 +321,124 @@ modules <- function(..., label = "root") {
...
+
+  format_list <- function(lst, empty = empty_text) {
+    if (is.null(lst) || length(lst) == 0) {
+      empty
+    } else {
+      paste(names(lst), collapse = ", ")
+    }
+  }
+
```
m7pr, line 340: Hey @vedhav wanted to talk about the depth and the width. Did you consider a parameter that controls the depth of the print? Maybe let's print up to 3rd depth by default and we could trim lines at 80 or 120 or 240 digits/characters?


Depth could control if we actually print the module marked with blue, and the width could control how long is the line with the output (marked with red)?

image

R/modules.R
m7pr: I think this is how you handle this in format.teal_modules

R/modules.R
m7pr, line 327: or `data`?
```suggestion
format.teal_module <- function(x, indent = 0, is_last = FALSE, parent_prefix = "", what = c("datasets", "properties", "ui_args", "server_args", "transformers"), ...) {
```

R/modules.R
```diff
@@ -321,11 +321,124 @@ modules <- function(..., label = "root") {
...
+#' @rdname teal_modules
+#' @export
+format.teal_modules <- function(x, indent = 0, is_root = TRUE, is_last = FALSE, parent_prefix = "", ...) {
+  if (is_root) {
+    header <- paste0(crayon::bold("TEAL ROOT"), "\n")
+    new_parent_prefix <- "  " # Initial indent for root level
+  } else {
+    if (!is.null(x$label)) {
+      branch <- if (is_last) "└─" else "├─"
+      header <- paste0(parent_prefix, branch, " ", crayon::bold(x$label), "\n")
```
m7pr, line 397: you use a lot of `paste0(x, y, "\n")`. Maybe create a helper `pasten <- funciton(...) paste0(..., "\n")`

R/modules.R
m7pr: They look quite the same (if is different, colour of crayon, name used in crayon, and whats formatted) but not quite the same so maybe function will not be that conveninent/useful. Function could be in the same file, so it's not that bad. Let's revisit this place, for potential function usage if we ever get more than 3 repetitions in here

R/modules.R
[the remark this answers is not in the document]
m7pr: Hmm, I wonder if we can omit this, and in the end just globally substitute all middle `|` if they are the last `|` in the line?

R/modules.R
m7pr: I think NULL would be my top preference. But we could ask product owner for the opinion in here

R/modules.R
m7pr: Sure, I can understand some users would like to see in-depth overview and some just a glimpse. Would you be able to provide such parameter. We could name it `transformer` and have options `simple` and `full`?

R/modules.R
m7pr: Or just Data?

R/modules.R
m7pr: This is just a nice to have for the people who like to print it on a small area vs those who wants to print EVERYTHING : P

R/modules.R
```diff
@@ -334,17 +447,11 @@ print.teal_module <- function(x, ...) {
   invisible(x)
 }
 
-
 #' @rdname teal_modules
 #' @export
```
m7pr, line 451: @vedhav maybe add some examples for print/format? Take the one you proposed in the opening of this PR

R/modules.R
vedhav: Cool, going with NULL to be safe now. We can think about adding emojies later

R/modules.R
vedhav: Using this in the 3 places now.

R/modules.R
vedhav: I tried to achive this but it was non-trivial. It made the code much complex. I'll think about this some more.

R/modules.R
m7pr: What's the repetition threshold for the wrapper :D ?

R/modules.R
vedhav: When I wend with the implementation I **_feel_** that this makes it complex because the `paste0` is not avoided here.
We will have to use both the paste and appending to the list together, now we just do one.
For the modules it was different, we were fetching and storing it.

R/modules.R
vedhav: I think we can circle back to this when we work on teal transform module print method,

R/modules.R
vedhav: Done, only thing bothering me is the wrap, without that it overflows terribly.
image

R/modules.R
m7pr: It's creating a list at the begninning
`list_with_lines <- list()`
and then changing paste into
```
new_line <- pasting stuff
list_with_lines <- append(list_with_lines, new_line)
```

This way you can work on a whole output to adjust the width.
However you can now also work on the output to adjust the width. You can split by `\n` sign so the whole thing with keeping output in a list is not a big deal. 

Just trying to sync our print/format method with the way it was written for teal.slice https://github.com/insightsengineering/teal.slice/blob/main/R/teal_slice.R

R/modules.R
m7pr: Do we have an issue for a print method for teal transform module? Can you put the link to this conversation in that issue, so we don't forget, once we start writing print for teal transform module?

R/modules.R
m7pr: I think that's how we can control the width.

```r
# complete_modules is taken from your example from the opening of the discussion on this PR
cat(
  substr(capture.output(complete_modules), 1, 40),
  sep = "\n"
)
```

R/modules.R
m7pr: That's why I think we should control for the width and trim the width, so there is no overlap

m7pr: Or UI Arguments  / Server Arguments can be printed vertically instead of horizontally.

So instead of

```
  │  │  ├─ UI Arguments: plot_type, height, width, color_scheme
  │  │  ├─ Server Arguments: render_type, cache_plots
```

we get

```
  │  │  ├─ UI Arguments: 
                       - plot_type, 
                       - height, 
                       - width, 
                       - color_scheme
  │  │  ├─ Server Arguments: 
                       - render_type, 
                       - cache_plots
```

R/modules.R
vedhav: Issue created #1403

vedhav:
The new print looks like this in black console and RStudio's default console.
Screenshot 2024-11-07 141053
Screenshot 2024-11-07 141224

R/modules.R
vedhav: Thanks! This change is implemented now. Great suggestion!

R/modules.R
vedhav: Resolving this since we went with the other suggestion to keep them vertical instead.

m7pr:
@vedhav I think this looks really great

DESCRIPTION
m7pr, line 44: @vedhav do you need to add this to pre-commit yaml?

m7pr: and verdepcheck part? later on in DESCRIPTION file?

vedhav: Oh yeah, I should start using the pre-commit.

R/modules.R
```diff
@@ -321,11 +321,263 @@ modules <- function(..., label = "root") {
...
+#'   server_args = list(x = 5, y = list(p = 1))
+#' )
+#' cat(format(mod))
 #' @export
-format.teal_module <- function(x, indent = 0, ...) {
-  paste0(paste(rep(" ", indent), collapse = ""), "+ ", x$label, "\n", collapse = "")
+format.teal_module <- function(
+    x, indent = 0, is_last = FALSE, parent_prefix = "",
+    what = c("datasets", "properties", "ui_args", "server_args", "transformers"), ...) {
+  empty_text <- "NULL"
```
m7pr, line 347: Hey @vedhav I thought about regular NULL (not a character) and empty spaces in the final print. Now we get a NULL sign.

DESCRIPTION
vedhav: Yup my commits are 🐢now...
How long does it take, it's been over 3 mins for me now.

R/modules.R
vedhav: I don't get what you mean. Can you please explain what you mean by `NULL` and empty spaces?

DESCRIPTION
vedhav: Did not time it but it took over 4 mins for sure. Resolving it now.

m7pr: It takes more time during first build becuase it creates cache. If you have global R RENV CACHE it reuses your cache

R/modules.R
m7pr: Instead of 

```r
empty_text <- "NULL"
```

try 

```r
empty_text <- NULL
```

m7pr: And those are empty spaces that are in the output 
image

vedhav: For me the empty spaces only come when I have `NULL` and not `"NULL"`

m7pr (approved):
Left one last commit, but this is already in a good shape. Really nice job!

R/modules.R
m7pr: Yes, that was my idea. Let's change `"NULL"` to `""`

vedhav:
The new print method for the teal modules look like this in black console and RStudio's white console.

Screenshot 2024-11-07 150908
Screenshot 2024-11-07 150848

PR #1402, opened 2024-11-04
`teal.data::datanames()` is deprecated in favor of dot-prefix and `ls()`

averissimo:
# Pull Request

Part of https://github.com/insightsengineering/teal.data/issues/333

Blocked by:

- https://github.com/insightsengineering/teal.code/pull/218
- https://github.com/insightsengineering/teal.data/pull/347


R/module_nested_tabs.R
```diff
@@ -364,11 +364,11 @@ srv_teal_module.teal_module <- function(id,
 .resolve_module_datanames <- function(data, modules) {
   stopifnot("data_rv must be teal_data object." = inherits(data, "teal_data"))
   if (is.null(modules$datanames) || identical(modules$datanames, "all")) {
-    .topologically_sort_datanames(ls(teal.code::get_env(data)), teal.data::join_keys(data))
+    names(data) # TODO: Reviewer: Check that previous code does nothing.
```
averissimo, line 367: Please confirm that the previous statement did nothing

```suggestion
    names(data)
```

R/module_nested_tabs.R
```diff
@@ -364,11 +364,11 @@ srv_teal_module.teal_module <- function(id,
...
   } else {
     intersect(
-      .include_parent_datanames(modules$datanames, teal.data::join_keys(data)),
-      ls(teal.code::get_env(data))
+      names(data), # Keep topological order from teal.data::names()
+      .include_parent_datanames(modules$datanames, teal.data::join_keys(data))
```
gogonzo, line 371: This would solve the problem of `datanames = ".iris"`, but even if we included dot-prefixed-dataset in module's `data` it will be hidden when using `names` in the module. I think your proposition is relevant and consistent - we ignore all dot-prefixed-datasets and that's it, period.

```r
    intersect(
      .include_parent_datanames(modules$datanames, teal.data::join_keys(data)),
      ls(data)
    )
```

R/utils.R
```diff
@@ -46,43 +46,32 @@ get_teal_bs_theme <- function() {
...
   ordered_datanames <- datanames
-  for (i in datanames) {
-    parents <- character(0)
-    while (length(i) > 0) {
-      parent_i <- teal.data::parent(join_keys, i)
-      parents <- c(parent_i, parents)
-      i <- parent_i
+  for (current in datanames) {
+    parents <- character(0L)
+    while (length(current) > 0 && !current %in% parents[-1]) {
```
gogonzo, line 51: This looks like a bug. If dataset can have a single parent then neither ancestor can have two descendants if they are in a line. It means that either `&&` would throw because `length(current) > 1`  or condition will never be TRUE. Am I right?
 
```suggestion
    while (length(current) > 0) {
```

averissimo: The second part ensures it cannot go into infinite recursion. The `join_keys` data structure prevents that, but I think we should ensure that is never an option.

R/module_nested_tabs.R
averissimo: So do you want to keep `names()` for now and rely on the topological order determined in `teal.data::names.teal_data` or use `ls()` here and order it with `.include_parent_....`?

R/utils.R
averissimo: Good point, no need to double check this here

gogonzo (dismissed):
👍

gogonzo (approved):

PR #1406, opened 2024-11-11
Remove TealData checks

### LINKED ISSUE #1386: Review Deprecated Functions
### Summary

Let's review the functions we deprecated in the previous release and decide whether to hard deprecate them (`lifecycle::deprecate_stop`) or remove them from the codebase entirely.

| Function | Action |
| --- | --- |
| hard deprecate on [TealData](https://github.com/insightsengineering/teal/blob/9c183f8c474698e20ec5842299cc8e189c28e7ee/R/init.R#L106-L115) | Remove |
| datasets argument in [ui](https://github.com/insightsengineering/teal/blob/9c183f8c474698e20ec5842299cc8e189c28e7ee/R/modules.R#L213-L220) and [srv](https://github.com/insightsengineering/teal/blob/9c183f8c474698e20ec5842299cc8e189c28e7ee/R/modules.R#L192-L199) formals in `module()`| Remove |

llrs-roche:
# Pull Request

Fixes #1386


llrs-roche:

NEWS.md
```diff
@@ -15,6 +15,7 @@
 `datanames` one should modify `module$datanames`.
 * The `landing_popup_module()` needs to be passed as the `landing_popup` argument of `init` instead of being passed as a module of the `modules` argument of `init`.
 * `teal` no longer re-export `%>%`. Please load `library(magrittr)` instead or use `|>` from `base`.
+* `module(server)` argument does not accept the `datasets` argument.
```
m7pr, line 18:
```suggestion
* `module(server)` parameter does not accept the `datasets` argument.
```

m7pr: However, above we use `argument`

> * The `landing_popup_module()` needs to be passed as the `landing_popup` argument of `init` instead of being passed as a module of the `modules` argument of `init`

m7pr: Not a crucial comment

R/modules.R
```diff
@@ -197,7 +197,7 @@ module <- function(label = "module",
     )
   }
   if ("datasets" %in% server_formals) {
-    warning(
+    stop(
       sprintf("Called from module(label = \"%s\", ...)\n  ", label),
```
m7pr, line 201: I think we should just remove.
It's been a while. And remove above in the `stop` condition

m7pr:
@llrs-roche left you 2 comments.
Also R CMD CHECK is failing so that was not ready for the review yet

NEWS.md
```diff
@@ -15,6 +15,7 @@
 `datanames` one should modify `module$datanames`.
 * The `landing_popup_module()` needs to be passed as the `landing_popup` argument of `init` instead of being passed as a module of the `modules` argument of `init`.
 * `teal` no longer re-export `%>%`. Please load `library(magrittr)` instead or use `|>` from `base`.
+* `module(server)` parameter does not accept the `datasets` argument.
```
m7pr, line 18:
```suggestion
* `module(server)` parameter no longer supports the `datasets` argument.
```

gogonzo:
I wouldn't remove datasets for some time. I know there are some internal applications which works on our latest release but still using `datasets`. Removing `datasets` completely  should be followed with some help for those users (snowflake people)

R/modules.R
gogonzo: If we remove it we need to provide a support to people using this feature. I know at least one team which uses `datasets` in their modules.

gogonzo: I suggest to make an issue to provide a support (when removed)

m7pr: Let's maybe keep it as `stop` and point to an issue that provides the support?
And let's create the instructions on this issue?

llrs-roche: I think raising the bar from `warning` to `stop` will be enough. The instructions on the warning were long enough to be helpful. I don't think it needs a new issue to explain this.

m7pr: @llrs-roche can you bring back the error-message and use stop, instead of warning?

R/modules.R
```diff
@@ -217,7 +217,8 @@ module <- function(label = "module",
       "\n - `...` ui_args elements will be passed to the module argument of the same name or to the `...`"
     )
   }
-  if (any(c("data", "datasets") %in% ui_formals)) {
+
+  if (any("datasets" %in% ui_formals)) {
```
gogonzo, line 221: Please revert. We don't support any data in the ui functions - data related content can be created only via server functions. Thanks to this UI "skeleton" can be rendered on init without a need to wait on data.


```suggestion
  if (any(c("data", "datasets") %in% ui_formals)) {
```

R/modules.R
```diff
@@ -196,16 +196,16 @@ module <- function(label = "module",
       "\n - `...` server_args elements will be passed to the module named argument or to the `...`"
     )
   }
+
   if ("datasets" %in% server_formals) {
-    warning(
+    stop(
```
gogonzo, line 201: I think this needs to be discussed. I know it is one of the acceptance criteria but I'm not sure if we really want this now.

donyunardi: Thanks for raising this.

I'm not too keen on adding extra work for us if we end up needing to support our internal end users due to this change.

Let’s hold off on hard-deprecating `datasets` in `module(server)` for now and proceed with the rest.

@llrs-roche @m7pr any concern?

llrs-roche: Ok, I'll revert back to a warning. 

Perhaps as part of the https://github.com/insightsengineering/nestdevs-tasks/issues/86 we could  decide a policy on the life cycle of functions and arguments? It could incorporate how we search and find internal (looking at internal Roche's code repositories?, asking someone?)  and external repositories (searching github?, asking other programmers from other companies?) that could be affected. Similar to how the tidyverse sends PR to affected packages before sending their new version to CRAN.

llrs-roche:
To summarize the changes after the discussion: I only removed TealData checks on module.
There is the tdata.R file with hard deprecated related functions that I haven't removed.

modules ui and server formals: Not modified, waiting for guidelines to decide how to proceed.

gogonzo (approved):
👍

PR #1414, opened 2024-11-14
Update teal.slice version

m7pr:
Once this was merged in teal.slice and the version was bumped to `0.5.1.9013` https://github.com/insightsengineering/teal.slice/pull/622

We now allow non-alphabetical non-standard names in filter panel.
Without it, some tests break. Below is for teal.slice `0.5.1.9012` 

```r
data = within(
  teal.data::teal_data(),
  {
    iris <- iris
    mtcars <- mtcars
    `_a variable with spaces_` <- "new_column" # nolint: object_name.
    iris <- cbind(iris, data.frame(`_a variable with spaces_`))
  }
)
teal_data_to_filtered_data(data)

Error in check_simple_name(dataname) : 
  name '_a variable with spaces_' must only contain alphanumeric characters (with underscores) and the first character must be an alphabetic character
Called from: check_simple_name(dataname)
```


averissimo (approved):
All good!

Issue #1409, opened 2024-11-13
[Question]: What do display when missing datasets passed to teal


gogonzo:
### What is your question?

When missing data is passed to the application we need to show the warning somewhere. When data is loaded from `teal_data_module`, then the warnings about which-dataset-is-missing-for-which-module is displayed in "Data" tab. When one passes a data eagerly, then warning is displayed over tabs. For teal applications wrapped within other applications data can be passed as reactive. In this case "Data" tab is also ommited (as there is no teal_data_module), but otherwise warning is displayed over the tabs (as it needs to be displayed somewhere). 

| Static or wrapped   | Dynamic |
| -------- | ------- |
| <img width="1217" alt="image" src="https://github.com/user-attachments/assets/423f0721-cb1a-4b3c-958f-595efd5ef6e1">  | <img width="1458" alt="image" src="https://github.com/user-attachments/assets/24febc4e-8b8e-418b-a02d-27eef8fbf2a6">   |

<details>
<summary> app with datanames (static) </summary>

```r
options(
  teal.log_level = "TRACE",
  teal.show_js_log = TRUE,
  # teal.bs_theme = bslib::bs_theme(version = 5),
  shiny.bookmarkStore = "server"
)

library(bslib)
library(teal)

trans <- list(
  anl_w_datanames = teal_transform_module(
    label = "ANL with datanames",
    datanames = "ADTTE",
    ui = function(id) {
      ns <- NS(id)
      tagList(
        div("This transformer adds ANL based on specified ADSL, ADTTE"),
        numericInput(ns("obs"), "Number of subjects", value = 400, min = 0, max = 400)
      )
    },
    server = function(id, data) {
      moduleServer(id, function(input, output, session) {
        reactive({
          within(data(),
            {
              ANL <- dplyr::inner_join(
                head(ADSL, nobs),
                ADTTE[c("STUDYID", "USUBJID", setdiff(colnames(ADTTE), colnames(ADSL)))],
                by = c("USUBJID", "STUDYID")
              )
            },
            nobs = input$obs
          )
        })
      })
    }
  ),
  anl_wout_datanames = teal_transform_module(
    label = "ANL without datanames",
    ui = function(id) {
      ns <- NS(id)
      tagList(
        div("This transformer adds ANL based on unspecified ADSL, ADTTE"),
        numericInput(ns("obs"), "Number of subjects", value = 400, min = 0, max = 400)
      )
    },
    server = function(id, data) {
[...]

PR #1416, opened 2024-11-15
Adds warning on reserved datanames ("all")

### LINKED ISSUE #1379: Warn when data contains an object named "all"
Because "all" is a reserved word in `module()`, otherwise this would be ambiguous.

averissimo:
# Pull Request

Fixes #1379

### Changes description:

- [x] Add warning on console/UI when checking for datanames
- [x] Add `pluralize()` function that allows a simplification of code (removing a bunch of `if` / `ifelse` statements)
- Adds tests for new functionality
  - [x] in UI
  - [x] in Console

### Special ask for reviewer

- Consider if the message could be improved :-)


R/utils.R
```diff
@@ -134,14 +134,40 @@ check_modules_datanames <- function(modules, datanames) {
...
+  reserved_datanames <- check_reserved_datanames(datanames)
+
   if (!length(check_datanames)) {
-    return(TRUE)
+    out <- if (is.null(reserved_datanames)) {
+      TRUE
+    } else {
+      reserved_datanames
+    }
+    return(out)
```
llrs-roche, line 167: I'm not too familiar with this code, but returning early omits the checks on missing_datanames below

R/utils.R
```diff
@@ -445,3 +471,26 @@ build_datanames_error_message <- function(label = NULL,
...
+#' Pluralize a word depending on the size of the input
+#'
+#' @param x (`object`) to check length for plural.
+#' @param singular (`character`) singular form of the word.
+#' @param plural (optional `character`) plural form of the word. If not given an "s"
+#' is added to the singular form.
+#'
+#' @return A `character` that correctly represents the size of the `x` argument.
+#' @keywords internal
+pluralize <- function(x, singular, plural = NULL) {
```
llrs-roche, line 484: This new function really needed? Are there other cases were it will be used? If not, it will be simpler to just check on line 144 https://github.com/insightsengineering/teal/pull/1416/files#diff-60bef5af43a1bd541299ce9bceeb226afb23deaa5e637b145c85e05ffd552b0cR144 and chose one or the other sentence.

llrs-roche (commented):
When I try with the example code at the issue about [.raw_data](https://github.com/insightsengineering/teal/issues/1379#issuecomment-2422602723) it doesn't show a warning for .raw_data and I don't see a test for this case.

I assume it was added to check_reserved_datanames because it is what is is for, but when I run the testing apps, I see on my terminal:

```
[WARN] 2024-11-15 13:09:16.0411 pid:9640 token:[] teal <span>
  <code>all</code>
  is reserved for internal use. Please avoid using it as a dataset name.
</span>
```
Is this expected? I thought it would only show up on the browser?

R/utils.R
averissimo: This returns early when there are no messages on missing datanames.

The actual check happens earlier

R/utils.R
averissimo: I believe this will simplify the whole file, as we have a bunch of `if` and `ifelse` statements

I didn't commit yet to make sure everything works as expected.

R/utils.R
llrs-roche: Sorry, I checked more carefully the code and now I think I understand it. Although I find it more readable to use && instead of assignment on if clauses that have else statements, or simply using several times return inside the conditions.

averissimo:
@llrs-roche the `.raw_data` reserved namespace cannot be tested right now as we currently don't support dot-prefixed as explicit datanames.

Thanks for catching the console issue :+1:

R/utils.R
averissimo: I think here is a matter of style. I don't like any of the possible styles so I ended up with a single return.

I'm flexible on this, is the suggestion below ok, or would you prefer an alternative implementation?

### 1. `if` ... `else` ...

```r
  if (!length(check_datanames) && is.null(reserved_datanames)) {
    return(TRUE)
  } else if (!length(check_datanames)) {
    return(shiny::tagList(reserved_datanames))
  }
```

<details>

<summary>Other options</summary>

### 2. 2 separate early returns

```r
  if (!length(check_datanames) && is.null(reserved_datanames)) {
    return(TRUE)
  } 
  if (!length(check_datanames)) {
    return(shiny::tagList(reserved_datanames))
  }
```

### 3. As is but double return

```r
  if (!length(check_datanames)) {
    if (is.null(reserved_datanames)) {
      return(TRUE)
    } else {
      return(shiny::tagList(reserved_datanames))
    }
  }
```

### 4. As is

```r
  if (!length(check_datanames)) {
    out <- if (is.null(reserved_datanames)) {
      TRUE
    } else {
      shiny::tagList(reserved_datanames)
    }
    return(out)
  }
```

As sidenote my favourite in terms of readability would be: (if we were using `%||%`)


```r
  if (!length(check_datanames)) {
    reserved_datanames %||% return(TRUE)
    return(shiny::tagList(reserved_datanames))
  }
```

</details>

llrs-roche:
I can confirm that the messages on the R console disappeared. Thanks also for simplifying some tests. 

I realized that there are two messages on the app about the same issue, see image below. Could this be simplified so that it only appears once or is it by design?

warning_restriced_words_teal

----

Locally I see a new error related to renv, I haven't seen modified anything related to this, so not sure why now it is failing now:

```r
Failure (test-module_teal.R:92:13): creation process is invoked for teal.lockfile.mode = "enabled" and snapshot is copied to teal_app.lock and removed after session ended

`actual`:   FALSE
`expected`: TRUE 
```

Let's see what happens on the GHA, if it pass I'll approve the PR to be merged

averissimo:
Good eye @llrs-roche I noticed the same oddity

We need to confirm if that is a bug added in the last few weeks or a feature that groups all errors together at the top.

Discussion moved to: https://github.com/insightsengineering/teal/issues/1409

llrs-roche (approved):
LGTM

PR #1357, opened 2024-09-30
Adds decorate functionality to module output

gogonzo:
closes #1383 #1384 

Companion PRs:

- https://github.com/insightsengineering/teal.modules.general/pull/795

<details>
<summary>example tmg app</summary>

```r
pkgload::load_all("teal")
pkgload::load_all("teal.modules.general")
library(teal.widgets)

data <- teal_data()
data <- within(data, {
  require(nestcolor)
  ADSL <- rADSL
})
join_keys(data) <- default_cdisc_join_keys[c("ADSL")]

footnote_regression <- teal_transform_module(
  server = make_teal_transform_server(expression(
    plot <- plot + labs(caption = deparse(summary(fit)[[1]]))
  ))
)

fact_vars_adsl <- names(Filter(isTRUE, sapply(data[["ADSL"]], is.factor)))
vars <- choices_selected(variable_choices(data[["ADSL"]], fact_vars_adsl))

app <- init(
  data = data,
  modules = modules(
    tm_a_regression(
      label = "Regression",
      response = data_extract_spec(
        dataname = "ADSL",
        select = select_spec(
          label = "Select variable:",
          choices = "BMRKR1",
          selected = "BMRKR1",
          multiple = FALSE,
          fixed = TRUE
        )
      ),
      regressor = data_extract_spec(
        dataname = "ADSL",
        select = select_spec(
          label = "Select variables:",
          choices = variable_choices(data[["ADSL"]], c("AGE", "SEX", "RACE")),
          selected = "AGE",
          multiple = TRUE,
          fixed = FALSE
        )
      ),
      ggplot2_args = ggplot2_args(
        labs = list(subtitle = "Plot generated by Regression Module")
      ),
      decorators = list(footnote_regression)
    )
  )
)

shinyApp(app$ui, app$server)

```

</details>

PR #1423, opened 2024-11-22
Fix examples on decorator vignette

llrs-roche:
# Pull Request

Fixes comments on vignette example on #1357


llrs-roche:
Thank Marcin, but this doesn't fix all examples, specially the interactive-from lang issue as shown on the video from https://github.com/insightsengineering/teal/pull/1357#discussion_r1853745262 is not fixed. At least on my computer.

My current understanding is that at https://github.com/insightsengineering/teal/blob/1187_decorate_output%40main/R/module_transform_data.R#L96 the teal_data has already the new x_axis_title (on third or fourth pass), but how it got there and how to trigger it again when I change the input still eludes me.

llrs-roche:
Digging a b it deeper, I think that [`make_teal_transform_server()` ](https://github.com/insightsengineering/teal/blob/347b82a5246b1d2cc31b453de58c38cd8328b042/R/teal_transform_module.R#L193) should have some kind of observe or binding to the input to ensure that when the input is updated the decorator is executed again.  Otherwise it seems that it is only executed once

PR #1425, opened 2024-12-09
Prevent too early evaluation

gogonzo:
When `server_args` contains element being a `language` then it is being evaluated by `do.call`. We call `do.call` when calling modules in teal. Adding `quote = TRUE` to the call fixes the problem. See an example app:

```r
tm_query <- function(query) {
  module(
    ui = function(id) {
      ns <- NS(id)
      verbatimTextOutput(ns("out"))
    },
    server = function(id, data, query) {
        output$out <- renderPrint({
          q <- eval_code(data(), query)
          print(q)
        })
      })
    },
    server_args = list(query = query)
  )
}


library(teal.code)
library(teal.data)
pkgload::load_all("teal")

app <- init(
  data = teal_data() |> within(a <- iris),
  modules = modules(
    tm_query(query = quote(a_filtered <- subset(a, a$Species == "setosa")))
  )
)

runApp(app)

```

Alternative is to ignore this error and replace `quote()` with `expression()` as `expression` is not evaluated in `do.call`.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,8 +1,8 @@
 Type: Package
 Package: teal
 Title: Exploratory Web Apps for Analyzing Clinical Trials Data
-Version: 0.15.2.9098
-Date: 2025-01-08
+Version: 0.15.2.9100
+Date: 2025-01-15
 Authors@R: c(
     person("Dawid", "Kaledkowski", , "dawid.kaledkowski@roche.com", role = c("aut", "cre"),
            comment = c(ORCID = "0000-0001-9533-457X")),
@@ -50,7 +50,7 @@ Imports:
     shinyjs,
     stats,
     teal.code (>= 0.5.0.9015),
-    teal.logger (>= 0.2.0),
+    teal.logger (>= 0.3.0.9003),
     teal.reporter (>= 0.3.1.9004),
     teal.widgets (>= 0.4.0),
     tools,
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,4 @@
-# teal 0.15.2.9098
+# teal 0.15.2.9100
 
 ### New features
 
--- a/R/modules.R
+++ b/R/modules.R
@@ -73,8 +73,8 @@ setOldClass("teal_modules")
 #' @param server_args (named `list`) with additional arguments passed on to the server function.
 #' @param ui_args (named `list`) with additional arguments passed on to the UI function.
 #' @param x (`teal_module` or `teal_modules`) Object to format/print.
-#' @param transformators (`list` of `teal_transform_module`) that will be applied to transformator module's data input.
-#'
+#' @param transformators (`list` of `teal_transform_module`) that will be applied to transform module's data input.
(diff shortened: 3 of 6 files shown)


llrs-roche:
I'm not that familiar with expressions, calls and languages, this seems a simple fix to the issue. But I leave the review for other that might understand better the implications of this change.

I noticed that `callModule` is "soft-deprecated":

>Note: As of Shiny 1.5.0, we recommend using moduleServer() instead of callModule(), because the syntax is a little easier to understand, and modules created with moduleServer can be tested with testServer().

I don't think this would help on line 344, but perhaps is something to keep in mind.

R/module_nested_tabs.R
```diff
@@ -339,9 +339,9 @@ srv_teal_module.teal_module <- function(id,
   }
 
   if (is_arg_used(modules$server, "id")) {
-    do.call(modules$server, args)
+    do.call(modules$server, args, quote = TRUE)
```
averissimo, line 342: Consider adding a test before merging, such as:

```r
testthat::it("srv_teal_module.teal_module does not pass data if not in the args explicitly", {
  tm_query <- function(query) {
    module(
      "module_1", 
      server = function(id, data, query) {
          reactive(q <- eval_code(data(), query))
        )
      },
      server_args = list(query = query)
    )
  }
  shiny::testServer(
    app = srv_teal,
    args = list(
      id = "test",
      data = teal.data::teal_data(a_dataset = iris),
      modules = modules(tm_query(quote(a_dataset <- subset(a_dataset, Species == "setosa"))))
    ),
    expr = {
      session$setInputs(`teal_modules-active_tab` = "module_1")
      session$flushReact()
      
      testthat::expect_setequal(
        "setosa",
        unique(modules_output$module_1()()[["a_dataset"]]$Species)
      )
    }
  )
})
```

averissimo (commented):
I think this responsibility is on the module developer. See example below that works in `main` with only 1 line changing (`query <- enquote(query)`

However, I don't see any bad consequences on this change, the PR is fine and I tested with different types of native language objects.

I left a comment about adding a test for this so we don't risk a regression in the future.

```r
pkgload::load_all("teal", export_all = FALSE)
tm_query <- function(query) {
  query <- enquote(query)
  module(
    ui = function(id) {
      ns <- NS(id)
      shiny::tagList(
        verbatimTextOutput(ns("out")),
        verbatimTextOutput(ns("out2"))
      )
    },
    server = function(id, data, query, yada) {
        output$out <- renderPrint({
          q <- eval_code(data(), query)
          print(q)
        })
      })
    },
    server_args = list(query = query)
  )
}


library(teal.code)
library(teal.data)

app <- init(
  data = teal_data() |> within(a <- iris),
  modules = modules(
    tm_query(query = quote(a_filtered <- subset(a, a$Species == "setosa")))
  )
)

runApp(app)
```

llrs-roche (approved):
The new test passes on this branch and not on main, and this PR also fixes the issue on https://github.com/insightsengineering/teal.modules.general/pull/824#issuecomment-2590020038.
I updated the branch to check again and be able to merge it soon

PR #1430, opened 2024-12-13
Rename  teal_data_module.R functions

### LINKED ISSUE #1323: Request for renames in `teal_data_module`
Follow-up on https://github.com/insightsengineering/teal/pull/1253

Please find below a list of objects which names I personally found confusing. If you agree - let's try to find a good name.

* `srv_teal_data()` & `ui_teal_data()`
  This is a module to resolve `teal_data_module` class of objects, in particular: not a `teal_data` class of objects
* inside `srv_teal_data()`: it's easy to confuse `data`, `data_module` and `data_out` (for this: there is `srv_validate_reactive_teal_data(..., data = data_out, ...)` which is even more confusing)
  It's fully functional but it's very easy to miss what object means what and you have to put a lot of extra focus to correctly follow the logic. I propose the following (open for discussion):
  * rename `data` to `r_teal_data` (reactive containing `teal_data`)
  * rename `data_module` to `teal_data_module`
  * rename `data_out` into `module_out` (to keep it separate from `data` but close to `teal_data_module`)
 * inside `srv_validate_reactive_teal_data()`: rename `data` argument into `r_teal_data` (reactive containing `teal_data`)
  * inside `srv_validate_reactive_teal_data()`: rename `data_out_r` variable into `r_try_teal_data` (reactive containing try-catch of `r_teal_data` evaluation (see above))
* inside `.fallback_on_failure()`:
  * rename `this` argument into `x`
  * rename `that` argument into `y`
  `this` and `that` are just bad arg names. This feels very similar to `%>%` or `rlang::%||%` inflix functions (with exception of `label` arg) and these usually have `x` and `y`

llrs-roche:
# Pull Request

* `srv_teal_data()` & `ui_teal_data()`:  renamed to `srv_teal_data_module()` & `ui_teal_data_module()`. 
* inside `srv_teal_data_module()`:  
  * renamed `data` to `teal_data_r` (`teal_data` reactive ): This suffix is what is used inside tmc and tmg modules.
  * rename `data_module` to `teal_data_module`. 
  * rename `data_out` into `module_out` (to keep it separate from `data` but close to `teal_data_module`)
 * inside `srv_validate_reactive_teal_data()`: rename `data` argument into `teal_data_r` to match previous renaming and schema in the framework. 
  * inside `srv_validate_reactive_teal_data()`: The tryCatch code moved to `srv_teal_data_module` and as the code checked is `module_out` I renamed the output to `try_module_out`. 
* inside `.fallback_on_failure()`: There is no longer `.fallback_on_failure()`. 

Fixes #1323 

Due to the renaming I had to change some tests. 

I checked interactively the examples with `devtools::run_examples()` as well as with R CMD check. I'm not sure whether this PR broke the examples on ?`teal::teal_slices`. This is an internal function with some examples that only display the ID, and report there is no matching id.


R/module_teal_data.R
```diff
@@ -89,14 +97,16 @@ srv_teal_data <- function(id,
...
-      data = try_module_out,
+      teal_data_r = try_module_out,
       modules = modules,
       validate_shiny_silent_error = validate_shiny_silent_error,
       hide_validation_error = is_previous_failed
     )
   })
 }
 
+srv_teal_data <- srv_teal_data_module
```
m7pr, line 108: so inside code this is still `srv_teal_data`?

R/module_teal_data.R
```diff
@@ -48,12 +51,17 @@ ui_teal_data <- function(id, data_module = function(id) NULL) {
   )
 }
 
+ui_teal_data <- ui_teal_data_module
```
m7pr, line 54: so inside code this is still `ui_teal_data`?

llrs-roche:
Renaming `srv_validate_reactive_teal_data` data to something else have some code implications, as even if it is an internal function the documentation is then reused on some exported function (`srv_transform_teal_data`). 

I can rename the external function parameter to match the new one but this will require renaming multiple packages and modules (notably the new `srv_decorate_teal_data` function introduced on teal.modules.general (tmg) and  teal.modules.clinical (tmc)). I am not sure how many packages could be affected by this renaming.

I could also duplicate the documentation on the internal function `srv_validate_reactive_teal_data` with the renamed parameter and keep using the old parameter on `srv_transform_teal_data` with the documentation for that parameter copied on its documentation file. 

Renaming {ui,srv}_teal_data to {ui,srv}_teal_data_module was made via an alias, so I don't expect any function to break due to this change. But it might be worth to check how this renaming would affect teal 

Pushing to see checks results but I expect them to fail, as the path I took was to rename the parameter name but haven't changed anything related to it on tests (or dependencies).

R/module_teal_data.R
llrs-roche: Yes, I made this decision to make it easier to keep the code as is before evaluating how much code should be changed if we replaced srv_teal_data by srv_teal_data_module

R/module_teal_data.R
llrs-roche: Yes, I made this decision to make it easier to keep the code as is before evaluating how much code should be changed if we replaced ui_teal_data by ui_teal_data_module. I'm checking the code locally and I'm seeing lots of tests failing and I don't want to make it even harder to fix everything (yet).

R/module_teal_data.R
```diff
@@ -37,7 +37,10 @@
 NULL
 
 #' @rdname module_teal_data
-ui_teal_data <- function(id, data_module = function(id) NULL) {
+#' @aliases ui_teal_data
+#' @note
+#' `ui_teal_data_module` was renamed from `ui_teal_data`.
+ui_teal_data_module <- function(id, data_module = function(id) NULL) {
```
m7pr, line 43: Is `ui_teal_data` used in other teal packages? Or is it internal `teal` function?

llrs-roche: I searched via `grep` ui_teal_data and srv_teal_data and I didn't found any usage in other packages (and just a single call to srv_teal_data in teal).  They are marked as internal too on R/module_teal_data.R L36. 

I didn't modify the exported/imported status. Yes, it is internal and not exported anymore.

m7pr:
@llrs-roche 

> I could also duplicate the documentation on the internal function srv_validate_reactive_teal_data with the renamed parameter and keep using the old parameter on srv_transform_teal_data with the documentation for that parameter copied on its documentation file.

Please do!

R/dummy_functions.R
```diff
@@ -62,7 +62,7 @@ example_module <- function(label = "example teal module",
 
         table_data_decorated_no_print <- srv_transform_teal_data(
           "decorate",
-          data = table_data,
+          teal_data_r = table_data,
```
m7pr, line 65: I'm torn between `r_teal_data` and `teal_data_r`. Ask yourself how would you read in your mind this code `reactive(teal_data())`? In my mind it's `reactive teal data`. It doesn't matter what we use in `tmg` or `tmc`. It's actually `tmg` and `tmc` that should follow `teal`. Anyway, not a big deal

R/module_nested_tabs.R
```diff
@@ -263,7 +263,7 @@ srv_teal_module.teal_module <- function(id,
       is_transform_failed <- reactiveValues()
       transformed_teal_data <- srv_transform_teal_data(
         "data_transform",
-        data = filtered_teal_data,
+        teal_data_r = filtered_teal_data,
```
m7pr, line 266: @llrs-roche since `srv_transform_teal_data` is exported, and you changed the name of the parameter, then I would expect you also provide a PR to `tmg` and `tmc` with this change. 
https://github.com/insightsengineering/teal.modules.general/blob/main/R/utils.R#L306
Please also adjust `srv_decorate_teal_data`

R/dummy_functions.R
llrs-roche: Ok, yes this was perhaps the more controversial decision while renaming, I will rename to r_real_data. I too think it will be easier to read  (and maybe leave a note/issue on tmg and tmc to match this).

R/module_nested_tabs.R
llrs-roche: I will send the accompanying PRs. I first wanted to have the renaming approved before start opening PR to other repositories.

m7pr: yes, I think we can go with `r_teal_data`

gogonzo: I'm not a fan of renaming `data` to both `teal_data_r` and `r_teal_data`. In the server of teal_module we have argument `data` which is a "reactive returning teal_data object".  `srv_decorate_teal_data` have `data` argument which is exactly the same class. Module dev is going to have two different argument names for the same object in the exported functions just to "adjust" names with .

This is double sided sword, as we have two definitions of `data`: `module$server` has `data` described above and `init` has an argument `data` (teal_data (static), reactive teal_data and teal_data_module).

-----

Let's consider perspective of two different users:

- app developer provides a `teal_data` to the `teal::init(data)` and specifies `tm_xyz(transformators)`. `transformator` has a server `function(id, data)` where `data` is a `reactive teal_data`. App developer might be confused by this, assuming that transformator's `data` is the same object.
- module developer makes the code of the server where `data` is a `reactive teal_data`. This `data` object is chained via some reactives (usually prefixed something_q) and passed to the `srv_decorate_teal_data(data)`. Changing to `srv_decorate_teal_data(teal_data_r or r_teal_data)` is nto compatible with current naming standard in the module's server, where we use:
    - <something>_r - reactive returning object like plot, data.frame (often taken from inside of teal_data).
    - <something>_q - reactive returning teal_data (qenv) object.

gogonzo: If we decide to name "reactive teal_data" from `data` to `r_data` then we should be consistent and change it:
- in `module(server)` ⚠️ 
- in `teal_transform_module(server)`
-  `srv_transform_teal_data`, `srv_decorate_teal_data`

gogonzo:
Renaming is done selectively, there are bunch of functions which still haven't been touch. This is a diagram of `data` processing in teal. I've ommited some calls like `srv_check_module_datanames` or `.resolve_module_datanames`. Bold boarder means exported functions. Please analyse and compare with the changes in PR - there are some gaps. `init` and `module$server` have been released for a long time and any changes there will require unanimous decision and deprecation process (probably out of the discussion)

 
data arg in teal

R/module_nested_tabs.R
llrs-roche: Yes, I feared that changes in internal parameters could end up leaking to major exported functions. 

It seems that to be more consistent we would need to modify exported functions. @pawelru, any preference or comment on this discussion?

R/module_transform_data.R
```diff
@@ -74,7 +74,11 @@ ui_transform_teal_data <- function(id, transformators, class = "well") {
 
 #' @export
 #' @rdname module_transform_data
-srv_transform_teal_data <- function(id, data, transformators, modules = NULL, is_transform_failed = reactiveValues()) {
+srv_transform_teal_data <- function(id,
+                                    data,
+                                    transformators,
+                                    modules = NULL,
+                                    is_transform_failed = reactiveValues()) {
```
gogonzo, line 81: this change is not needed. Theoretically nothing has changed

gogonzo (commented):
Please check these functions:

- In `.call_teal_module`: `filtered_teal_data` -> `data` and set `checkmate::assert_class(data, "teal_data")`
- in `srv_data_summary`: `teal_data` -> `data`

R/module_transform_data.R
llrs-roche: Ah, yes this is a leftover from teal_data_r pushing the line over the length limits on the style. I'll remove it.

llrs-roche:
As discussed on the SU, I renamed parameters back to data (on this PR the different commits replaced like this: data -> teal_data_r -> data), and renamed `data_rv`, `filtered_teal_data`  and `teal_data` parameter to `data`. This simplifies the different parameters used across teal and instead of having 3 diferents parameters we will have one. This will avoid affecting the end users and issues with documentation diverging on different places. 

The data parameter may refer to different classes, but developers should rely on the assertions to know which class is accepted instead of the name.  

@gogonzo setting `checkmate::assert_class(data, "teal_data")` inside `.call_tea_module` caused more than 50 test failures. I'm replacing the assertion with "reactives" as this is what it seems to be expected on tests.

R/module_nested_tabs.R
pawelru: Thanks for the ping. I think that the teal framework is pretty mature hence I think I would rather prefer limiting breaking changes

R/module_init_data.R
```diff
@@ -53,7 +53,7 @@ srv_init_data <- function(id, data) {
 
   moduleServer(id, function(input, output, session) {
     logger::log_debug("srv_data initializing.")
-    # data_rv contains teal_data object
+    # data contains teal_data object
```
gogonzo, line 56: `data` in this comment rather refers to the `data_out` but I think the whole comment can be removed

gogonzo (commented):
> @gogonzo setting checkmate::assert_class(data, "teal_data") inside .call_tea_module caused more than 50 test failures. I'm replacing the assertion with "reactives" as this is what it seems to be expected on tests.

Yup, I'm glad you've figured it out. I have no further comments

R/module_teal.R
```diff
@@ -201,7 +201,7 @@ srv_teal <- function(id, data, modules, filter = teal_slices()) {
 
     data_validated <- .trigger_on_success(data_pulled)
 
-    data_rv <- reactive({
+    data <- reactive({
```
gogonzo, line 204: Possible collision by overwriting a `data` argument

R/module_teal.R
```diff
@@ -201,7 +201,7 @@ srv_teal <- function(id, data, modules, filter = teal_slices()) {
 
     data_validated <- .trigger_on_success(data_pulled)
 
-    data_rv <- reactive({
+    data_validated_r <- reactive({
```
gogonzo, line 204: Suddenly `_r` appears. I propose:

`data_pulled` -> `data_handled` (to follow object name returned from srv_init_data)
`data_validated` -> `data_triggered` (or keep `data_validated`)
`data_validated_r` -> `data_signatured`

gogonzo (approved):
👍

PR #1432, opened 2024-12-18
TealAppDriver skips tests when shinytest2 or rvest are not installed

averissimo:
# Pull Request

Part of https://github.com/insightsengineering/teal.modules.general/pull/774

See https://github.com/insightsengineering/teal.modules.general/pull/774#discussion_r1718027258


R/TealAppDriver.R
```diff
@@ -10,12 +10,15 @@
...
   inherit = {
-    if (!requireNamespace("shinytest2", quietly = TRUE)) {
-      stop("Please install 'shinytest2' package to use this class.")
-    }
-    if (!requireNamespace("rvest", quietly = TRUE)) {
-      stop("Please install 'rvest' package to use this class.")
-    }
+    lapply(c("shinytest2", "rvest"), function(.x, use_testthat) {
+      if (!requireNamespace(.x, quietly = TRUE)) {
+        if (use_testthat) {
```
llrs-roche, line 15: What about inverting the order of the if else clause, so if there is no testthat and it is not in use the TealAppDriver class won't check for the suggested package used for testing (this uses && shortcut) .

Here is my suggestion (indentation would need to change accordingly): 

```suggestion
      if (use_testthat && !requireNamespace(.x, quietly = TRUE)) {
```

averissimo: Am I misunderstanding or would we still need an `else if` after for the current `else` clause with your suggestion?

```r
if (use_testthat && !requireNamespace(.x, quietly = TRUE)) {
  testthat::skip(sprintf("%s is not installed", .x))
} else if (!requireNamespace(.x, quietly = TRUE)) {
  stop("Please install '", .x, "' package to use this class.", call. = FALSE)
}
```

averissimo: Note that we could remove the `else` from the current code to make it a bit leaner

```r
      if (!requireNamespace(.x, quietly = TRUE)) {
        if (use_testthat) {
          testthat::skip(sprintf("%s is not installed", .x))
        }
        stop("Please install '", .x, "' package to use this class.", call. = FALSE)
      }
```

Second note: for a moment I considered the following, but the `tryCatch` would be too broad and from all the options I dislike this the most

```r
if (!requireNamespace(.x, quietly = TRUE)) {
  tryCatch(
    testthat::skip(sprintf("%s is not installed", .x)),
    error = function(err) stop("Please install '", .x, "' package to use this class.", call. = FALSE)
  )
}
```

llrs-roche: I was thinking this code will be executed mostly by developers checking & testing the package, so that should be checked first as it can lead to speed improvements and code readability. I think t

I would expect developers to know they need to install a package if it cannot be loaded and a message reported this. 
What do you think about this simplification? 

```r
      if (requireNamespace("testthat", quietly = TRUE) && testthat::is_testing() identical(Sys.getenv("RTESTS"), "true")) {
          lapply(c("shinytest2", "rvest"), testthat::skip_if_not_installed)
      }

```

The `identical(Sys.getenv("RTESTS"), "true")` is to ensure that when one does devtools::check it is triggered (but can be omitted). 

I tested several snippets but I'm not happy with any regarding readability, perhaps it would be easier if we had a little helper/uril function to solve this that we can reuse across packages.

llrs-roche (commented):
The code works well, but I am not sure about the readability (but I also don't think my proposal is easier to read)...

llrs-roche (approved):
Sorry, I forgot to aprove yesterday!

PR #1435, opened 2024-12-19
Fixes broken links by leveraging `pkgdown` autodetection of function use in inline code

averissimo:
#### Changes description

- Change absolute paths to `insightsengineering.github.io/teal/...` to `fun_name()`
    - downside:
        - `knitr` keeps the inline code as is, so CRAN's vignettes don't have the links
    - upside:
        - pkgdown is smart to detect and create a link to the reference page (see within)
        - `urlcheck` CI doesn't fail on `main` and PRs with unreleased code
        - Articles in version-specific documentation don't suddenly take you to `latest-tag`

#### Alternatives considered

Using relative links (`../reference`), but this will break CRAN vignettes as well

### DIFF
--- a/vignettes/decorate-module-output.Rmd
+++ b/vignettes/decorate-module-output.Rmd
@@ -79,7 +79,7 @@ static_decorator <- teal_transform_module(
 
 To simplify the repetitive elements of writing new decorators 
 (e.g., `function(id, data), moduleServer, reactive, within(data, ...)`), 
-you can use the `make_teal_transform_server()`] convenience function, which takes a `language` as input:
+you can use the `make_teal_transform_server()` convenience function, which takes a `language` as input:
 
 ```{r}
 static_decorator_lang <- teal_transform_module(
--- a/vignettes/decorate-module-output.Rmd
+++ b/vignettes/decorate-module-output.Rmd
@@ -14,7 +14,7 @@ vignette: >
 
 `teal` is a powerful `shiny`-based framework with built-in modules for interactive data analysis. 
 This document outlines the customization options available for modifying the output of `teal` modules. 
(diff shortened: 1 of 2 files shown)


vignettes/decorate-module-output.Rmd
```diff
@@ -79,7 +79,7 @@ static_decorator <- teal_transform_module(
 
 To simplify the repetitive elements of writing new decorators 
 (e.g., `function(id, data), moduleServer, reactive, within(data, ...)`), 
-you can use the [`make_teal_transform_server()`](https://insightsengineering.github.io/teal/latest-tag/reference/make_teal_transform_server.html) convenience function, which takes a `language` as input:
+you can use the `make_teal_transform_server()`] convenience function, which takes a `language` as input:
```
gogonzo, line 82:
```suggestion
you can use the `make_teal_transform_server()` convenience function, which takes a `language` as input:
```

gogonzo (commented):
Found out that `https://rstudio.github.io/shiny/reference/moduleServer.html` can be replaced with `shiny::moduleServer()`. Does it make sense to do this in this PR?

averissimo:
> Found out that `https://rstudio.github.io/shiny/reference/moduleServer.html` can be replaced with `shiny::moduleServer()`. Does it make sense to do this in this PR?

@gogonzo I noticed that too on the blueprint when searching for `/references`, I didn't want to replace it as the link makes sense in the text, I'm neutral on changing it,  if so, maybe we could use something like:

```diff
-The `teal` framework leverages the [`shiny` module concept](https://rstudio.github.io/shiny/reference/moduleServer.html) to enable encapsulation of analytical actions in `teal` modules, while maintaining seamless communication between the modules and the application.
+The `teal` framework leverages the `shiny` module concept (`shiny::moduleServer()`) to enable encapsulation of analytical actions in `teal` modules, while maintaining seamless communication between the modules and the application.
```

gogonzo (approved):
👍 Please ignore my comment about `moduleServer`

Issue #1143, opened 2024-03-11
use "|>" for teal families, as teal's R dependency is already at 4.0


shajoezhu:
ok, i think we can do "|>" for teal families, as teal's R dependency is already at 4.0

but we can leave the static output ones a bit behind, (including tern)

eventually, when making cards an upstream dependencies of tern, we will switch |> for tern, the rest can stay %>% for a bit longer

_Originally posted by @shajoezhu in https://github.com/insightsengineering/rtables/issues/838#issuecomment-1983554000_

m7pr:
`|>` was introduced in 4.1.0, not 4.0.0

https://cran.r-project.org/bin/windows/base/old/4.1.0/NEWS.R-4.1.0.html

> R now provides a simple native forward pipe syntax |>. The simple form of the forward pipe inserts the left-hand side as the first argument in the right-hand side call. The pipe implementation as a syntax transformation was motivated by suggestions from Jim Hester and Lionel Henry.

m7pr:
And `teal` just depends on 4.0.0 
https://github.com/insightsengineering/teal/blob/beeb8da362e8c471b1ae30e4559244769defc947/DESCRIPTION#L38

averissimo:
4.1 is 3 years old, maybe it's time we bump it :-)

vedhav:
R version bumped in PR #1440

Issue #1268, opened 2024-07-19
[Question]: Should `teal` codebase use `magrittr`


m7pr:
Currently `magrittr` pipe is reexported https://github.com/insightsengineering/teal/blob/f0e3969141abb97587e696dbb8a57ad081949173/R/teal.R#L14

The motivation for that is probably for the users to be able to use pipe, after teal is attached. Should we use `%>%` in the codebase, since it's already in Imports, or should we reconsider cleaning this up?

chlebowa:
Related issue: https://github.com/insightsengineering/teal.modules.general/issues/786

m7pr:
I agree with the removal of `%>%` from the codebase, but I'm not sure we really rely on `4.1`

chlebowa:
`teal` `Depends` on R 4.0.

averissimo:
`{teal}`'s current dependencies depend on `4.4` due to `{Matrix}` package, but I think there's no real technical requirement to force us to bump to `4.1`

Users using a CRAN snapshot from before the 4.4 release date can probably use 4.0.

donyunardi:
The teal codebase and its framework packages no longer use `|>` or `%>%`. We have also removed the `%>%` dependency mentioned in the original post.

So, the consensus is to keep the R version dependency for teal at 4.0.

However, [tmg](https://github.com/insightsengineering/teal.modules.general/blob/643644eb0d839c755cc796f6074b6f8c4b9b59ea/R/teal.modules.general.R#L11) and [tmc](https://github.com/insightsengineering/teal.modules.clinical/blob/af36c1740e9016cab0fbf3fd9d48959472a7d90c/R/teal.modules.clinical.R#L12) are still uses `%>%`.

We could consider removing this dependency from the teal module packages that we maintain.
Additionally, please be mindful of the behavior of `|>` in the teal_data/qenv context on code reproducibility.
![image](https://github.com/user-attachments/assets/d571271b-f776-450d-bdcc-407bfba0562d)

As you see, the code is updated to not use the `|>` operator.

And here is example if we use `%>%` operator where code stored with the operator.
![image](https://github.com/user-attachments/assets/f43503a2-c997-46a4-84c2-658b92e50d02)

gogonzo:
Closing this as @averissimo made it clear in his comments

PR #1440, opened 2025-01-07
Add teal app modifiers and deprecate init args

### LINKED ISSUE #1310: Move landing popup feature back to the `modules` argument
Recently we have moved the landing popup feature to become one of the arguments of `init()`. I don't think it's correct. If you compare the relevance of all the other arguments including `modules` or `data` - this is clearly not a good fit. Conceptually this should remain as a module. Let's move it back.

vedhav:
Closes #1310 and #1143

### Changes

- The `title`, `header`, and `footer` are soft deprecated in favor of using `add_title`, `add_header`, `add_footer`, and `add_landing_popup`
- Adds a new `add_custom_server` modifier to add custom server logic to the main shiny app.
- Reverts the usage of `landing_popup_module` to the previous way by passing it inside the `modules`
- Bumps R version to 4.1 to start using native pipe `|>`

### Example app with the new app modifiers.

```r
devtools::load_all("teal")

app <- init(
  modules = modules(
    example_module("Module 1"),
    example_module("Module 2")
  ),
  filter = teal_slices(
    teal_slice(dataname = "IRIS", varname = "Species", selected = "setosa")
  )
) |>
  modify_title("Custom title") |>
  modify_header(
    tags$div(
      h3("Header with a button that can be observed!"),
      actionButton("notify", "Show notification")
    )
  ) |>
  modify_footer("Custom footer") |>
   add_landing_popup(
      content = "Popup content",
      buttons = modalButton("Proceed")
   ) |>
  add_custom_server(function(input, output, session) {
    observeEvent(input$notify, {
      showNotification("Yes, the button works!")
    })
  })

shinyApp(app$ui, app$server)
```


R/init.R
```diff
@@ -222,27 +186,246 @@ init <- function(data,
...
-  res <- list(
-    ui = function(request) {
-      ui_teal(
-        id = ns("teal"),
-        modules = modules,
-        title = title,
-        header = header,
-        footer = footer
+  res <- list()
+  res$landing_popup_server <- NULL
```
gogonzo, line 190: `res$landing_popup_server` is not needed as it can be passed directly to do.call down there ⬇️ 
```suggestion
```

R/init.R
```diff
@@ -222,27 +186,246 @@ init <- function(data,
...
+        tags$title(title),
+        tags$link(
+          rel = "icon",
+          href = favicon,
+          sizes = "any"
+        )
+      )
+    )
+    ui_tq <- do.call(app$ui, c(list(request = request), args)) |>
+      htmltools::tagQuery()
```
gogonzo, line 293: No pipe before changing R version in Description file
```suggestion
    ui_tq <- htmltools::tagQuery(do.call(app$ui, c(list(request = request), args)))
```

R/init.R
```diff
@@ -97,14 +98,12 @@ init <- function(data,
                  title = build_app_title(),
```
gogonzo, line 98: Deprecated function can't be a default. It will throw each time when `title` is not specified

R/init.R
```diff
@@ -222,27 +186,246 @@ init <- function(data,
...
+    showModal(
+      modalDialog(
+        id = id,
+        title = title,
+        content,
+        footer = buttons
+      )
+    )
+  }
+  app
```
gogonzo, line 403:
```suggestion
  custom_server <- showModal(
      modalDialog(
        id = id,
        title = title,
        content,
        footer = buttons
      )
    )
    add_custom_server(app, custom_server)
```

R/init.R
```diff
@@ -222,27 +186,246 @@ init <- function(data,
...
+        "Use `modify_footer()` on the teal app object instead."
+      )
+    )
+  }
+
+  # argument transformations
+  ## `modules` - landing module
+  landing <- extract_module(modules, "teal_module_landing")
+  if (length(landing) == 1L) {
+    res$landing_popup_server <- landing[[1L]]$server
```
gogonzo, line 225: We can use `add_custom_server` here and move this `if` after `server <- `
```suggestion
    app <- add_custom_server(app, landing[[1L]]$server)
```

R/init.R
```diff
@@ -222,27 +186,246 @@ init <- function(data,
...
+        "landing_popup_module() is deprecated.",
+        "Use add_landing_popup() on the teal app object instead."
+      )
+    )
+  } else if (length(landing) > 1L) {
+    stop("Only one `landing_popup_module` can be used.")
+  }
+
+  # Note: UI must be a function to support bookmarking.
+  res$ui <- function(request, ...) {
```
gogonzo, line 240:
```suggestion
  ui <- function(request, ...) {
```

R/init.R
```diff
@@ -222,27 +186,246 @@ init <- function(data,
...
+    ui_teal(
+      id = "teal",
+      modules = modules,
+      title = title,
+      header = header,
+      footer = footer
+    )
+  }
+  res$server <- function(input, output, session) {
```
gogonzo, line 249:
```suggestion
  server <- function(input, output, session) {
```

R/init.R
```diff
@@ -222,27 +186,246 @@ init <- function(data,
...
+#' @description Adds a custom server function to the `teal` app. This function can define additional server logic.
+#'
+#' @param app (`environment`) The `teal` app environment.
+#' @param custom_server (`function`) The custom server function to set.
+#' @export
+#' @examples
+#' app <- init(
+#'   data = teal_data(IRIS = iris),
+#'   modules = modules(example_module())
+#' ) |>
```
gogonzo, line 417: pipe available from R 4.1

gogonzo (commented):
Some room for improvement

R/init.R
vedhav: Yeah, that's why I changed it to `deprecate_soft()`. This function will still exist, it will just become internal in the next release. Because we use it in two places.

R/init.R
vedhav: I remember we had a discussion about this earlier. Did we decide on bumping the R version? It's already over 3 years since `|>` is available.

R/init.R
gogonzo: Yes, but it will throw deprecation even if one doesn't do anything with `title`. It will for sure come up in tests of init

R/init.R
vedhav: I think the if has to stay above because it removes the landing popup from the `modules`.

R/init.R
gogonzo: If you'd like to bump up R in the Description please link [related issue](https://github.com/insightsengineering/teal/issues/1143) and put issue on the board. I would accept this change thanks to [this comment](https://github.com/insightsengineering/teal/issues/1268#issuecomment-2406719999)

R/init.R
```diff
@@ -222,27 +186,242 @@ init <- function(data,
...
+  res$ui <- function(request, ...) {
+    args <- list(...)
+    args$title <- tags$div(
+      id = "teal-title",
+      build_app_title(title, favicon)
+    )
+    ui_tq <- htmltools::tagQuery(do.call(app$ui, c(list(request = request), args)))
+
+    ui_tq$find("#teal-title")$replaceWith(args$title)$allTags()
+  }
```
gogonzo, line 293: `...` seems to be redundant everywhere as in `init` it is not consumed.

```suggestion
  res$ui <- function(request) {
    title <- tags$div(
      id = "teal-title",
      build_app_title(title, favicon)
    )
    ui_tq <- htmltools::tagQuery(app$ui(request = request))

    ui_tq$find("#teal-title")$replaceWith(title)$allTags()
  }
```

R/init.R
vedhav: Let me rearrange and make use of `add_custom_server`

R/init.R
```diff
@@ -222,27 +186,235 @@ init <- function(data,
...
+#'   add_custom_server(function(input, output, session) {
+#'   })
+#'
+#' shinyApp(app$ui, app$server)
+add_custom_server <- function(app, custom_server) {
+  old_server <- app$server
+
+  app$server <- function(input, output, session) {
+    old_server(input, output, session)
+    custom_server(input, output, session)
```
gogonzo, line 417:
```suggestion
    if (all(c("input", "output", "session") %in% names(formals(custom_server)))) {
      custom_server(input, output, session)
    } else if ("id" %in% names(formals(custom_server))) {
      custom_server("wrapper")
    }
```

vedhav: Once we deprecate the `id` parameter, this level will be a non-shiny-module root and I think it will keep things simpler if we keep it that way.
Do you see value in allowing app developers to pass shiny module ui in header/footer and adding the server module here?

gogonzo: I'm not pushing this. I just realised that we shouldn't call custom_server(input, output, session) but use `callModule(custom_server, "wrapper") instead to separate possible namespace collision.

vedhav: There won't be any namespace collision as `ui_teal`/`srv_teal` is shiny module with a "teal" namespace and the root level of shiny is free for the taking.

gogonzo: What if inside of the custom server one calls a module with the namespace `teal`

R/init.R
```diff
@@ -222,27 +186,235 @@ init <- function(data,
...
+    old_server(input, output, session)
+    showModal(
+      modalDialog(
+        id = id,
+        title = title,
+        content,
+        footer = buttons
+      )
+    )
+  }
```
gogonzo, line 392:
```suggestion
  custom_server <- function(input, output, session() {
    showModal(
      modalDialog(
        id = id,
        title = title,
        content,
        footer = buttons
      )
    )
  }
  app$server <- add_custom_server(app, custom_server)
```

R/init.R
```diff
@@ -94,17 +95,15 @@
...
                  modules,
                  filter = teal_slices(),
-                 title = build_app_title(),
-                 header = tags$p(),
-                 footer = tags$p(),
-                 id = character(0),
-                 landing_popup = NULL) {
+                 title = NULL,
+                 header = NULL,
+                 footer = NULL,
```
gogonzo, line 100: Please use `lifecycle::deprecated()` and `lifecycle::is_present()` instead of handling `NULL`

```suggestion
                 title = lifecycle::deprecated(),
                 header = lifecycle::deprecated(),
                 footer = lifecycle::deprecated(),
```

R/init.R
```diff
@@ -222,27 +186,235 @@ init <- function(data,
...
+#'   modules = modules(example_module())
+#' ) |>
+#'   modify_footer("Custom footer")
+#'
+#' shinyApp(app$ui, app$server)
+modify_footer <- function(app, footer = tags$p()) {
+  res <- app
+  res$ui <- function(request) {
+    ui_tq <- htmltools::tagQuery(app$ui(request = request))
+    ui_tq$find("#teal-footer")$replaceWith(tags$div(id = "teal-footer", footer))$allTags()
```
gogonzo, line 342:
```suggestion
    ui_tq$find("#teal-footer")$replaceWith(tags$footer(id = "teal-footer", footer))$allTags()
```

R/TealAppDriver.R
```diff
@@ -45,25 +45,51 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
+          list(
+            id = "landingpopup",
+            title = NULL,
+            content = NULL,
+            buttons = modalButton("Accept")
+          ),
+          landing_popup_args
+        )
+        app <- app |>
+          add_landing_popup(
```
gogonzo, line 86: To introduce the pipe operator in the internal code we need to agree to certain style rules within a team. Pipe might be overused to replace a single call which in my opinion is bad as it doesn't bring clarity plus it is a extra sugar to evaluate.

```suggestion
        app <- add_landing_popup(
          app,
```

R/TealAppDriver.R
```diff
@@ -45,25 +45,51 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
         data = data,
         modules = modules,
-        filter = filter,
-        title = title,
-        header = header,
-        footer = footer,
-        landing_popup = landing_popup,
-      )
+        filter = filter
+      ) |>
```
gogonzo, line 70: Please bump dependency on `R > 4.1`

gogonzo (commented):
`ui_teal` shouldn't have `header`, `footer`, `title` arguments. Deprecation in init only is a half-measure. Please use `add_header` and `add_title` in `teal::init` body instead of passing arguments to `ui_teal`.

Edit: Same applies to `module_teal_with_splash`

R/init.R
```diff
@@ -121,46 +120,11 @@ init <- function(data,
...
-  checkmate::assert(
-    .var.name = "header",
-    checkmate::check_string(header),
-    checkmate::check_multi_class(header, c("shiny.tag", "shiny.tag.list", "html"))
-  )
-  checkmate::assert(
-    .var.name = "footer",
-    checkmate::check_string(footer),
-    checkmate::check_multi_class(footer, c("shiny.tag", "shiny.tag.list", "html"))
-  )
```
m7pr, line 140: maybe do not delete, but wrap around `lifecycle::is_present`. delete only when parameters are deleted in the future?

R/init.R
```diff
@@ -222,27 +186,235 @@ init <- function(data,
...
+  } else if (length(landing) > 1L) {
+    stop("Only one `landing_popup_module` can be used.")
+  }
+
   logger::log_debug("init teal app has been initialized.")
 
   res
 }
+
+#' Add a custom title to the app
```
m7pr, line 266:
```suggestion
#' Add a custom title to `teal` application
```

R/init.R
```diff
@@ -222,27 +186,235 @@ init <- function(data,
...
+#'
+#' @param title The title to add
```
m7pr, line 268:
```suggestion
#' @param title The title to be used
```

R/init.R
```diff
@@ -222,27 +186,235 @@ init <- function(data,
...
+#'   data = teal_data(IRIS = iris, MTCARS = mtcars),
+#'   modules = modules(example_module())
+#' ) |>
+#'   modify_title("Custom title")
+#'
+#' shinyApp(app$ui, app$server)
+modify_title <- function(
+    app,
+    title = "teal app",
+    favicon = "https://raw.githubusercontent.com/insightsengineering/hex-stickers/main/PNG/nest.png") {
```
m7pr, line 282:
```suggestion
    favicon = "https://raw.githubusercontent.com/insightsengineering/hex-stickers/main/PNG/nest.png"
    ) {
```

R/init.R
```diff
@@ -222,27 +186,235 @@ init <- function(data,
...
+      id = "teal-title",
+      build_app_title(title, favicon)
+    )
+    ui_tq <- htmltools::tagQuery(app$ui(request = request))
+    ui_tq$find("#teal-title")$replaceWith(title_tag)$allTags()
+  }
+  res
+}
+
+#' Add a Header to a `teal` App
```
m7pr, line 295:
```suggestion
#' Add a Header to `teal` Application
```

R/init.R
m7pr:
```r
(`shiny.tag` or `character(1)`)
```
?

R/init.R
```diff
@@ -222,27 +186,235 @@ init <- function(data,
...
+modify_header <- function(app, header = tags$p()) {
+  res <- app
+  res$ui <- function(request) {
+    ui_tq <- htmltools::tagQuery(app$ui(request = request))
+    ui_tq$find("#teal-header")$replaceWith(tags$header(id = "teal-header", header))$allTags()
+  }
+  res
+}
+
+#' Add a Footer to a `teal` App
```
m7pr, line 323:
```suggestion
#' Add a Footer to `teal` Application
```

R/init.R
```diff
@@ -222,27 +186,235 @@ init <- function(data,
...
+modify_footer <- function(app, footer = tags$p()) {
+  res <- app
+  res$ui <- function(request) {
+    ui_tq <- htmltools::tagQuery(app$ui(request = request))
+    ui_tq$find("#teal-footer")$replaceWith(tags$div(id = "teal-footer", footer))$allTags()
+  }
+  res
+}
+
+#' Add a Landing Popup to a `teal` App
```
m7pr, line 347:
```suggestion
#' Add a Landing Popup to `teal` Application
```

R/init.R
```diff
@@ -222,27 +186,235 @@ init <- function(data,
...
+        content,
+        footer = buttons
+      )
+    )
+  }
+  app$server <- add_custom_server(app, custom_server)
+  app
+}
+
+#' Add a Custom Server Logic to a `teal` App
```
m7pr, line 396:
```suggestion
#' Add a Custom Server Logic to `teal` Application
```

_pkgdown.yml
```diff
@@ -108,6 +108,14 @@ reference:
       - module
       - modules
       - teal_slices
+  - title: Teal app modifiers
```
m7pr, line 111:
```suggestion
  - title: `teal` application modifiers
```

R/init.R
vedhav:
```suggestion
#' @param title (`shiny.tag` or `character(1)`) The title to be used
```
👍🏽

R/init.R
```diff
@@ -222,27 +186,235 @@ init <- function(data,
...
-  ns <- NS(id)
+  if (!is.null(title)) {
+    checkmate::assert_multi_class(title, c("shiny.tag", "shiny.tag.list", "html", "character"))
+    lifecycle::deprecate_warn(
+      when = "0.15.3",
+      what = "init(title)",
+      details = "Use `modify_title()` on the teal app object instead."
+    )
+  } else {
+    title <- build_app_title()
```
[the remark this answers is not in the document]
vedhav: I'm on the fence here. I don't mind removing it completely or stop exporting it. I lean towards completely removing it.

R/init.R
vedhav: So, I think it will be reverted once we format using formatter

m7pr:
Hey, really great start! A food for thought below

- I think this solution would benefit if we can craft a vignette explaining how you can customise teal app
- Just an idea, for another issue: what if we allow user to modify the layout (allow filter panel to be on the left, allow to pass any custom JS) of aesthetics (colours, fonts) of the app if modify_layout/modify_ui function?

R/init.R
vedhav: We want to replace the div that encloses the footer in the `ui_teal`:
```r
tags$div(id = "teal-footer", footer)
```

R/init.R
gogonzo: @vedhav ⬆️

R/init.R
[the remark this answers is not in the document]
gogonzo: This is an exported function. We should soft_deprecated it first. It should be also taken away from default ui_teal(title) and ui_teal_with_splash(title).

R/init.R
gogonzo: > What if inside of the custom server one calls a module with the namespace teal

@vedhav Sorry I didn't use question mark in the above comment. I hope we continue to discuss this. 

> There won't be any namespace collision as ui_teal/srv_teal is shiny module with a "teal" namespace and the root level of shiny session namespace is free for the taking.

Yes, there will be namespace collision if somebody calls twice `add_custom_server` where some `ids` are duplicated. There will be a collision because if we don't `callModule` both modules would operate under `teal` namespace.

My suggestion though set another trap - if one uses `add_custom_server` twice, then both will be called with an `id ="wrapper"` and we'll have a collision.

P.S. you've committed my two suggestions where `assert_function(custom_server, args = c("input", "output", "session")` makes `if ("id" %in% names(formals(...)))` impossible.

vedhav: I applied the suggestion and added an example to call the module server. Please have a look tomorrow!

R/teal_modifiers.R
```diff
@@ -0,0 +1,193 @@
...
+#'   )
+#'
+#' if (interactive()) {
+#'   shinyApp(app$ui, app$server)
+#' }
+add_landing_popup <- function(
+    x,
+    title = NULL,
+    content = NULL,
+    buttons = modalButton("Accept")) {
```
gogonzo, line 144: I think function could be named `add_landing_modal`. Arguments are passed to `modalDialog` and I think they should be inherited. `modalDialog` has also `size`, `easyClose`, `fade` arguments - these we can consider to add in the future (explicitly or through ...). 

```suggestion
    title = NULL,
    content = NULL,
    footer = modalButton("Accept")) {
```

R/teal_modifiers.R
```diff
@@ -0,0 +1,193 @@
...
+#' }
+modify_footer <- function(x, element = tags$p()) {
+  checkmate::assert_multi_class(element, c("shiny.tag", "shiny.tag.list", "html", "character"))
+  teal_replace_ui(x, "#teal-footer-content", element)
+}
+
+#' Add a Landing Popup to `teal` Application
+#'
+#' @description Adds a landing popup to the `teal` app. This popup will be shown when the app starts.
+#' The dialog blocks access to the application and must be closed with a button before the application can be viewed.
```
gogonzo, line 118:
```suggestion
#' The dialog must be closed by the app user to proceed to the main application.
```

R/teal_modifiers.R
```diff
@@ -0,0 +1,193 @@
...
+#' app <- init(
+#'   data = teal_data(IRIS = iris),
+#'   modules = modules(example_module())
+#' ) |>
+#'   modify_header(element = tags$div(h3("Custom header")))
+#'
+#' if (interactive()) {
+#'   shinyApp(app$ui, app$server)
+#' }
+modify_header <- function(x, element = tags$p()) {
```
gogonzo, line 93:
```suggestion
modify_header <- function(x, element = tags$p()) {
  checkmate::assert_multi_class(x, "teal_app")
```

R/teal_modifiers.R
```diff
@@ -0,0 +1,193 @@
...
+#' app <- init(
+#'   data = teal_data(IRIS = iris),
+#'   modules = modules(example_module())
+#' ) |>
+#'   modify_footer(element = "Custom footer")
+#'
+#' if (interactive()) {
+#'   shinyApp(app$ui, app$server)
+#' }
+modify_footer <- function(x, element = tags$p()) {
```
gogonzo, line 110:
```suggestion
modify_footer <- function(x, element = tags$p()) {
  checkmate::assert_multi_class(x, "teal_app")
```

R/module_teal.R
```diff
@@ -42,42 +42,8 @@ NULL
 
 #' @rdname module_teal
 #' @export
-ui_teal <- function(id,
-                    modules,
-                    title = build_app_title(),
-                    header = tags$p(),
-                    footer = tags$p()) {
+ui_teal <- function(id, modules) {
```
gogonzo, line 45: documentation of `module_teal` needs to be modified as it says `this module can be called directly instead of and included in your custom application`. We need to note that it lacks header and footer.

R/module_teal.R
```diff
@@ -132,20 +96,15 @@ ui_teal <- function(id,
...
-    )
+    tags$hr()
   )
 }
 
 #' @rdname module_teal
+#' @param modules (`teal_modules`)
+#'   `teal_modules` object. These are the specific output modules which
+#'   will be displayed in the `teal` application. See [modules()] and [module()] for
+#'   more details.
```
gogonzo, line 107: Move this documentation to the common `module_teal` docs as ui also has a `modules` argument. I know it will be rendered in common docs anyway but this is inconsistent location for docs.

R/teal_modifiers.R
vedhav: I think there is no harm in adding `...` even now. I'll add it.

R/module_teal.R
vedhav: The reason I added it explicitly for the `srv_teal` is because it only accepts a `teal_modules` object, while the `ui_teal` and `init` can also accept `teal_module` object.

R/module_teal.R
vedhav: Please have a look at the document now.

R/module_teal.R
gogonzo: I think it is inconsistent then between `ui_teal`/`srv_teal` so `ui_teal` should have `checkmate::assert_class(modules, "teal_modules)`

R/teal_modifiers.R
```diff
@@ -0,0 +1,193 @@
...
+#'
+#' @description Adds a landing popup to the `teal` app. This popup will be shown when the app starts.
+#' The dialog must be closed by the app user to proceed to the main application.
+#'
+#' @param x (`teal_app`) A `teal_app` object created using the `init` function.
+#' @param title (`character(1)`) Text to be displayed as popup title.
+#' @param content (`character(1)`, `shiny.tag` or `shiny.tag.list`) with the content of the popup.
+#'  Passed to `...` of `shiny::modalDialog`.
+#' @param footer (`shiny.tag` or `shiny.tag.list`) Typically a `modalButton` or `actionButton`.
+#' @param ... Additional arguments to `shiny::modalDialog`.
```
gogonzo, line 127:
```suggestion
#' @inheritParams shiny::modalDialog
#' @param ... Additional arguments to [shiny::modalDialog()].
```

R/module_teal.R
gogonzo: I think there is a duplicated content between @description and @details section. I'd shorten the description and extend details (without repeating same meaning in few sentences).

The description - `Module to create a `teal` app as a Shiny Module`
The details - the rest combined.

P.S. `reporter_previever_module` should probably be inserted by `ui/srv_teal`. I'll think about this (not for this PR)

R/landing_popup_module.R
```diff
@@ -17,11 +17,11 @@
 #' app1 <- init(
 #'   data = teal_data(iris = iris),
 #'   modules = modules(
-#'     example_module()
-#'   ),
-#'   landing_popup = landing_popup_module(
-#'     content = "A place for the welcome message or a disclaimer statement.",
-#'     buttons = modalButton("Proceed")
+#'     example_module(),
```
gogonzo, line 20: please add deprecated label in the roxy description. Redirect to use `add_landing_modal`. I'd also remove  examples or replace them with `add_landing_modal`.

R/init.R
```diff
@@ -95,17 +99,15 @@
 init <- function(data,
                  modules,
                  filter = teal_slices(),
-                 title = build_app_title(),
-                 header = tags$p(),
-                 footer = tags$p(),
-                 id = lifecycle::deprecated(),
-                 landing_popup = NULL) {
+                 title = lifecycle::deprecated(),
```
gogonzo, line 102: init's example still contains title, header and footer.

vedhav: 🤦🏽 Removed it now

vignettes/quick-usage-guide-for-teal.Rmd
```diff
@@ -0,0 +1,80 @@
+---
+title: "Quick usage guide for teal"
```
gogonzo, line 2: I'm not a fan of this vignette to be created only to show how to modify footer, header and title. Name is competing with "Getting started with teal" and it doesn't say exactly what this vignette is about. Vignette is more about "Modifying teal app elements", but the control we provide is so small that it rather shouldn't be yet published.

vedhav: Okay, makes sense. I will remove it. We can circle back on vignettes later.

gogonzo (approved):
Thanks 👍

R/TealAppDriver.R
```diff
@@ -45,25 +46,51 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
                           load_timeout = rlang::missing_arg(),
                           ...) {
       private$data <- data
       private$modules <- modules
       private$filter <- filter
+
+      new_title <- modifyList(
+        list(
+          title = "Custom Teal App Title",
+          favicon = "https://raw.githubusercontent.com/insightsengineering/hex-stickers/main/PNG/teal.png"
```
pawelru, line 63: Hey @vedhav sorry for commenting in already closed PR but I do have a few comments. Please open a new issue / PR if you think it's relevant and you agree with me.

Can we use the logo file from the installed teal package to avoid url requests? This is under assumption that the logo is already there (haven't checked)

R/init.R
```diff
@@ -237,25 +206,94 @@ init <- function(data,
...
+    list(
+      ui = function(request) {
+        fluidPage(
+          title = tags$div(
+            id = "teal-app-title",
+            tags$head(
+              tags$title("teal app"),
+              tags$link(
+                rel = "icon",
+                href = "https://raw.githubusercontent.com/insightsengineering/hex-stickers/main/PNG/nest.png",
```
pawelru, line 221: https://github.com/insightsengineering/teal/pull/1440/files#r1920393327
same here

R/module_teal_with_splash.R
```diff
@@ -16,24 +17,46 @@ NULL
...
-  ui_teal(id = id, title = title, header = header, footer = footer)
+  ns <- shiny::NS(id)
+  fluidPage(
+    title = tags$div(
+      id = ns("teal-app-title"),
+      tags$head(
+        tags$title("teal app"),
+        tags$link(
+          rel = "icon",
+          href = "https://raw.githubusercontent.com/insightsengineering/hex-stickers/main/PNG/nest.png",
```
pawelru, line 37: https://github.com/insightsengineering/teal/pull/1440/files#r1920393327
same here

R/init.R
```diff
@@ -237,25 +206,94 @@ init <- function(data,
...
-      srv_teal(id = ns("teal"), data = data, modules = modules, filter = deep_copy_filter(filter))
-    }
+    ),
+    class = "teal_app"
   )
 
+  if (lifecycle::is_present(title)) {
+    checkmate::assert_multi_class(title, c("shiny.tag", "shiny.tag.list", "html", "character"))
+    lifecycle::deprecate_warn(
+      when = "0.15.3",
```
pawelru, line 252: the new number has been already determined to be 0.16.0
https://github.com/insightsengineering/teal/issues/1439

R/init.R
```diff
@@ -237,25 +206,94 @@ init <- function(data,
...
+      what = "init(title)",
+      details = "Use `modify_title()` on the teal app object instead."
```
pawelru, line 254: Please include `?modify_title()` in the message so that users can see man page with examples etc.

R/init.R
```diff
@@ -237,25 +206,94 @@ init <- function(data,
...
       }
-      srv_teal(id = ns("teal"), data = data, modules = modules, filter = deep_copy_filter(filter))
-    }
+    ),
+    class = "teal_app"
   )
 
+  if (lifecycle::is_present(title)) {
+    checkmate::assert_multi_class(title, c("shiny.tag", "shiny.tag.list", "html", "character"))
+    lifecycle::deprecate_warn(
```
pawelru, line 251: I thin it's better to throw deprecation warning first before making assertions.

Suppose I provide the `title` argument and I failed assertions. I am putting a lot of effort into "fixing it" just to see that this is deprecated.

NEWS.md
```diff
@@ -4,18 +4,22 @@
 
 * Possible to call `ui_teal` and `srv_teal` directly in any application by delivering `data` argument as a `reactive` returning `teal_data` object. #669
 * Since introduction of `ui_teal` and `srv_teal` functions `id` argument in `init` is being deprecated. #1438
+* Introduce `ui_session_info` and `srv_session_info` shiny module to create the user session info and teal app lockfile lockfile download button.
```
pawelru, line 7: double `"lockfile"`

R/init.R
pawelru: I can see that you have changed old deprecation messages from 0.16.0 into 0.15.3 so in case this change has been confirmed feel free to ignore this comment

R/TealAppDriver.R
vedhav: Thanks for the comments @pawelru. I will address them first thing on Monday, most of them seem reasonable.

R/init.R
vedhav: Sure

R/init.R
vedhav: Yes, we plan to have these changes for the Feb release. Constant changes to the main `init` function is not great so we would like to make all the deprecations as soon as we can.

R/init.R
vedhav: Sure

R/init.R
vedhav: Agreed

NEWS.md
vedhav: 🤦🏽

NAMESPACE
```diff
@@ -23,19 +24,24 @@ export(get_metadata)
...
+export(modify_footer)
+export(modify_header)
+export(modify_title)
 export(module)
 export(modules)
 export(new_tdata)
 export(report_card_template)
 export(reporter_previewer_module)
 export(show_rcode_modal)
+export(srv_session_info)
```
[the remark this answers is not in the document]
vedhav: Yes, we need this because when the user chooses to create teal app as a shiny module using `ui_teal` and `srv_teal` they can embed multiple "teal shiny modules" in their giant shiny app, it does not make sense to have session info in multiple places, so we give the user flexibility to call this to their preferred location in their shiny app.

R/TealAppDriver.R
vedhav: I'm having second thoughts about using a static logo from the package path. In order to achieve this we need to do these two things:
1. Add the favicon to the inst, not a big deal as we can reduce its size.
2. Serve this file as static content ([addResourcePath](https://shiny.posit.co/r/reference/shiny/1.0.1/addresourcepath.html)) so it can be accessed by shiny.
It is the second point where I have the trouble, I do not feel that serving static content just for a favicon is a good idea.
I think it's safe to use this remote URL as it is controlled by us.

gogonzo: I'm not a fan of addResourcePath for favicon neither. I think URL is the best solution

pawelru: I suggest to revisit this once again as a separate task. We already have lived with urls for a while so it's not a big issue. It's just an idea for performance enhancement - not a bug per se.

Issue #1438, opened 2024-12-20
Assess `id` argument in `init()`


donyunardi:
### Summary  
We need to evaluate whether the `id` argument is still necessary given recent improvements to make teal a modular framework:  
- https://github.com/insightsengineering/teal/issues/669
- https://github.com/insightsengineering/nestdevs-tasks/issues/62

As noted in the documentation, the `id` argument is currently a legacy feature:  
https://github.com/insightsengineering/teal/blob/997da60fd983200f98e4dcaf12fa768e53291a1a/R/init.R#L31-L33

### Acceptance Criteria  
- Assess whether the `id` argument should be retained or removed.  
- If removal is needed, initiate the soft-deprecation process.  
- Update the documentation (roxygen and vignette) to reflect any changes.

gogonzo:
`id` argument is not needed anymore since introduction of `ui/srv_teal` in #669. Instead of `id` in `teal::init` one can call modules directly instead of indirectly set `id`.

m7pr:
CC @donyunardi

PR #1442, opened 2025-01-13
deprecate id from init

gogonzo:
- deprecate id in init #1438  
- fixing (wrong) id documentation in inherited docs. Please notice that methods were inheriting `id` documentation from init. It means that all modules had a wrong description.

### DIFF
--- a/R/init.R
+++ b/R/init.R
@@ -27,9 +27,11 @@
 #'   the header of the app.
 #' @param footer (`shiny.tag` or `character(1)`) Optionally,
 #'   the footer of the app.
-#' @param id (`character`) Optionally,
+#' @param id `r lifecycle::badge("deprecated")` (`character`) Optionally,
 #'   a string specifying the `shiny` module id in cases it is used as a `shiny` module
-#'   rather than a standalone `shiny` app. This is a legacy feature.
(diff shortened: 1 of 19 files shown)


R/init.R
```diff
@@ -221,6 +222,20 @@ init <- function(data,
...
   }
 
+
+  if (lifecycle::is_present(id)) {
+    lifecycle::deprecate_soft(
+      when = "0.15.3",
+      what = "init(id)",
+      details = paste(
+        "To wrap `teal` application within other shiny application please use",
+        "`ui_teal()` and `srv_teal()` and call them as normal shiny modules"
```
m7pr, line 232:
```suggestion
        "`ui_teal()` and `srv_teal()` and call them as regular shiny modules."
```

m7pr:
@gogonzo is this the proper moment to update tests, or those should be adjusted once `id` gets completely removed?

Right now, `shiny::testServer` still uses `id` parameter in many places.

```r
        shiny::testServer(
          app = srv_teal,
          args = list(
            id = "test",
            data = teal.data::teal_data(iris = iris),
            modules = modules(example_module())
          ),
```

m7pr (approved):

PR #1445, opened 2025-01-15
Fix CI pipelines

m7pr:
Part of https://github.com/insightsengineering/nestdevs-tasks/issues/98

### DIFF
--- a/.github/workflows/scheduled.yaml
+++ b/.github/workflows/scheduled.yaml
@@ -34,6 +34,8 @@ jobs:
     secrets:
       REPO_GITHUB_TOKEN: ${{ secrets.REPO_GITHUB_TOKEN }}
       GCHAT_WEBHOOK: ${{ secrets.GCHAT_WEBHOOK }}
+    extra-deps: |
+      SummarizedExperiment (>= 1.36.0)
     with:
       strategy: ${{ matrix.test-strategy }}
       additional-env-vars: |
--- a/.github/workflows/scheduled.yaml
+++ b/.github/workflows/scheduled.yaml
@@ -34,12 +34,12 @@ jobs:
     secrets:
       REPO_GITHUB_TOKEN: ${{ secrets.REPO_GITHUB_TOKEN }}
       GCHAT_WEBHOOK: ${{ secrets.GCHAT_WEBHOOK }}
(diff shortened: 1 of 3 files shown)


m7pr:
Started a job in here https://github.com/insightsengineering/teal/actions/runs/12784552889

m7pr:
Started a new job after recent changes https://github.com/insightsengineering/teal/actions/runs/12784947499

m7pr:
one more run https://github.com/insightsengineering/teal/actions/runs/12785524755

m7pr:
@pawelru an interesting case for you.
`teal` depends on `MultiAssayExperiment` in Suggests, but during `min_isolated` build
https://github.com/insightsengineering/teal/actions/runs/12784947499/job/35639440420#step:9:710 you can see that it tries to load very old version of `SummarizedExperiment` which is a dependency of `MultiAssayExperiment`. This old version depends on old `MatrixGenerics` that depends on `matrixStats` that can not be installed.

However, newest version of `SummarizedExperiment` depends on newer `MatrixGenerics` that doesn't have installation issues and all the build is satisfied https://github.com/insightsengineering/teal/actions/runs/12785524755 but I needed to add `SummarizedExperiment` experiment to package Suggests just to satisfy the build, even though this package is not directly needed for `teal`.

Do you have any advice on how we should proceed? Let the `min_isolated` fail, allow `SummarizedExperiment` to be added to Suggests, or to some Enhances or Linking to fields?

pawelru:
Please have a look at the [`extra-deps` workflow parameter](https://github.com/insightsengineering/r-verdepcheck-action/blob/55a041094b5da3305dcf11ab90575d0eb7bd3282/action.yaml#L15-L18). It was designed to handle such cases and allow to provide specification of indirect dependencies not to pollute DESCIRPTION file as you mentioned

m7pr:
Amazing, thanks for the hint :)!

llrs-roche (approved):
The extra-deps fixed the min_isolate check: [before](https://github.com/insightsengineering/teal/actions/runs/12784947499/job/35639440420) vs [after](https://github.com/insightsengineering/teal/actions/runs/12789723140/job/35653827894). 
The other dependencies fixed the rest.

PR #1449, opened 2025-01-17
1441 change default value of decorators parameter in `example_module`

m7pr:
The same changes as in https://github.com/insightsengineering/teal.modules.general/pull/826
We are unifying `decorators` and `transformators` default values for modules


R/dummy_functions.R
```diff
@@ -9,8 +9,8 @@
 #' to read more about decorators.
 #'
 #' @inheritParams teal_modules
-#' @param decorators `r lifecycle::badge("experimental")` (`list` of `teal_transform_module` or `NULL`) optional,
-#' if not `NULL`, decorator for tables or plots included in the module.
+#' @param decorators `r lifecycle::badge("experimental")` (`list` of `teal_transform_module`) optional,
+#' decorator for `object` included in the module.
```
llrs-roche, line 13:
```suggestion
#' decorators for `object`s included in the module.
```

llrs-roche (commented):
Looks good, just one minor comment, the current documentation is specific for this example module. Could it be made a bit more general? See suggestion (not sure how if it would render well)

R/dummy_functions.R
m7pr: @llrs-roche no. there is only one object called `object`.

> #' This module creates an object called `object` that can be modified with decorators.
#' The `object` is determined by what's selected in `Choose a dataset` input in UI.
#' The object can be anything that can be handled by `renderPrint()`.
#' See the `vignette("decorate-modules-output", package = "teal")` or [`teal_transform_module`]
#' to read more about decorators.

llrs-roche: I understand that on this example there is only one object named `object`, but on the previous documentation it was documented generally: "tables or plots" that is something that is lost with this PR.

m7pr: Yes, but this is outdated. `plots` and `tables` were copied from some generic place. Unsure where it came from. I actually think it's better to focus on a specific object in this case.

llrs-roche (approved):

PR #1451, opened 2025-01-20
runApp can't work due to as.shinyapp.obj dispatch failure

gogonzo:
Run this to see what happens

```r
library(teal.modules.general)

app <- teal::init(
  data = teal_data() |> within(iris <- iris),
  modules = modules(
    tm_missing_data()
  )
)

runApp(app)
```


vedhav (approved):
👍🏽

PR #1456, opened 2025-01-20
Cleanup after merge

vedhav:
Addresses the minor comments after the #1440 merge


R/init.R
```diff
@@ -247,41 +247,38 @@ init <- function(data,
...
     res <- modify_title(res, title)
   }
   if (lifecycle::is_present(header)) {
-    checkmate::assert_multi_class(header, c("shiny.tag", "shiny.tag.list", "html", "character"))
     lifecycle::deprecate_warn(
       when = "0.15.3",
       what = "init(header)",
       details = paste(
         "Use `modify_header()` on the teal app object instead."
       )
```
llrs-roche, line 264: I know this was not modified, but there is no need to use `paste`. In addition if you could use `with` it will provide a standard rlang message.

R/init.R
```diff
@@ -247,41 +247,38 @@ init <- function(data,
...
+    checkmate::assert_multi_class(header, c("shiny.tag", "shiny.tag.list", "html", "character"))
     res <- modify_header(res, header)
   }
   if (lifecycle::is_present(footer)) {
-    checkmate::assert_multi_class(footer, c("shiny.tag", "shiny.tag.list", "html", "character"))
     lifecycle::deprecate_warn(
       when = "0.15.3",
       what = "init(footer)",
       details = paste(
         "Use `modify_footer()` on the teal app object instead."
```
llrs-roche, line 274: Similar to the previous comment, use `with = ` to signal which is the replacement.

llrs-roche (approved):
Changes look good to me: first raising the deprecation and then checking the input. I made some small suggestions to use `deprecate_*(with = )` as I have this fresh from working on other PR, Perhaps you clean some (or all) of them of teal.

PR #1467, opened 2025-01-29
Vignette: shiny as a module

### LINKED ISSUE #1450: Create Vignette for teal as a module
With the completion of #669, let's write a vignette to showcase to show teal app as a module.

gogonzo:
closes #1450


vignettes/teal-as-a-module.Rmd
```diff
@@ -0,0 +1,57 @@
...
+srv_app <- function(input, output, session) {
+  data_subset <- reactive(data[input$datasets])
+  srv_teal("teal", data = data_subset, modules = mods)
+  srv_session_info("session_info")
+}
+
+if (interactive()) {
+  shinyApp(ui_app, srv_app)
+}
+```
```
llrs-roche, line 57: Perhaps a final comment on the example would be good. Like describing the differences with a pure teal app from the final end user (or for the app developer: would it work without `teal_data()` ? )

llrs-roche (commented):
Looks good some minor comments about wording. Feel free to ignore

vignettes/teal-as-a-module.Rmd
```diff
@@ -0,0 +1,57 @@
...
+    toc: true
+vignette: >
+  %\VignetteIndexEntry{teal as a module}
+  %\VignetteEngine{knitr::rmarkdown}
+  %\VignetteEncoding{UTF-8}
+---
+
+# Introduction
+
+Shiny developer who is interested in embedding Teal application into its own app, can use the Teal module composed of `ui_teal()` and `srv_teal()` functions. Unlike `init()`, this module will not automatically include session info
```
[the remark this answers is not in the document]
donyunardi: Also, can we make sure it's `teal` instead of Teal?

vignettes/teal-as-a-module.Rmd
```diff
@@ -0,0 +1,57 @@
...
+footer, but it is possible to add it manually with `ui_session_info()` and `srv_session_info()`.
+Team as module offers several advantages such as:
+
+- Including one or multiple teal application in other app.
+- Run teal applications based on the dynamically created components like initial data, modules, filters.
+
+# Example
+
+Example below shows how to use the Teal module in Shiny app. In the example `srv_teal()` is called with the dynamic
+data created in the server of the parent app.
```
donyunardi, line 25:
```suggestion
Example below demonstrates how to embed `teal` as a module in a Shiny application. 

Here, `srv_teal()` is called using the dynamic data generated within the server of the parent app.
```

vignettes/teal-as-a-module.Rmd
```diff
@@ -0,0 +1,57 @@
...
+
+```{r setup}
```
donyunardi, line 27:
```suggestion
```{r setup, message = FALSE}
```

We got message in the vignettes when the library is loaded:
image

I want to hide these messages.

llrs-roche: I think it is better to use: `suppressPackageStartupMessages(library("teal"))` as it makes it evident to the user that there will be some messages on loading the library (and are not caught off guard if the messages appear when on the vignette there are none).

vignettes/teal-as-a-module.Rmd
gogonzo: > describing the differences with a pure teal app from the final end user 

What do you mean? Isn't "_Unlike `init()`, this module will not automatically include session info footer_" enough?

> would it work without `teal_data()`

No it wouldn't. But what do you think I should add? `data`, `modules` and `filter` arguments are well documented in `module_teal`.

vignettes/teal-as-a-module.Rmd
gogonzo: I don't have an opinion about what is better, but if you believe so I probably should change all `library(teal)` in the package's vignettes.

vignettes/teal-as-a-module.Rmd
llrs-roche: > What do you mean? Isn't "_Unlike `init()`, this module will not automatically include session info footer_" enough?

For example, when I created this app, the left panel was not there,  this is evident for us but it might not be so for new users or app developers that might not know how it works. I usually struggle with vignettes that don't lower the level and assume the user already knows the package. 

> 
> > would it work without `teal_data()`
> 
> No it wouldn't. But what do you think I should add? `data`, `modules` and `filter` arguments are well documented in `module_teal`.

The idea behind vignettes in my opinion are to serve as introduction to the user. If this is documented on `module_teal` we can mention this to redirect the users to the detailed documentation of the function or to other vignettes. 

My personal preference is to provide extra guidance than assume the user already knows something and not following through the vignette. But it is hard to set the level required for a vignette. Please feel free to resolve this conversation

vignettes/teal-as-a-module.Rmd
llrs-roche: I mean for me it is not super critica. On the other side sometimes people is surprised when a certain function appears on their namespace or it is masked by a new loaded package.

vignettes/teal-as-a-module.Rmd
gogonzo: > For example, when I created this app, the left panel was not there, this is evident for us but it might not be so for new users or app developers that might not know how it works

I don't understand why left panel would be missing. Maybe setup problems?


> The idea behind vignettes in my opinion are to serve as introduction to the user. If this is documented on module_teal we can mention this to redirect the users to the detailed documentation of the function or to other vignettes.

I have made a small change clarifying what kind of object is created where - "Here, `srv_teal()` is called using the reactive `teal_data` object passed from the server of the parent app."

Vignettes in `teal` are organised in a way that first in "Getting started with teal" which contains links to subtopics. There are topics about `modules`, `data`, `filters` and "Embedding teal in shiny application" which starts with "Advanced users...".

For me it is clear and I don't see what can we add here but if you think that we need to extra clarify please make a suggestion.

vignettes/getting-started-with-teal.Rmd
```diff
@@ -122,6 +122,10 @@ Note that `teal` does not display the code, that is the modules' responsibility.
 For example, the `example_module` function used above shows the code in the main panel together with other outputs.
 For more details see [this vignette](including-data-in-teal-applications.html).
 
+### Embedding teal in shiny application
+
+Advanced shiny users can include teal application in their Shiny application. For further details see [teal as a module](teal-as-a-module.html).
```
llrs-roche, line 127: Usually vignettes are linked with `vignette("title")`, I'm not sure if this approach is better or not. But it seems like a valid approach too: https://blog.r-hub.io/2020/06/03/vignettes/#cross-references

vignettes/teal-as-a-module.Rmd
llrs-roche: The visual difference I saw was due to having a narrow window for the app. When I resized it appeared on my expected location. 

Here is a suggestion, feel free to modify the wording or not add it:

On this application the selection of the input datasets are done outside the teal module.
You can see that when a dataset is added by shiny the Teal modules show it, but not before.

llrs-roche (approved):
Looks good just two minor comments. 

I was also thinking if the vignette could use shinylive to show the apps or not. It is not in Suggests and I'm not sure if is added in other packages' vignettes. But currently is short and to the point.

vignettes/getting-started-with-teal.Rmd
gogonzo: Using `vignette("title")` is the only possible option when you link roxy docs because they are in .Rd file. Vignettes are rendered to html on CRAN and pkgdown, so it is better to use html-relative-links

PR #1468, opened 2025-01-29
Move the reporter module creation to `module_teal`

vedhav:
Part of #1450 as it provides better API for people who consume teal as a shiny module.


gogonzo (approved):
👍

PR #1469, opened 2025-01-29
fix error

gogonzo:
closes insightsengineering/teal.reporter#302
`TealReporterCard` inherits from `ReporterCard`, when loading from file `TealReporterCard$dispatch_block(block_class)` is triggered. Because `TealReporterCard` is in `teal` namespace, `TextBlock` is failing as `TextBlock` is a `teal.reporter` (private) object. 

I hope code fix is self explanatory.

### DIFF
--- a/R/teal_reporter.R
+++ b/R/teal_reporter.R
@@ -74,7 +74,11 @@ TealReportCard <- R6::R6Class( # nolint: object_name.
   ),
   private = list(
     dispatch_block = function(block_class) {
-      eval(str2lang(block_class))
+      if (exists(block_class, getNamespace("teal"))) {
+        get(block_class)
+      } else {
+        super$dispatch_block(block_class)
+      }
     }
   )
 )
--- a/R/teal_reporter.R
+++ b/R/teal_reporter.R
@@ -75,8 +75,10 @@ TealReportCard <- R6::R6Class( # nolint: object_name.
   private = list(
     dispatch_block = function(block_class) {
       if (exists(block_class, getNamespace("teal"))) {
+        # for block classes which are in teal (TealSlicesBlock)
         get(block_class)
       } else {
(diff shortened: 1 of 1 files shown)


R/teal_reporter.R
```diff
@@ -74,7 +74,11 @@ TealReportCard <- R6::R6Class( # nolint: object_name.
   ),
   private = list(
     dispatch_block = function(block_class) {
-      eval(str2lang(block_class))
+      if (exists(block_class, getNamespace("teal"))) {
```
llrs-roche, line 77: Perhaps the code is evident now, but I would leave a comment why this is needed. But feel free to ignore this

llrs-roche (approved):
I can confirm I see an error on the bottom right corner when uploading the report (even if generated with this branch)  on main branch:

```
Loading reporter error!
Please contact app developer
```

When I upload the report on this branch I do not longer see this error.

I wonder if a way to test this. This function seems important and that it had slipped through checks doesn't seem good.

R/teal_reporter.R
gogonzo: @llrs-roche I'll try to write the tests tomorrow. It might take too much effort in tests to cover this particular case (downloading and uploading zip reporter file)

llrs-roche: No problem, I understand this is more an integration test than a unit test. Those are always hard to write and run. Thanks!

gogonzo: I skip writing unit test for download/loading a report. In both cases test will be problematic:
1. Add a reporter card > go to reporter previewer module > click download > reset reporter > upload file > check if the reporter card is set
2. Creating a file with the reporter before and put it in testthat directory. go to reporter previewer > upload 

In (1) it will take a lot of code to automatize this, in the (2) we will be dependent on a file created in this particular app. Any changes in the app or framework will require to recreate a file.

PR #1470, opened 2025-01-30
Review decorator vignette

donyunardi:
Part of https://github.com/insightsengineering/nestdevs-tasks/issues/99


vignettes/decorate-module-output.Rmd
```diff
@@ -210,12 +214,11 @@ failing_decorator <- teal_transform_module(
 
 ### Example Module
 
-To include decorators in a `teal` module, pass them as arguments (`ui_args` and `server_args`) to the module’s `ui` and 
-`server` components, where they will be used by `ui/srv_teal_transform_module`. 
+To include decorators in a `teal` module, pass them as arguments (`ui_args` and `server_args`) to the module’s `ui` and
+`server` components, where they will be used by `ui_transform_teal_data` and `srv_transform_teal_data`.
```
donyunardi, line 218: @gogonzo am I using the right function name?

vignettes/decorate-module-output.Rmd
```diff
@@ -510,7 +515,7 @@ interactive_decorator_3 <- teal_transform_module(
 
 ### Application
 
-As you might have noted, the x axis title from the first decorator will be used but won't show up on the resulting plot:
+As you might have noticed, the x-axis title from the first decorator will be used but will not appear in the resulting plot:
```
donyunardi, line 518: @gogonzo 
Can you double check the app for this section? From what I can see, only `decorator_1` is functional.
No changes to the plot's x or y axis when I toggled to `decorator_2` and `decorator_3`. 

Is this as expected?

vignettes/decorate-module-output.Rmd
gogonzo: Yes. Closing a conversation

vignettes/decorate-module-output.Rmd
gogonzo: Yes, third decorator doesn't make sense.

I think this vignette is too long. There are too many examples doing every possible thing. I think that instead of having a section with multiple decorators there should be a tab in the first example with two decorators. Also the last example with multiple outputs - Yes, it is a real case but come on, let's give app developers space for creativity and let them utilise decorators in their way.

donyunardi: >  I think that instead of having a section with multiple decorators there should be a tab in the first example with two decorators.

I like this idea.

> Also the last example with multiple outputs - Yes, it is a real case but come on, let's give app developers space for creativity and let them utilise decorators in their way.

Now that you pointed out, it does feel repetitive with the section `Decorating Plots`. Maybe the section `Decorating Plots` should be named something that's more generic that can easily be applied to either plot, table, or whatever the final output of the module.

Let me try to rewrite this.

donyunardi: Update in e97a7da97f3c577acb7b925b284637a75b20be22

vignettes/decorate-module-output.Rmd
```diff
@@ -12,16 +12,21 @@ vignette: >
 
 ## Introduction
 
-`teal` is a powerful `shiny`-based framework with built-in modules for interactive data analysis. 
-This document outlines the customization options available for modifying the output of `teal` modules. 
-You will learn how to use `teal_transform_module()` to modify and enhance the objects created by `teal::modules()`, 
+`teal` is a powerful `shiny`-based framework with built-in modules for interactive data analysis.
+This document outlines the customization options available for modifying the output of `teal` modules.
```
kumamiao, line 16:
```suggestion
Yet when these built-in module outputs don’t fully meet your needs, you have the flexibility to customize the output. This document outlines the customization options available for modifying the output of `teal` modules.
```

vignettes/decorate-module-output.Rmd
```diff
@@ -12,16 +12,21 @@ vignette: >
...
 
 ## Decorators
 
-In programming, **decoration** refers to the process of modifying an object while preserving its original class. For instance, given an object `x` of class `"my_class"`, a function `foo(x)` is considered a **decorator function** if it modifies `x` and returns an object that retains the same class. In this context, `x` is referred to as the **decorated object**, and `foo()` is the **decorator function** or **decorator**. Decorators can perform a variety of operations, such as adding new methods or modifying data, while ensuring the object remains compatible with its original usage.
+In programming, **decoration** refers to the process of modifying an object while preserving its original class.
+For instance, given an object `x` of class `"my_class"`, a function `foo(x)` is considered a **decorator function** if it modifies `x` and returns an object that retains the same class.
+In this context, `x` is referred to as the **decorated object**, and `foo()` is the **decorator function** or **decorator**. Decorators can perform a variety of operations, such as adding new methods or modifying data, while ensuring the object remains compatible with its original usage.
```
kumamiao, line 25:
```suggestion
In this context, `x` is referred to as the **object to be decorated**, and `foo()` is the **decorator function** or **decorator**. Decorators can perform a variety of operations, such as adding new methods or modifying data, while ensuring the object remains compatible with its original usage.
```

"decorated object" sounds like it's the result of `foo(x)`

vignettes/decorate-module-output.Rmd
```diff
@@ -43,22 +48,23 @@ It is recommended to review the module documentation or source code to understan
```
kumamiao, line 48: On the first point under Requirements and Limitations, should we be more explicit on how a user could check whether a module is designed to work with decorators or not?

vignettes/decorate-module-output.Rmd
```diff
@@ -96,8 +102,8 @@ static_decorator_lang <- teal_transform_module(
 
 ### UI
```
kumamiao, line 103: maybe UI+Server? UI makes the bullets a bit confusing, like we're talking about server and then UI

vignettes/decorate-module-output.Rmd
```diff
@@ -127,7 +133,7 @@ interactive_decorator <- teal_transform_module(
 )
 ```
 
-As in the earlier examples, `make_teal_transform_server()` can simplify the creation of the server component. 
+As in the earlier examples, `make_teal_transform_server()` can simplify the creation of the server component.
 This wrapper requires you to use `input` object names directly in the expression - note that we have `xlab(x_axis_table)` and not `my_title = input$x_axis_title` together with `xlab(my_title)`.
```
kumamiao, line 137:
```suggestion
This wrapper requires you to use `input` object names directly in the expression - note that we have `xlab(x_axis_title)` and not `my_title = input$x_axis_title` together with `xlab(my_title)`.
```
I'm a bit confused here, the explanation seems to be talking about `x_axis_title` and code for `y_axis_title`

vignettes/decorate-module-output.Rmd
```diff
@@ -12,16 +12,21 @@ vignette: >
 
 ## Introduction
 
-`teal` is a powerful `shiny`-based framework with built-in modules for interactive data analysis. 
-This document outlines the customization options available for modifying the output of `teal` modules. 
-You will learn how to use `teal_transform_module()` to modify and enhance the objects created by `teal::modules()`, 
+`teal` is a powerful `shiny`-based framework with built-in modules for interactive data analysis.
```
gogonzo, line 15: I'd skip this line at all. This is decorator vignette and everyone should already know what is `teal`. Also "powerful" is a bragging which shouldn't really have a place in the vignette. (this is just my opinion)

vignettes/decorate-module-output.Rmd
gogonzo: I think that the while paragraph has no meaningful value. Referring to `decoration` in programming is patronizing. I think word "decorate" is self exploratory I could even write that "Decoration is a process of changing the look of the object without altering its fundamental function". I would simply skip this section and leave the line below "In the context of `teal` applications, decoration...". WDYT?

vignettes/customizing-module-output.Rmd
```diff
@@ -0,0 +1,327 @@
...
+    toc: true
+vignette: >
+  %\VignetteIndexEntry{Customizing Module Output}
+  %\VignetteEngine{knitr::rmarkdown}
+  %\VignetteEncoding{UTF-8}
+---
+
+## Introduction
+
+`teal` is a powerful `shiny`-based framework with built-in modules for interactive data analysis.
```
gogonzo, line 15: I'd skip this line at all. This is decorator vignette and everyone should already know what is teal. Also "powerful" is a bragging which shouldn't really have a place in the vignette. (this is just my opinion)

```suggestion
```

vignettes/decorate-module-output.Rmd
gogonzo: I've added a link to "include-decorators-in-a-teal-module" section

vignettes/decorate-module-output.Rmd
gogonzo: I changed section titles to "Non-interactive Decorators" and "Interactive Decorators"

vignettes/customizing-module-output.Rmd
donyunardi: Agree.

vignettes/decorate-module-output.Rmd
kumamiao: makes sense to me.

donyunardi: Done in af32a06b5990e4f6cbd18680cd088697781ac141

vignettes/customizing-module-output.Rmd
```diff
@@ -0,0 +1,326 @@
...
+You will learn how to use `teal_transform_module()` to modify and enhance the objects created by `teal::modules()`,
+allowing you to tailor the outputs without rewriting the original module code.
+
+## Decorators
+
+In `teal`, decorators are used to modify module outputs, such as plots or tables, without changing their core structure.
+
+For example, imagine you have a plot and you want to add a title.
+A decorator function, like `add_title(x, "My Plot")`, takes an existing plot and updates it by adding a title while keeping everything else the same.
+This makes it different from a function that creates a brand-new plot from scratch, as decorators work by enhancing what’s already there.
```
gogonzo, line 29:
```suggestion
For example, imagine you have a `plot` object (ggplot) and you want to add a title. To do so one simply needs to call `plot <- plot + ggtitle("My Plot")`. It is important to enhance the current object object instead of creating a new one. We call this operation a decoration.
```

vignettes/customizing-module-output.Rmd
```diff
@@ -0,0 +1,326 @@
...
+
+The decorators process can vary in complexity:
+
+- **Simple Decorators**: Single-step modifications, such as a single method call that does not require additional data.
+- **Complex Decorators**: Multi-step operations that may involve interdependent transformations, potentially requiring input from dedicated `shiny` UI elements.
+
+This functionality empowers application developers to significantly customize outputs beyond the default capabilities provided by existing module parameters.
+Decorators allow for advanced modifications, enabling highly tailored and dynamic user experiences in `teal` applications.
```
gogonzo, line 37: I don't understand what these two sentences means. They don't help app developer grasp what is decoration. Besides these statements doesn't have to be true because:

> This functionality empowers application developers to significantly customize outputs beyond the default capabilities provided by existing module parameters

Signification is debatable. For example one can't change object type from plot to for example table. Also decorator is "existing module parameter" so it doesn't go beyond "existing module parameters" - it goes beyond default output appearance.

vignettes/customizing-module-output.Rmd
```diff
@@ -0,0 +1,326 @@
...
+
+## Requirements and Limitations
+
+To use decorators effectively, certain requirements must be met:
+
+1. **Module Support**: While `teal` provides the core functionality for decorators, the module must explicitly support this functionality. Developers should ensure that the module has been designed to work with decorators (see [Include Decorators in a `teal` Module](#include-decorators-in-a-teal-module)).
+2. **Matching Object Names**: Decorators must reference object names that align with the internal naming conventions of the module. Each module may use different names for its output objects, such as `plot` or `table`. This alignment is critical for successful decorator.
```
gogonzo, line 44:
```suggestion
2. **Matching Object Names**: Decorators must reference object names that align with the internal naming conventions of the module. Each module may use different names for its output objects, such as `plot` or `table`. This alignment is critical for successful decorator.
3. **Matching Object Class**: Decorated objects are used in certain context and they are use to be consumed by relevant `render*` functions. For example `ggplot` objects are used in `renderPlot` while `plotly` objects are consumed by `renderPlotly`. It is important that the object doesn't change its basic class so the module can deal with the modified object.     
```

vignettes/customizing-module-output.Rmd
```diff
@@ -0,0 +1,326 @@
...
+
+It is recommended to review the module documentation or source code to understand its internal object naming before applying decorators.
```
gogonzo, line 46:
```suggestion
It is recommended to review the module documentation or source code to understand its internal object naming and object class before applying decorators.
```

vignettes/customizing-module-output.Rmd
donyunardi: These two sentences feel like closing statements to me. 

However, since we already have this sentence at the beginning of the section:
```
In `teal`, decorators are used to modify module outputs, such as plots or tables, without changing their core structure.
```  

I think we can remove these two sentences without losing clarity.

WDYT?

gogonzo: Yes, I think also to remove.

gogonzo (approved):
Approving and merging this one in order to proceed to https://github.com/insightsengineering/teal/pull/1476

PR #1472, opened 2025-01-31
Add roxy.shinylive

### LINKED ISSUE #1404: [idea] use Shinylive in pkg docs
This is an idea that I would let you refine it better. 

I would love to see the documentation (README, vignettes, function reference, etc.) more _interactive_ using Shinylive.

Recently we created a small tool for that purpose `roxy.shinylive`. Feel free to use it. It's already in the modules packages and I think it's robust enough to be implemented in the _core_ packages.

vedhav:
Closes #1404 

Note that the shinylive R apps will be broken till the `teal` is released on CRAN because we are unable to use the dev version of the packages from pharmaverse.r-universe because the webr binaries build there are only compatible with the [webr REPL](https://webr.r-wasm.org/latest/) and not the [shinylive](https://shinylive.io/r/examples/) yet.


gogonzo (approved):
All shiny examples in the docs do have `@examplesShinylive` section

PR #1490, opened 2025-02-12
Use shinylive to show the app on the README

### LINKED ISSUE #1404: [idea] use Shinylive in pkg docs

llrs-roche:
# Pull Request

Fixes #1404 

Converted the README.md to README.Rmd to be able to genereate a shinylive of the example app.
The space modifications are automatically but I just made very few modifications to the file: add some code chunk options and two chunks at the end. 
It adds some steps on the docs.yaml to render the README (iff modified).

Note: I haven't removed the preview yet (I want to see how it works and then I'll remove it).

There is [a comment](https://github.com/insightsengineering/teal/issues/1404#issuecomment-2650872198) about having a git hook locally.  `usethis::use_readme_rmd()` generates a hook to be added to `.git/hooks/pre-commit` (not sure if this could be shared via .pre-commit-config.yaml). 

<details>
<summary>Hook</summary>

```
#!/bin/bash
README=($(git diff --cached --name-only | grep -Ei '^README\.[R]?md$'))
MSG="use 'git commit --no-verify' to override this check"

if [[ ${#README[@]} == 0 ]]; then
  exit 0
fi

if [[ README.Rmd -nt README.md ]]; then
  echo -e "README.md is out of date; please re-knit README.Rmd\n$MSG"
  exit 1
elif [[ ${#README[@]} -lt 2 ]]; then
  echo -e "README.Rmd and README.md should be both staged\n$MSG"
  exit 1
fi

```

</details>


llrs-roche:
From `pkgdown::build_home()`:

> If you use index.Rmd or README.Rmd it's your responsibility to knit the document to create the corresponding .md. pkgdown does not do this for you because it only touches files in the ⁠doc/⁠ directory.

If we want to include the app we'll need a different README for the repo/github view and another for github pages. This would require tweaking the  action on insightsengineering/r.pkg.template/.github/workflows/pkgdown.yaml@main to pick the right README.md as it currently pulls the repository from a branch (or that a commit is made directly to that branch). This becomes quite complicated and wouldn't be ready for the release tomorrow. 

Adding a link it is fairly simple and portable: I'll keep the video recorded and just the link to where users can visualize the output.

m7pr:
> There is https://github.com/insightsengineering/teal/issues/1404#issuecomment-2650872198 about having a git hook locally. usethis::use_readme_rmd() generates a hook to be added to .git/hooks/pre-commit (not sure if this could be shared via .pre-commit-config.yaml).

Our .pre-commit-confit.yaml is based on https://github.com/lorenzwalthert/precommit which has this hook already included:
https://github.com/lorenzwalthert/precommit/blob/main/.pre-commit-hooks.yaml#L59

You just need to extend `.pre-commit-confit.yaml` with 

```r
      - id: readme-rmd-rendered
        name: Render README.md
```

.github/workflows/docs.yaml
```diff
@@ -34,8 +34,38 @@ on:
...
+        run: |
+          RMD_PATH=($(git diff --name-only ${{ github.event.before }} ${{ github.sha }} *.Rmd))
+          Rscript -e 'for (f in commandArgs(TRUE)) if (file.exists(f)) rmarkdown::render(f)' ${RMD_PATH[*]}
+
+      - name: Commit results ⬆️
+        run: |
+          git config --local user.name "$GITHUB_ACTOR"
+          git config --local user.email "$GITHUB_ACTOR@users.noreply.github.com"
+          git commit ${RMD_PATH[*]/.Rmd/.md} -m 'Re-build Rmarkdown files' || echo "No changes to commit"
+          git push origin || echo "No changes to commit"
```
m7pr, line 65: So it's the docs build that has a step that renders the README.Rmd file?

m7pr:
@llrs-roche 

> From pkgdown::build_home():

>> If you use index.Rmd or README.Rmd it's your responsibility to knit the document to create the corresponding .md. pkgdown does not do this for you because it only touches files in the ⁠doc/⁠ directory.

> If we want to include the app we'll need a different README for the repo/github view and another for github pages. 

I don't understand this. The docs action has a step that renders README.Rmd to Readme.md right? This Readme.md is committed to the rspository. And then pkgdown, in the next step, takes the freshly knitted Readme.md. So why we want two README?

.github/workflows/docs.yaml
llrs-roche: Yes, I added it here, as I didn't found it on other actions.

llrs-roche:
Yes, the README.Rmd is rended to README.md and committed to the repository. I've added the job on the docs.yaml to render the README file, but this is only to ensure the file is up to date.

The README.Rmd must be rendered to a README.md so that pkgdown can use it. 
But READMe.md won't have the shinylive_iframe if rendered with `requireNamespace("roxy.shinylive", quietly = TRUE) && knitr::is_html_output() && identical(Sys.getenv("IN_PKGDOWN"), "true")`:
 - It is not an html_output: it is github_document/md
 - It doesn't have the IN_PKGDOWN: Easy to add to the Github Action
 - Even if we removed that and  used just `eval = requireNamespace("roxy.shinylive", quietly = TRUE)` the output format doesn't allow to include the iframes:

    ```
    Quitting from lines 143-144 [shinylive_iframe] (README.Rmd)
    Error in s$close() : attempt to apply non-function
    Error: 
    ! in callr subprocess.
    Caused by error in `s$close()`:
    ! attempt to apply non-function
    ℹ See `$stdout` for standard output.
    Type .Last.error to see the more details.
    ```

If we want to include we would need to render it into different format but not sure how it would play with the CI and the pkgdown internal machinery.

m7pr:
> If we want to include we would need to render it into different format but not sure how it would play with the CI and the pkgdown internal machinery.

Yeah, so the action can render into index.html  and into README.md and I think it should be fine. Action can also remove any unnecessary chunk options that will be problematic or unused in README.md by GitHub when showing the welcome page on the repository.

I would suggest rendering README.Rmd into 2 files: index.html and README.md

vedhav:
Let me raise a small PR to just add the shinylive in the Getting started vignette. Readme can wait, but I think the getting started vignette can go in before the release.

vedhav:
Can any one of you just review this PR #1491?

README.Rmd
```diff
@@ -0,0 +1,170 @@
...
+```{r shinylive_url, echo = FALSE, results = 'asis', eval = requireNamespace("roxy.shinylive", quietly = TRUE)}
+code <- paste0(c(
+  knitr::knit_code$get("teal")
+), collapse = "\n")
+
+url <- roxy.shinylive::create_shinylive_url(code)
+cat(sprintf("[Open in Shinylive](%s)\n\n", url))
+```
+
+```{r shinylive_iframe, echo = FALSE, out.width = '150%', out.extra = 'style = "position: relative; z-index:1"', eval = requireNamespace("roxy.shinylive", quietly = TRUE) && knitr::is_html_output() && identical(Sys.getenv("IN_PKGDOWN"), "true")}
```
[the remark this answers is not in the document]
llrs-roche: Thanks for the feedback! I'm not sure we can insert iframes on markdown format with knitr. See the error message when I simplified the condition: https://github.com/insightsengineering/teal/pull/1490#issuecomment-2653721428. Some sites say iframes are supported while other says it is disabled due to security concerns. But this might be a knitr limitation as I saw some comments about [using some flags on pandoc](https://github.com/jgm/pandoc/issues/2244#issuecomment-115822392) to insert iframes.

In addition, if we want the iframe to show on the pkgdown index.html we need to render it to markdown ourselves as pkgdown won't do it (see this [previous comment](https://github.com/insightsengineering/teal/pull/1490#issuecomment-2653477379)). But it might take time and won't be ready for today's release.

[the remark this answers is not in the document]
pawelru: ok thanks for the reply. I agree - let's not block the release because of that and let's investigate this afterwards

llrs-roche:
The shinylive is not included on the pkgdown file. Test code:

```r
# Build the README.md for repository
devtools::build_readme()

# Build the README.md for pkgdown
Sys.setenv(IN_PKGDOWN = "true")
devtools::build_readme(output_format = "html_document", output_file = "index.html")
knitr::knit("index.html", output = "index.md")

# Check website
unlink("index.html") # Delete just in case it is not overwritten
pkgdown::build_home() # Build index.html file from the index.md file
```

If we follow this instructions we will see that index.md has the iframe code `<iframe src="https://shinylive.io/r/app/` on line 539 (`grep -n "<iframe src=" index.md`). But it won't show up on the index.html generated by pkgdown:

image

I think adding a link to the README and pkgdown home page is the best we can do for now. In addition, the index.md is a 2.7Mb file that is very slow to open (it crashed my computer), not sure if GitHub pages will render that directly (see also [previous comments](https://github.com/insightsengineering/teal/issues/1404#issuecomment-2649740644) suggesting it is not possible to render iframes).

pawelru:
Yeah indeed - looks like iframe is not supported as per [official docs](https://github.github.com/gfm/#example-630:~:text=quot%3B%26gt%3B%3C/p%3E-,6.11,-Disallowed%20Raw%20HTML). 

I have one more idea but a little bit more difficult: a separate `index.Rmd` file that renders into `README.md` (without iframe) and `index.md` (with iframe). Then in pkgdown we say that `index.md` should be used (instead of `README.md`). This way we can achieve iframe in pkgdown at least. WDYT?

llrs-roche:
@pawelru The [script above ](https://github.com/insightsengineering/teal/pull/1490#issuecomment-2665078665) generates an index.md with an iframe. But pkgdown doesn't include the iframe when built locally (see the screenshot). I don't think building the website differently would fix that, I don't see any parameter that could help.

There are some [issues regarding iframes on pkgdown issue tracker](https://github.com/r-lib/pkgdown/issues?q=is%3Aissue%20state%3Aclosed%20iframe), but nothing about shinylive (yet?). In general it is mentioned as a workaround around an issue.

README.Rmd
```diff
@@ -0,0 +1,170 @@
...
+  )
+)
+
+shinyApp(app$ui, app$server)
+```
+
+![App recording](man/figures/readme_app.webp)
+
+
+## Try it out in Shinylive
```
pawelru, line 131:
```suggestion
### Try it out in Shinylive
```

pawelru:
I still think it's doable - please have a look at the `{shinymobile}` - its [readme](https://github.com/RinteRface/shinyMobile/tree/master) and [website](https://shinymobile.rinterface.com/).

I tried to follow your example and I was not able to confirm - I was able to have iframe in the main page of pkgdown website. I did it even more simpler with just one call `devtools::build_readme(output_file = "index.md")` (and simplify the chunk eval condition accordingly).
I had it and then I got some errors about Chrome Headless mode so unfortunately cannot send a screenshot or anything - sorry about that. It looks like I need to do restart but I really don't want it - I have so many important stuff opened

README.Rmd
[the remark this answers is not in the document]
llrs-roche: Working on adding the shinylive on pkgdown to show the updated UI and I could simplify the condition. Both the README and the `pkgdown::build_home_index()` show the app in a iframe. See the updated branch (and resolve the conversation so we can merge if you are happy with the outcome).

.github/workflows/docs.yaml
```diff
@@ -34,8 +34,38 @@ on:
...
+      - name: Checkout repo
+        uses: actions/checkout@v4.1.1
+        with:
+          fetch-depth: 0
+
+      - uses: r-lib/actions/setup-pandoc@v2
+
+      - uses: r-lib/actions/setup-r@v2
+
+      - uses: r-lib/actions/setup-r-dependencies@v2
```
pawelru: should we use `insightsengineering/setup-r-dependencies` instead?

llrs-roche: Good idea, it might avoid some issues. But I'm not sure the actual versions of the packages used on the shinylive. 

```suggestion
      - uses: insightsengineering/setup-r-dependencies
```

.github/workflows/docs.yaml
```diff
@@ -34,8 +34,38 @@ on:
...
+
+      - name: Render Rmarkdown files 🖹
+        run: |
+          RMD_PATH=($(git diff --name-only ${{ github.event.before }} ${{ github.sha }} *.Rmd))
+          Rscript -e 'for (f in commandArgs(TRUE)) if (file.exists(f)) rmarkdown::render(f)' ${RMD_PATH[*]}
+
+      - name: Commit results ⬆️
+        run: |
+          git config --local user.name "$GITHUB_ACTOR"
+          git config --local user.email "$GITHUB_ACTOR@users.noreply.github.com"
```
pawelru: looks like we already have a generic account CI user
https://github.com/search?q=repo%3Ainsightsengineering%2Fr.pkg.template+user.name&type=code

llrs-roche: Removed, but while the automation is nice this is the area I'm less confident of the PR and how it will work in practice. Will the developers need to pull the changes on the README.md? Will there be conflicts? ... Looking forward to the feedback from the review

llrs-roche:
In light of a recent change/notification from this PR: It was decided to keep things simple (as to avoid the Rmd and the GHA to build the README.md) and use a gif, here is the other PR: https://github.com/insightsengineering/teal/pull/1569

PR #1499, opened 2025-02-26
Redesign of `teal.reporter` Cards

m7pr:
Companion to
- [teal.modules.general/report_redesign_poc@main](https://github.com/insightsengineering/teal.modules.general/tree/report_redesign_poc%40main)
- [teal.reporter/redesign@main](https://github.com/insightsengineering/teal.reporter/tree/redesign%40main)

<details><summary> Code examples of dealing with reporter </summary>

```r
devtools::load_all("../teal.reporter")
devtools::load_all(".")

# DISABLE teal.reporter::Reporter globally
app <- init(
  data = within(teal_data(), {iris <- iris}),
  modules = example_module(label = "example teal module"),
  reporter = NULL
)
if (interactive()) {
  shinyApp(app$ui, app$server)
}

# predefined teal.reporter::Reporter with inital card

reporter <- teal.reporter::Reporter$new()
doc1 <- teal.reporter::report_document("## Header 2 text", "Regular text")
reporter$append_cards(setNames(list(doc1), "Welcome card"))

app <- init(
  data = within(teal_data(), {iris <- iris}),
  modules = example_module(label = "example teal module"),
  reporter = reporter
)
if (interactive()) {
  shinyApp(app$ui, app$server)
}

# predefined teal.reporter::Reporter
# that has a template that is always used to prepend all cards

reporter <- teal.reporter::Reporter$new()
template_fun <- function(document) {
  disclaimer <- teal.reporter::report_document("Here comes disclaimer text")
  c(disclaimer, document)
}
reporter$set_template(template_fun)

app <- init(
  data = within(teal_data(), {iris <- iris}),
  modules = example_module(label = "example teal module"),
  reporter = reporter
)
if (interactive()) {
  shinyApp(app$ui, app$server)
}

# remove reporter ADD BUTTON for a single module

app <- init(
  data = within(teal_data(), {iris <- iris}),
  modules = modules(
    example_module(label = "module with reporter"),
    example_module(label = "module without reporter") |> nullify_teal_module_report_card()
    # the same nullify_teal_module_report_card = disable_teal_module_report
    #example_module(label = "module without reporter") |> disable_teal_module_report()
  )
)
if (interactive()) {
  shinyApp(app$ui, app$server)
}

```

</details>

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,8 +1,8 @@
 Type: Package
 Package: teal
 Title: Exploratory Web Apps for Analyzing Clinical Trials Data
-Version: 0.16.0.9006
-Date: 2025-05-07
+Version: 0.16.0.9007
+Date: 2025-06-02
 Authors@R: c(
     person("Dawid", "Kaledkowski", , "dawid.kaledkowski@roche.com", role = c("aut", "cre"),
            comment = c(ORCID = "0000-0001-9533-457X")),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,4 @@
-# teal 0.16.0.9006
+# teal 0.16.0.9007
 
 ### New features
 
--- a/inst/WORDLIST
+++ b/inst/WORDLIST
@@ -4,7 +4,6 @@ Forkers
 Hoffmann
 MAEs
 ORCID
-Reproducibility
 Shinylive
 TLG
 Transformators
@@ -12,8 +11,6 @@ UI
 UX
 args
 bookmarkable
-cloneable
-customizable
 favicon
 favicons
 funder
@@ -21,11 +18,8 @@ lockfile
 modularized
 omics
 pre
-programmatically
 quosure
 reactively
-repo
-reproducibility
 summarization
 tabset
 themer
--- a/vignettes/including-data-in-teal-applications.Rmd
+++ b/vignettes/including-data-in-teal-applications.Rmd
@@ -43,7 +43,7 @@ library(teal)
 data <- teal_data(iris = iris, cars = mtcars)
 ```
 
-Note that `iris` and `cars` have been added to the `datanames` property of `data` (see [`datanames` property](#datanames)).
+Note that `iris` and `cars` have been added to the `datanames` property of `data` (see [`datanames` property](#teal_data-properties)).
 
 This is sufficient to run a `teal` app.
 


DESCRIPTION
```diff
@@ -38,7 +38,7 @@ Depends:
     R (>= 4.1),
     shiny (>= 1.8.1),
     teal.data (>= 0.7.0),
-    teal.slice (>= 0.6.0.9001)
+    teal.slice (>= 0.6.0.9000)
```
m7pr, line 41:
```suggestion
    teal.slice (>= 0.6.0.9001)
```

R/module_teal.R
```diff
@@ -103,13 +105,13 @@ ui_teal <- function(id, modules) {
 
 #' @rdname module_teal
 #' @export
-srv_teal <- function(id, data, modules, filter = teal_slices()) {
+srv_teal <- function(id, data, modules, filter = teal_slices(), reporter = teal.reporter::Reporter$new()) {
```
m7pr, line 108:
```suggestion
srv_teal <- function(id, data, modules, filter = teal_slices(), reporter = teal.reporter::Reporter$new()) {
reporter = reporter$deep_clone()/clone() ?
```

R/module_nested_tabs.R
```diff
@@ -207,7 +207,11 @@ ui_teal_module.teal_module <- function(id, modules, depth = 0L) {
         )
       } else {
         ui_teal
-      }
+      },
+      div( # todo: fix the position to the top right of the tab content?
+        uiOutput(ns("reporter_add_container"))
```
donyunardi, line 212: If the add button is at the bottom, it won’t be visible when the module has a large height, therefore forcing user to scroll down. 

I personally prefer the button to be somewhere at the to so that it's visible when the module loads.

Here are couple options to review it visually:

### Option 1: Inside the sidebar


One more as link instead of button:



The only downside is that the button/link will be hidden if the filter is collapsed.

### Option 2: Inside the teal module


**Cons**: I doesn't look good when using tmc or tmc where we have encoding as a sidebar. 

image

image

### Option 3: At the top right, outside of module
image

@kumamiao what do you think?

m7pr: The add button is still in the development phase. I remember we kindly asked @vedhav for some help in moving the button somewhere more helpful but I guess we had a ton of issues on our plates and the button placement was not the most important for now. But maybe it's time to investigate the position : )

If we go with the text I don't think it should be `Add to Report Card` but rather `Add to Report`

R/modify_teal_module_report.R
```diff
@@ -0,0 +1,178 @@
...
+    server = new_server,
+    ui = new_ui,
+    datanames = teal_module$datanames,
+    server_args = teal_module$server_args,
+    ui_args = teal_module$ui_args,
+    transformators = teal_module$transformators
+  )
+}
+
+modify_teal_module_report_card <- function(teal_module, modify_fun) {
```
donyunardi, line 116: Just to confirm that we will export `modify_teal_module_output` so that user will only need to know `modify_teal_module_report_card` or `nullify_teal_module_report_card`, correct?

R/module_teal.R
```diff
@@ -40,18 +40,20 @@
...
+#' @param reporter (`Reporter`) object used to store report contents. Set to `NULL` to globally disable reporting.
+#' @param disable Whether to disable `reporter` in `ui_teal`. Useful for using `ui_teal` and `srv_teal` as shiny module.
 #'
 #' @return `NULL` invisibly
 NULL
 
 #' @rdname module_teal
 #' @export
-ui_teal <- function(id, modules) {
+ui_teal <- function(id, modules, disable = FALSE) {
```
donyunardi, line 51: Should we consider more descriptive argument name for `disable`?
Maybe like `disable_reporter_ui`?

R/modify_teal_module_report.R
m7pr: I think we will export all 3. `modify_teal_module_output` can be used to build more complex things. other two are convenience wrappers

R/module_teal.R
m7pr: make sense

R/teal_reporter.R
```diff
@@ -194,3 +194,101 @@ TealSlicesBlock <- R6::R6Class( # nolint: object_name_linter.
     teal_slices = NULL # teal_slices
   )
 )
+
+#' Server function for `ReportDocument` card class.
+#' @keywords internal
+add_document_button_srv <- function(id, reporter, r_card_fun) {
```
donyunardi, line 200: As discussed before, in addition to adding a card title, please also include the feature to allow adding an initial comment when adding a card.

R/module_nested_tabs.R
vedhav: I propose that we place this button on the right side to maintain the flow from left to right (Data manipulation -> Outputs -> Report)
Since we have some spare space on the right side due to the standard padding we can utilize this space. We can minimize the button to only show the "+" icon by default and when the button is hovered, we can expand and show the full text. Note that this same behavior should be possible when teal apps are used as shiny modules and multiple teal shiny modules are used in a single app too.
Screenshot 2025-04-24 at 7 45 40 AM
Screenshot 2025-04-24 at 7 46 00 AM
Screenshot 2025-04-24 at 7 54 28 AM

I have the changes on my local, I can commit to this branch if it's okay with you @m7pr :)

R/modify_teal_module_report.R
```diff
@@ -0,0 +1,178 @@
...
+  new_ui <- function(id, ...){
+    ns <- NS(id)
+    original_ui(ns('modified'), ...)
+  }
+
+  new_server <- function(id,
+                         data,
+                         filter_panel_api,
+                         response,
+                         regressor,
```
gogonzo, line 23: new_server should be built dynamically and nothing can be fixed. We can't determine argument names because they are unknown for repos outside of insightsengineering

R/module_teal.R
gogonzo: The only question is where it needs to be cloned. Reason to clone is to avoid the situation when app users modify shared Reporter object. `init` is called before `shiny` session starts reporter object will be the same across sessions. `srv_teal` though is called through `shinyApp()` it means that it has one instance per session.

gogonzo (commented):
- No hardcoded argnames in the module wrapper
- clone Reporter in init
- no `disable` argument in the ui but instead use appendTab in server if reporter used

R/module_nested_tabs.R
m7pr: @donyunardi @kumamiao how do you like the proposition? Looks great for me

R/teal_reporter.R
m7pr: This was also mentioned on teal.reporter PR here https://github.com/insightsengineering/teal.reporter/pull/307#issuecomment-2821530417

Maybe we can go one step further

R/module_teal.R
```diff
@@ -40,18 +40,20 @@
...
   checkmate::assert_character(id, max.len = 1, any.missing = FALSE)
   checkmate::assert_class(modules, "teal_modules")
   ns <- NS(id)
 
-  modules <- append_reporter_module(modules)
+  modules <- append_reporter_module(modules, disable = disable)
```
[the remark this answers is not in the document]
m7pr: Thanks, good idea! Will have a look at this one

R/modify_teal_module_report.R
m7pr: Please look at the `modify_teal_module_output`. `modify_teal_module_output_named_params` is only in the file to make the review process easier - it has a comment:

```r
# This function is just here, so it is easier to understand what's happening in
# modify_teal_module_output.
# We want to have a match in parameter names in new_server and original_server
# assuming we don't know names of parameters in original_server.
# modify_teal_module_output is the final implementation that shows how to do it without knowing parameters.
# modify_teal_module_output_named_params is just here to show how would that look like for a custom and stable
# set of parameter names.
```

but it looks like it only made the things confusing.

I will remove `modify_teal_module_output_named_params`

R/module_teal.R
m7pr: Thanks for checking on this one. I guess we should be cloning only in `init` then

R/module_nested_tabs.R
donyunardi: I think the button position is forced and squeezed to the right padding. We only have small padding on right side of the `teal-body` element so I'm not sure how visible or intuitive this position would be.

I tried increasing the right padding to give some space for the button but it looks odd. 

I also tried adding the padding on the right side but it looks odd due to the border setup since we have a `layout_sidebar` inside a `layout_sidebar`.

image

I do like the idea but want to see if there's a way that we don't have to squeeze the button into that small padding.

kumamiao: Thanks all for the discussion. I like the idea of placing the reporter button at the right side, which follows our flow of data, and is not inside the module-level encoding panel. Although I do hope it would be more visible for the users rather than only expanding when hovering. Agree with Dony on the right padding as well. So I am thinking if something like this may be feasible for consideration (although this may be inter-related with the UI/UX refactor):
Screen Shot 2025-04-25 at 2 51 10 PM

1. The icon will be more intuitive for users than a plus sign. Or we could explore "Add to Report" text option too.
2. For the 3 icons of bookmarking, snapshot manager and filter manager, to be honest I do not think those are being used frequently by users. So I wonder if those can be consolidated into an "dot dot dot", which will only show up when users click on it.

donyunardi: To my understand, the top right icon (snapshot manager, view filter mapping) was more geared toward global function where in this case, the button is specifically for a teal module. 

It might be technically hard to add the "Add to Report" button to that same area.

gogonzo: @kumamiao @donyunardi @m7pr please consider that reporter button won't be the only one added automatically. SRC button can also be removed from the module and added automatically by teal.

donyunardi: Now that we have two possible buttons (SRC and Add to Report) that can be generated automatically, having two buttons at the top right doesn't look that bad for me.

image

What do you think?

m7pr: there is too much white space imho

image

m7pr: Sgould we add the show r code to the scope of the todo list for the refactor? This will require modules to return qenv/teal_data. Here is the list https://github.com/insightsengineering/nestdevs-tasks/issues/104#issuecomment-2830613029

m7pr: for the buttons, maybe they could be inline with the module names, aligned to right
C

m7pr: @vedhav would you be able to set the `Add to Report` button in the discussed place?

R/teal_reporter.R
m7pr: For now I brought back comment feature 4daffe444cb0e56a374e34f788cedff25c43abab

R/modify_teal_module_report.R
```diff
@@ -0,0 +1,178 @@
...
+
+modify_teal_module_report_card <- function(teal_module, modify_fun) {
+  modify_teal_module_output(teal_module, modify_fun, output_name = 'report_card')
+}
+
+nullify_teal_module_report_card <- function(teal_module) {
+  modify_teal_module_output(teal_module, modify_fun = function(report_card) NULL, output_name = 'report_card')
+}
+
+disable_teal_module_report <- nullify_teal_module_report_card
```
gogonzo, line 124: Redundant code and name could be better.
```suggestion
disable_report <- function(teal_module) {
  checkmate::assert_class(teal_module, "teal_module")
  modify_teal_module_output(teal_module, modify_fun = function(report_card) NULL, output_name = 'report_card')
}
```

R/modify_teal_module_report.R
```diff
@@ -0,0 +1,178 @@
...
+    ui = new_ui,
+    datanames = teal_module$datanames,
+    server_args = teal_module$server_args,
+    ui_args = teal_module$ui_args,
+    transformators = teal_module$transformators
+  )
+}
+
+
+modify_teal_module_output <- function(teal_module, modify_fun, output_name) {
```
gogonzo, line 61: I don't like how this function is called. I see why it is implemented like this. 

```r
tm_xyz() |> 
  modify_teal_module_output(
    function(x) {
      x[1:2]
    },
    output_name = "report_card"
  )
```

I think we need to discuss with the rest of the team if `teal_data` can inherit from report_document - then the module could return only one object (`teal_data`) and would need to modify a single object.

gogonzo (commented):
Quick review

R/modify_teal_module_report.R
m7pr: Or as discussed today, and you can edit multiple outputs by passing multiple functions
```r
tm_xyz() |> 
  modify_teal_module_output(
    list(report_card = function(x) {
      x[1:2]
    })
  )
```

R/module_teal.R
m7pr: parameter removed in here https://github.com/insightsengineering/teal/pull/1499/commits/2fad14025beaafc9421f76b7fbdf0ab5f9c79848

R/module_nested_tabs.R
m7pr: Hey @donyunardi @kumamiao @gogonzo 
@vedhav has just commited changes with proposed solution for the placement of the button
https://github.com/insightsengineering/teal/pull/1499/commits/6cf260d745d109cc618ddfd874245a6cd1cc6616

image

m7pr: CC @donyunardi @kumamiao what you think about it? Can we close this?

m7pr:
Making it a regular PR so we can get CI/CD running

DESCRIPTION
```diff
@@ -79,7 +79,7 @@ VignetteBuilder:
 RdMacros:
     lifecycle
 Remotes:
-    insightsengineering/teal.reporter@main,
+    insightsengineering/teal.reporter@redesign@main,
```
averissimo, line 82: I think you have to escape the second `@`

```suggestion
    insightsengineering/teal.reporter@redesign%40main,
```

R/module_nested_tabs.R
donyunardi: This is different from what I proposed which to have the button outside of teal module area.

image

This way, the arrow button for the side bar stay aligned for tmc and tmg.
What do you think @kumamiao ?

kumamiao: Agreed that it would be ideal to have consistent position for arrow buttons for filter panel and the module encodings if possible.

m7pr: Sorry @vedhav , can we ask for you precious involvement once again : P ?

vedhav: 👍It is moved outside the teal module region.
image

averissimo: An alternative go to Dony's idea of sidebar

(also showing only 1 button, left or centered)

image

donyunardi: I'll mark this as done for now as I created a new issue. 
https://github.com/insightsengineering/teal/issues/1533

Please add other ideas there.

averissimo (commented):
Some comments to discuss with @m7pr

averissimo:
:lady_beetle: Bug: Can only add 1 card

How to reproduce:

- Load example shiny app
- Add card with any title (e.g. "title 1")
- Add second card with a different title  (e.g. "title 2")
- Open the reporter tab
- Only 1 card is present as screenshot below shows

image

R/teal_reporter.R
```diff
@@ -194,3 +194,129 @@ TealSlicesBlock <- R6::R6Class( # nolint: object_name_linter.
...
+              sprintf(
+                "
+                $('#shiny-modal').on('shown.bs.modal', () => {
+                  $('#%s').focus()
+                })
+                ",
+                ns("label")
+              )
+            )
+          ),
```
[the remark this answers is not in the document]
m7pr: love the idea

R/teal_reporter.R
```diff
@@ -194,3 +194,129 @@ TealSlicesBlock <- R6::R6Class( # nolint: object_name_linter.
...
+          footer = shiny::div(
```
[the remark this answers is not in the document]
m7pr: wow

m7pr:
Thanks @averissimo for the thorough review! Those are all really useful updates to the current flow :)!

m7pr:
@averissimo the bug that you reported I was aware of

> Found a 🐞 bug: Can only add 1 card
> How to reproduce:

> Load example shiny app
> Add card with any title (e.g. "title 1")
> Add second card with a different title (e.g. "title 2")
> Open the reporter tab
> Only 1 card is present as screenshot below shows


This is because you are adding a second card, that has the same content as the first card (they have just different names, but the content is the same). There is some reactiveVal that keeps the state of the last added card, and if a new card is added with the same name, it prevents adding this card. 

If you add a new card inside the same module, and you at least specify a comment, then the content of the second card will be different and this will be possible to be added.

R/module_nested_tabs.R
```diff
@@ -395,21 +397,33 @@ srv_teal_module.teal_module <- function(id,
...
+      output$reporter_add_container <- renderUI({
+        req(reporter_card_out())
+        tags$div(
+          class = "teal add-reporter-container",
+          teal.reporter::add_card_button_ui(session$ns("reporter_add"))
+        )
+      })
+
+      add_document_button_srv("reporter_add", reporter = reporter, r_card_fun = reporter_card_out)
+    }
```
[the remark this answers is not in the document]
averissimo: Even now, I merged it in a47a24a

R/module_teal.R
```diff
@@ -227,6 +222,47 @@ srv_teal <- function(id, data, modules, filter = teal_slices()) {
       data_load_status = data_load_status,
       reporter = reporter
     )
+
+    if (!is.null(reporter)) {
```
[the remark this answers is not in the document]
averissimo: a47a24a

averissimo:
As agreed we are closing this PR in favor of https://github.com/insightsengineering/teal.reporter/pull/331 that is built on top of this one.

The overhead of keeping the 2 feature branches is becoming too cumbersome with multiple function renames going on.

Issue #1505, opened 2025-03-07
[Bug]: UI/UX issues to come back to


vedhav:
### Related PRs

- [ ] https://github.com/insightsengineering/teal/pull/1510
- [ ] https://github.com/insightsengineering/teal.widgets/pull/293
- [ ] https://github.com/insightsengineering/teal.modules.general/pull/872
- [ ] https://github.com/insightsengineering/teal.modules.clinical/pull/1370

#### Apps to test all modules of tmg and tmc:

<details>
<summary> tmg all modules </summary>

```r
devtools::load_all("teal.widgets")
devtools::load_all("teal")
devtools::load_all("teal.modules.general")


# ######################################################
#
#   _____                           _
#  |  __ \                         | |
#  | |  | | ___  ___ ___  _ __ __ _| |_ ___  _ __ ___
#  | |  | |/ _ \/ __/ _ \| '__/ _` | __/ _ \| '__/ __|
#  | |__| |  __/ (_| (_) | | | (_| | || (_) | |  \__ \
#  |_____/ \___|\___\___/|_|  \__,_|\__\___/|_|  |___/
#
#
#
#  Decorators
# #####################################################

plot_grob_decorator <- function(default_footnote = "I am a good decorator", .var_to_replace = "plot") {
  teal_transform_module(
    label = "Caption (grob)",
    ui = function(id) shiny::textInput(shiny::NS(id, "footnote"), "Footnote", value = default_footnote),
    server = function(id, data) {
      moduleServer(id, function(input, output, session) {
        logger::log_info("🟠 plot_grob with default: {default_footnote}!", namespace = "teal.modules.general")
        reactive({
          req(data(), input$footnote)
          logger::log_info("changing the footnote {default_footnote}", namespace = "teal.modules.general")
          teal.code::eval_code(data(), substitute(
            {
              footnote_grob <- grid::textGrob(footnote, x = 0, hjust = 0, gp = grid::gpar(fontsize = 10, fontface = "italic", col = "gray50"))
              # Arrange the plot and footnote
              .var_to_replace <- gridExtra::arrangeGrob(
                .var_to_replace,
                footnote_grob,
                ncol = 1,
                heights = grid::unit.c(grid::unit(1, "npc") - grid::unit(1, "lines"), grid::unit(1, "lines"))
              )
            },
            env = list(
              footnote = input$footnote,
              .var_to_replace = as.name(.var_to_replace)
            )
          ))
        })
      })
    }
  )
}
[...]

PR #1510, opened 2025-04-04
Final UI and bslib changes

vedhav:
Part of #1505


averissimo:
Should we try to replace all hardcoded colors being used? (there are just a few left, besides `validation.css` which will be handled in the other PR in teal)

```css
color: white
color: #fff
color: grey
```

m7pr (approved):

PR #1519, opened 2025-04-25
update snapshot manager button title

donyunardi:
Small PR to update snapshot manager button title


averissimo (approved):
Looks good and atomic :100:

Issue #1526, opened 2025-05-07
[Question]: How to return report to teal?


gogonzo:
### What is your question?

WIP....

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [x] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [x] I agree to follow this project's Security Policy.

gogonzo:
Report is returned to teal in the following process:
- teal_module's server should return reactive `teal_report` object
- `teal_report` object is observed in teal and passed to the `Reporter` once add button is clicked

PR #1541, opened 2025-06-03
📑 Reporter Refactor: Adds support for new `teal_report` and `teal_card` class

averissimo:
# Pull Request

- Fixes https://github.com/insightsengineering/teal/issues/1526
- Fixes https://github.com/insightsengineering/coredev-tasks/issues/648

Built on top of:

- https://github.com/insightsengineering/teal/pull/1499 _(will be closed once this PR is stable)_

### Companion PRs:

- https://github.com/insightsengineering/teal/pull/1541
- https://github.com/insightsengineering/teal.code/pull/255
- https://github.com/insightsengineering/teal.data/pull/370
- https://github.com/insightsengineering/teal.reporter/pull/331
- https://github.com/insightsengineering/teal.modules.general/pull/884

### Changes description

- [x] Adds a "Add to reporter" when a `teal_report` object is returned from the module
- [x] Changes `modify_reactive_output()` to `after()`

### Sample app

```r
# - teal_reporter (all 4: teal, teal.reporter, teal.code, teal.data)
# - redesign@main (teal, teal.reporter, teal.code)
pkgload::load_all("../teal")

example_extended <- function(label = "example teal module",
                                 datanames = "all",
                                 transformators = list(),
                                 decorators = list()) {
  checkmate::assert_list(decorators, "teal_transform_module")
  
  mod <- example_module(label, datanames, transformators, decorators)
  
  module(
    label,
    server = function(id, data, decorators) {
        result <- mod$server("example", data, decorators)
        
        reactive({
          data <- result()
          report(data) <- c(doc("## Code"), report(data), "## Table", data$object)  
          data
        })
        
      })
    },
    ui = function(id, decorators) mod$ui(shiny::NS(id, "example"), decorators),
    ui_args = mod$ui_args,
    server_args = mod$server_args,
    datanames = mod$datanames,
    transformators = mod$transformators
  )
}

example_old_reporter <- function(label = "example teal module",
                                 datanames = "all",
                                 transformators = list(),
                                 decorators = list()) {
  checkmate::assert_list(decorators, "teal_transform_module")
  
  ans <- module(
    label,
    server = function(id, data, decorators, reporter, filter_panel_api) {
        result <- example_module()$server("example", data, decorators)
        
          id = "rcode",
          verbatim_content = reactive(teal.code::get_code(req(result()))),
          title = "Example Code"
        )
        
        if (inherits(reporter, "Reporter")) {
          card_fun <- function(comment, label) {
            card <- teal::report_card_template(
              title = "Example plot",
              label = label,
              with_filter = FALSE,
              filter_panel_api = filter_panel_api
            )
            card$append_rcode(get_code(result()))
            card$append_text("Table", "header3")
            card$append_table(result()[["object"]])
            card
          }
          teal.reporter::simple_reporter_srv("simple_reporter", reporter = reporter, card_fun = card_fun)
        }
      })
    },
    ui = function(id, decorators) {
      ns <- NS(id)
        output = verbatimTextOutput(ns("example-text")),
        encoding = tags$div(
          teal.reporter::simple_reporter_ui(ns("simple_reporter")),
          selectInput(ns("example-dataname"), "Choose a dataset", choices = NULL),
          ui_transform_teal_data(ns("example-decorate"), transformators = decorators),
        )
      )
    },
    ui_args = list(decorators = decorators),
    server_args = list(decorators = decorators),
    datanames = datanames,
    transformators = transformators
  )
  attr(ans, "teal_bookmarkable") <- TRUE
  ans
}


teal::init(
  data = within(teal_data(), {iris <- iris}),
  modules = modules(
    example_extended(label = "🆕 Module (extended)"),
    example_old_reporter(label = "⏲️ Old reporter"),
    example_module(label = "🆕 Module (from {teal})"),
    example_module(label = "❌️ No reporter") |> disable_report()
  )
) |> shiny::runApp()
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,11 @@
 # teal 1.0.0.9002
 
+### New features
+
+* `init` and `srv_teal` have new `reporter` parameter, that allows to pre-define `teal.reporter::Reporter` object to be
+used for storing the content of the report. You can also globally disable reporting by setting `reporter = NULL`
+(and `disable = TRUE` in `ui_teal` for cases when `ui_teal` is used as shiny module).
+
 # teal 1.0.0
 
 ### Breaking changes
@@ -17,12 +23,6 @@
     - Modules: The module navigation is moved from a nested tab selection to a "Module" drop-down selection. The module selection can be done from the nested button of modules.
     - Reporter: The Report previewer is no longer displayed as yet another teal module, it is placed inside a "Report" drop-down right next to the "Module" drop-down. The Report previewer is a modal that displays the added report cards. The "Report" drop-down also contains other global report options like download/load/reset the Report.
     - Teal Data Module: The UI created by the `teal_data_module()` is now placed inside a modal that can only be closed when it returns `teal_data`.
-
-### New features
-
-* `init` and `srv_teal` have new `reporter` parameter, that allows to pre-define `teal.reporter::Reporter` object to be
-used for storing the content of the report. You can also globally disable reporting by setting `reporter = NULL` 
-(and `disable = TRUE` in `ui_teal` for cases when `ui_teal` is used as shiny module).
 
 # teal 0.16.0
 


R/after.R
```diff
@@ -0,0 +1,87 @@
...
+  new_x <- function() {
+    original_args <- as.list(environment())
+    original_args$id <- "wrapped"
+    if ("..." %in% names(formals(`_x`))) {
+      original_args <- c(original_args, list(...))
+    }
+    moduleServer(id, function(input, output, session) {
+      original_out <- if (all(c("input", "output", "session") %in% names(formals(`_x`)))) {
+        original_args$module <- `_x`
+        do.call(call_module, args = original_args)
```
averissimo: @gogonzo is this a typo? `call_module` is not defined anywhere

```suggestion
        do.call(shiny::callModule, args = original_args)
```

averissimo:
Some thoughts before closing for the week:

- "Add to reporter" is vulnerable to module output having a validation error / shinysilenterror (See `tm_missing_data` and see it momentarily show the same validation error as the module)
- `teal.reporter::disable_reporter()` returns `NULL`, this seems like a bad practice if we want to expand on the feature.

NAMESPACE
```diff
@@ -15,9 +15,11 @@ S3method(ui_teal_module,teal_modules)
 S3method(within,teal_data_module)
 export(TealReportCard)
 export(add_landing_modal)
+export(after)
```
pawelru, line 18: I think that this is a really bad name that can mean anything. It feels very much async related. Please reassess the name.

m7pr: Just to let you know @pawelru - this is a function that edits/manipulates `teal_data` (now extended to `teal_report` class) after it was returned by module's `server`. It is applied on a module level. 

So basically it is:

```
module() |>
    edit_data_returned_by_server()
```

R/after.R
```diff
@@ -0,0 +1,91 @@
...
+#' @param x (`teal_data`)
+#' @param ui (`function(id, elem, ...)`) function to receive output (`shiny.tag`) from `x$ui`
+#' @param server (`function(input, output, session, data, ...)`) function to receive output data from `x$server`
+#' @param ... additional argument passed to `ui` and `server` by matching their formals names.
+#' @return A `teal_report` object with the result of the server function.
+#' @export
+after <- function(x,
+                  ui = function(id, elem) elem,
+                  server = function(input, output, session, data) data,
+                  ...) {
```
m7pr: Since `after` is applied like below

```
<teal_module> |> after(
        server = function(data) {
          teal.reporter::teal_card(data) <- c(
            teal.reporter::teal_card(data),
            "# added content",
            "This content has been added through `after`"
          )
          data
        }
      )
```

I think 

```
#' @param x (`teal_data`)
```

should be updated to 

```
#' @param x (`teal_module`)
```

and 

```
server = function(input, output, session, data) data
```

to 

```
server = function(data) data
```

?

R/after.R
```diff
@@ -0,0 +1,91 @@
...
+  # todo: make a method for teal_app and remove teal_extend_server?
+  checkmate::assert_multi_class(x, "teal_module")
+  if (!is.function(ui) || !all(names(formals(ui)) %in% c("id", "elem"))) {
+    stop("ui should be a function of id and elem")
+  }
+  if (!is.function(server) || !all(names(formals(server)) %in% c("input", "output", "session", "data"))) {
+    stop("server should be a function of `input` and `output`, `session`, `data`")
+  }
```
m7pr, line 22: Are those checks even working if we pass server function without `input`, `output` and `session`?

```r
<teal_module> |> after(
        server = function(data) {
          teal.reporter::teal_card(data) <- c(
            teal.reporter::teal_card(data),
            "# added content",
            "This content has been added through `after`"
          )
          data
        }
      )
```

R/after.R
```diff
@@ -0,0 +1,91 @@
+#' Executes modifications to the result of a module
+#'
+#' Primarily used to modify the output object of module to change the containing
+#' report.
+#' @param x (`teal_data`)
```
m7pr:
```suggestion
#' @param x (`teal_module`)
```

NAMESPACE
donyunardi: How about `modify_teal_module` or `edit_teal_module`? 
Then, we'll have `.modify/edit_teal_module_ui` and `.modify/edit_teal_module_srv`

EDIT: I saw the comment that this was previously called `modify_reactive_output`.

donyunardi:
First off, I really appreciate the exploration and research that been poured into this PR.

I have two observations so far:

### Automatic conversion from `teal_data` to `teal_report` in module's server
With the current approach, `teal_data` is automatically converted into a `teal_report` object as soon as it enters a module's server, even if I don’t intend to use the reporter feature in my custom teal module.

I personally would prefer to see `teal_data` object unless I intentionally need a reporter feature.
It makes more sense to keep the data as `teal_data` and only convert to `teal_report` when I decided to use the reporter feature using the  `as.teal_report()` function.

### Extending `teal_data` class to make `teal_report` class
I do see value in converting `teal_data` to `teal_report`. This approach allows me to continue using all `teal_data` methods while also gaining access to new `teal_report` methods. In practice, I would just need to learn how to use `teal_card` and how to apply the `keep_output` argument within `eval_code()` and `within()` to capture objects for reporting.

However, this raises a design question: is it appropriate for `teal_report` to extend `teal_data`?

Historically, `teal_data` and `teal_report` have served different purposes and been treated as separate entity. Personally, I can see them as naturally related. Our reporting feature is about capturing the results of data visualizations or outputs, which are typically driven by the data, filter setup, and transformation inside `teal_data`.

However, it's worth noting that a `teal_card` can include any content that may not be related to `teal_data` object. If that happens, this may weaken the relationship between `teal_report` and `teal_data` and might justify the reason to keep the two class separate.

I'll keep thinking about this PR.

gogonzo:
> With the current approach, `teal_data` is automatically converted into a `teal_report` object as soon as it enters a module's server, even if I don’t intend to use the reporter feature in my custom teal module.

This happens anyway now, nobody raised this problem before and even it was appreciated:
- If module's server doesn't use join_keys `teal_data` isn't needed as `qenv` is enough.
- If module's server doesn't use reproducibility feature, `qenv` wouldn't be needed neither - `list` containing data would be enough.

> I personally would prefer to see `teal_data` object unless I intentionally need a reporter feature. It makes more sense to keep the data as `teal_data` and only convert to `teal_report` when I decided to use the reporter feature using the `as.teal_report()` function.

Using this in all modules wouldn't be a big deal.
```r
function(id, data, ...) {
  d <- reactive(as.teal_report(data))
  
  ... # use d
}
```

> ### Extending `teal_data` class to make `teal_report` class
> ... In practice, I would just need to learn how to use `teal_card` and how to apply the `keep_output` argument within `eval_code()` and `within()` to capture objects for reporting.

Not true. In practice module-developer doesn't have to learn anything unless one decide to migrate to new reporting system.
 
> However, it's worth noting that a `teal_card` can include any content that may not be related to `teal_data` object. If that happens, this may weaken the relationship between `teal_report` and `teal_data` and might justify the reason to keep the two class separate.

Yes, it would weaken a relationship, expose a report for more errors and limit functionality. It would also have it's codebase consequences with increased number of lines in teal_module's server. 
To build the report one would have to write an extra reactive and reuse teal_data used in the module's server and insert into a report according to the order. This sounds good on a paper because `@code` is unsubsettable. It means that `get_code` would return the whole code as a single code-chunk


If one undermines `teal_report` on top of `teal_data` and says that code and environment are conceptually related, I'd come up with the idea that code is one of the elements of the document and it's coexistence with environment in qenv is driven by specific _historical_ (hehe) need. `code` is really a part of the document.

image

- code_chunk is a separate class same as markdown content and metadata (header or document properties)
- card contains blocks and it can be anything
- teal_report inherits card and environment and serves as a environment with the document
- teal_data inherits from teal_report

This is just a quick justification or a provocation, because at the end object delivered to the module would have the same methods as it has now (on the feature branch).

Now, jokes aside. I consider current feature branch design as a good one where we didn't really made any compromises. Every new feature extends previous classes and adds new slots/methods on top.

m7pr:
I will just add that I think learning about `teal_report` after acknowleding `teal_data` is easier than learning the current `teal.reporter` `R6` `ReportCards`, which I still struggle with after 2-3 years on a project.

vedhav:
I prefer a solution on top of `teal_data` as well. I think in general R users find it hard to use R6 classes primarily due to the fact that they are not as common as S3 or S4 objects.
Building `teal_reporter` on top of `teal_data` makes learning so much easier. In my head the whole teal framework is an abstraction on top of `teal_data` and shiny modules. And we should keep it this way.

donyunardi:
> This happens anyway now, nobody raised this problem before and even it was appreciated:
If module's server doesn't use join_keys teal_data isn't needed as qenv is enough.
If module's server doesn't use reproducibility feature, qenv wouldn't be needed neither - list containing data would be enough.

When I conduct training/workshop, I usually show users that the `data` in custom module is `teal_data` class and create that connection of using `teal_data()` constructor when creating `teal_data` object in `init()`. If we automatically convert this to `teal_report` class, I am worried that it will create confusion and thus have to explain why we convert this to `teal_report` class.

We agreed that `teal_data` is that default data object that's being passed around teal framework. Is `qenv` or list of containing data enough? Sure. But I appreciate the consistency of just using `teal_data` and train user to understand `teal_data`, whether or not they're joining datasets or using the code reproducibility feature.

----

Although it's possible, it's hard for me to think that someone will intentionally include any content that is not related to the teal_data/real_report object. Arguably, I can even do that with the current R6 solution with the `$append_text()` method.

----

In addition to `teal_data` and `cdisc_data`, with the current approach, I can now do:
```
x <- teal_report(
  iris = iris, 
  mtcars = mtcars
)
```

Is the idea to also introduce `teal_report` as a constructor? I think this may be one too many. Perhaps this is also further support my suggestion to convert `teal_data` to `teal_report` inside the module. 

----

From release perspective, this means teal.data will have to be released before teal.reporter. Not a big deal we agree if this is the way to go.

donyunardi:
The team had another go at discussing this solution.

The fundamental change in this approach is that we are now including `teal.reporter` as a default feature of the teal framework. This is reflected by the fact that the object we pass to the module will be reporter-ready.

We also talked about the benefit of automatically converting this to `teal_report`, which allows the initial `teal_card` creation to be handled by teal. This means we can create a separate code chunk for data pre-processing and filtering activity before the object enters the module server. The user can then focus only on adding the module’s analysis code chunk to the `teal_report` object.

This also means that module developers no longer need to worry about passing the `filter_panel_api` object to the module, which is mostly used to include filtering information when crafting the reporter. Advanced teal module developers can still bring in the `filter_panel_api` if needed as we're not removing this access.

Modules need to return a reactive `teal_report` object, so developers need to ensure this is the final object in the module. We need to make this convention clear. When followed, the "Add to Report" card will be added automatically, with no UI or server setup needed. It would be helpful to add a check to alert developers if the module doesn’t return a reactive `teal_report` object.

From the user’s perspective, I think this solution will be well received by our module developers and will simplify the process of building teal modules.

This solution may also enable automatic creation of the "Show R Code" button, another feature we may offer by default in the teal framework.

We will not introduce `teal_report` as a separate constructor, and `teal_data` will continue to be the entry point. We won’t expose this change in our documentation.

There is one question that I thought when drafting this summary, if in the future we decide to add another feature involving both data and reporting, how easy will it be to extend the proposed design? Would we need to extend `teal_report` into another S4 class to add new slots and methods? Is bookmarking feature relatable with this topic?

kumamiao:
From user's perspective, I like that this proposal makes it much simpler for the user (module developer) to be able to create a module with the reporter feature, in a nice structured way.

In terms of converting the `teal_data` object to `teal_report`, in my opinion, most users would not really care too much about the switch of class of the object, as long as we provide clear guidance on how this object is used to add the report feature within the module. We can also mention the switch in the vignette. Another point is that `teal_data` still inherits the `teal_data` class, as demonstrated in the call, if user ever encounters any confusion.

Agree to Not expose or instruct users to use `teal_report` to create a `teal_data` object, and keep `teal_data` as entry point.

R/teal_reporter.R
```diff
@@ -194,3 +194,168 @@ TealSlicesBlock <- R6::R6Class( # nolint: object_name_linter.
     teal_slices = NULL # teal_slices
   )
 )
+
+#' Server function for `teal_card` class.
+#' @keywords internal
+add_document_button_srv <- function(id, reporter, r_card_fun) {
```
m7pr: @gogonzo @averissimo is this function still used?
I think we use `teal.report::add_card_button_srv` in `teal::srv_add_reporter`

averissimo: Good catch, I'll remove it

R/module_teal.R
```diff
@@ -50,9 +50,7 @@ ui_teal <- function(id, modules) {
   checkmate::assert_character(id, max.len = 1, any.missing = FALSE)
   checkmate::assert_class(modules, "teal_modules")
   ns <- NS(id)
-
-  modules <- append_reporter_module(modules)
-
+  modules <- drop_module(modules, "teal_previewer_module")
```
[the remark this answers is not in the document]
averissimo: We don't need this here as the Reporter module is not yet added/used

```suggestion
```

DESCRIPTION
```diff
@@ -38,7 +38,7 @@ BugReports: https://github.com/insightsengineering/teal/issues
 Depends:
```
m7pr: @averissimo `teal` version should be vbumped to something higher than on main. so at least 1.0.0

for now it's

```r
Version: 0.16.0.9029
```

averissimo: Done! Thanks

DESCRIPTION
```diff
@@ -77,8 +77,13 @@ Suggests:
...
     rmarkdown
-RdMacros:
-    lifecycle
+Remotes:
+    insightsengineering/teal.reporter@teal_reportable,
+    insightsengineering/teal.logger@main,
+    insightsengineering/teal.data@main,
+    insightsengineering/teal.code@main,
+    insightsengineering/teal.slice@main,
+    insightsengineering/teal.widgets@main
```
m7pr:
```suggestion
    insightsengineering/teal.reporter@teal_reportable
```

NAMESPACE
m7pr: ok, we put a refernce on the vignettes issue and we will resoncider naming while writing the vignette
https://github.com/insightsengineering/teal.reporter/issues/314

R/after.R
m7pr: the check is fine, it's just the limitation to possible server names to input, output, session and data

NEWS.md
```diff
@@ -18,6 +18,13 @@
     - Reporter: The Report previewer is no longer displayed as yet another teal module, it is placed inside a "Report" drop-down right next to the "Module" drop-down. The Report previewer is a modal that displays the added report cards. The "Report" drop-down also contains other global report options like download/load/reset the Report.
     - Teal Data Module: The UI created by the `teal_data_module()` is now placed inside a modal that can only be closed when it returns `teal_data`.
 
+### New features
+
+* `init` and `srv_teal` have new `reporter` parameter, that allows to pre-define `teal.reporter::Reporter` object to be
+used for storing the content of the report. You can also globally disable reporting by setting `reporter = NULL` 
+(and `disable = TRUE` in `ui_teal` for cases when `ui_teal` is used as shiny module).
+* TODO: verify if we need to clone/deep_clone reporter in `srv_teal/init`.
```
averissimo:
```suggestion
```

R/after.R
```diff
@@ -0,0 +1,91 @@
...
+#' @param ui (`function(id, elem, ...)`) function to receive output (`shiny.tag`) from `x$ui`
+#' @param server (`function(input, output, session, data, ...)`) function to receive output data from `x$server`
+#' @param ... additional argument passed to `ui` and `server` by matching their formals names.
+#' @return A `teal_report` object with the result of the server function.
+#' @export
+after <- function(x,
+                  ui = function(id, elem) elem,
+                  server = function(input, output, session, data) data,
+                  ...) {
+  # todo: make a method for teal_app and remove teal_extend_server?
```
averissimo, line 15: Create issue for this before we merge

R/teal_reporter.R
```diff
@@ -85,112 +75,109 @@ TealReportCard <- R6::R6Class( # nolint: object_name.
   )
```
m7pr, line 65: Since `TealSlicesBlock` was deleted, do we need `dispatch_block`?

```r
  private = list(
    dispatch_block = function(block_class) {
      if (exists(block_class, getNamespace("teal"))) {
        # for block classes which are in teal (TealSlicesBlock)
        get(block_class)
      } else {
        # other block classes are in teal.reporter so we need to use super (ReporterCard) class
        super$dispatch_block(block_class)
      }
    }
  )
```

R/teal_reporter.R
```diff
@@ -22,14 +22,7 @@ TealReportCard <- R6::R6Class( # nolint: object_name.
     #' )
```
m7pr, line 23: `TealReportCard` inherits from `teal.reporter::ReportCard`. Should this be deprecated, if we ever deprecate `teal.reporter::ReportCard`? Or this will be auto-deprecated once it's deprecated in `teal.reporter`?

averissimo: I think it should self-deprecate if we place the call inside the `initialize` function. 

However we should update the badge to `deprecated` as well so that this may be flagged when removing it :-)

m7pr: Alrighty, so moving this comment also to https://github.com/insightsengineering/teal.reporter/issues/315

R/teal_reporter.R
m7pr: I think we can remove this `dispatch_block` and it will be used from the `super` class

m7pr: removed in here https://github.com/insightsengineering/teal/pull/1541/commits/823b008a05b1a43fa230877836060b2787aefef8

DESCRIPTION
```diff
@@ -77,8 +77,8 @@ Suggests:
 VignetteBuilder:
     knitr,
     rmarkdown
-RdMacros:
-    lifecycle
+Remotes:
```
m7pr, line 82:
```suggestion
RdMacros:
    lifecycle
Remotes:
```

PR #1542, opened 2025-06-04
Naming

m7pr:
Companion to https://github.com/insightsengineering/teal.reporter/pull/334
Consequence of changing naming convention for `teal_report` object.

### DIFF
--- a/R/module_init_data.R
+++ b/R/module_init_data.R
@@ -93,7 +93,7 @@ srv_init_data <- function(id, data) {
   data_teal_report <- as.teal_report(data)
   if (!inherits(data, "teal_report")) {
     teal.reporter::report(data_teal_report) <- c(
-      teal.reporter::doc(),
+      teal.reporter::document(data_teal_report),
       "# Code preparation",
       teal.reporter::report(data_teal_report)
     )
--- a/R/teal_data_utils.R
+++ b/R/teal_data_utils.R
@@ -21,8 +21,8 @@ NULL
   checkmate::assert_class(data, "teal_data")
   if (length(code) && !identical(code, "")) {
     data@code <- c(data@code, code2list(code))
-    teal.reporter::report(data) <- c(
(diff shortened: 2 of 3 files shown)


R/module_init_data.R
```diff
@@ -104,7 +104,7 @@ srv_init_data <- function(id, data) {
       list(
         code = trimws(c(teal.code::get_code(data_teal_report), hashes), which = "right"),
         join_keys = teal.data::join_keys(data_teal_report),
-        report = teal.reporter::report(data_teal_report)
+        report = teal.reporter::card(data_teal_report)
```
averissimo:
```suggestion
        card = teal.reporter::card(data_teal_report)
```

averissimo (approved):

R/module_init_data.R
```diff
@@ -92,10 +92,10 @@ srv_init_data <- function(id, data) {
   hashes <- .get_hashes_code(data)
   data_teal_report <- as.teal_report(data)
   if (!inherits(data, "teal_report")) {
-    teal.reporter::report(data_teal_report) <- c(
-      teal.reporter::doc(),
+    teal.reporter::teal_card(data_teal_report) <- c(
+      teal.reporter::teal_card(data_teal_report),
```
m7pr: @averissimo just double checking this change.
previously it looks like it was setting empty doc(), then adding `# Code preparation` and then extracting the report.

I think the first call should be `teal.reporter::teal_card(),` instead of `teal.reporter::teal_card(data_teal_report),` to maintain the same mechanism

averissimo: Ok! Keep as it was.

averissimo: Like this, right?

```suggestion
      teal.reporter::teal_card(),
```

m7pr: yes, thanks

averissimo (approved):

PR #1545, opened 2025-06-06
Vignettes

m7pr:
Extends teal vignettes with new reporting features in `teal`.

### DIFF
--- a/vignettes/getting-started-with-teal.Rmd
+++ b/vignettes/getting-started-with-teal.Rmd
@@ -126,8 +126,84 @@ For further details see [Filter Panel vignette](filter-panel.html).
 
 ### Reporting
 
-If any of the `modules` in your `teal` application support reporting (see [`teal.reporter`](https://insightsengineering.github.io/teal.reporter/) for more details), users of your application can add the outputs of the modules to a report.
-This report can then be downloaded and a special _Report Previewer_ module will be added to your application as an additional tab, where users can view and configure their reports before downloading them. See more details in [this vignette](adding-support-for-reporting.html).
+The `init` function includes a `reporter` parameter that controls the global reporting functionality of your `teal` application. This parameter determines whether and how users can generate reports from the analyses performed in your application.
+
+#### Configuring the Reporter Parameter
+
+The `reporter` parameter in `init` accepts three types of values:
+
+##### 1. Default Reporting (parameter omitted)
+When you don't specify the `reporter` parameter, `teal` automatically enables default reporting functionality:
+
+```{r default_reporting}
+# Default reporting - automatically enabled
+app_default <- init(
+  data = teal_data(IRIS = iris, MTCARS = mtcars),
+  modules = modules(
+    example_module("Analysis 1"),
+    example_module("Analysis 2")
+  )
+)
+```
+
+With default reporting enabled:
+- A "Report previewer" tab is automatically added to your application
+- Users can add module outputs to reports and download them
+- Reports include reproducible code for all analyses
+
+##### 2. Disabled Reporting (`reporter = NULL`)
+To completely disable reporting functionality across your entire application:
+
+```{r disabled_reporting}
+# Reporting disabled - no report functionality available
+app_no_reporting <- init(
+  data = teal_data(IRIS = iris, MTCARS = mtcars),
+  modules = modules(
+    example_module("Analysis 1"),
+    example_module("Analysis 2")
+  ),
+  reporter = NULL
+)
+```
+
+When `reporter = NULL`:
+- No "Report previewer" tab appears
+- Modules cannot add content to reports
+- All reporting-related UI elements are hidden
+
+##### 3. Custom Reporter (R6 object)
+For advanced use cases, you can provide a custom reporter object that inherits from the appropriate R6 class:
+
+```{r custom_reporter}
+library(teal.reporter)
+
+# Create a custom reporter with specific configuration
+custom_reporter <- teal.reporter::Reporter$new()
+
+# Use the custom reporter in your app
+app_custom <- init(
+  data = teal_data(IRIS = iris, MTCARS = mtcars),
+  modules = modules(
+    example_module("Analysis 1"),
+    example_module("Analysis 2")
+  ),
+  reporter = custom_reporter
+)
+```
+
+#### Choosing the Right Configuration
+
+* **Use default reporting** (omit parameter) for most applications where you want standard reporting functionality
+* **Use `reporter = NULL`** when you want to disable reporting entirely, perhaps for:
+  - Demo applications
+  - Apps where reporting is not needed
+  - Performance-critical applications
+* **Use a custom reporter** when you need:
+  - Special report formatting
+  - Custom report templates
+  - Integration with external reporting systems
+
+For more details on reporting functionality and how to implement it in your modules, see [this vignette](adding-support-for-reporting.html) and the [`teal.reporter`](https://insightsengineering.github.io/teal.reporter/) package documentation.
 
 
 ### Reproducible code


vignettes/getting-started-with-teal.Rmd
```diff
@@ -126,8 +126,109 @@ For further details see [Filter Panel vignette](filter-panel.html).
...
+)
+```
+
+When `reporter = NULL`:
+- No "Report previewer" tab appears
+- Modules cannot add content to reports
+- All reporting-related UI elements are hidden
+
+##### 3. Custom Reporter (R6 object)
+For advanced use cases, you can provide a custom reporter object that inherits from the appropriate R6 class:
```
gogonzo:
```suggestion
##### 3. Custom Reporter 
For advanced use cases, you can provide a custom initial reporter object:
```

gogonzo (commented):
1. "Adding Support for Reporting to Custom Modules" uses old solution with `reporter` arg in the <teal modules>$server
2. "getting started with teal" has very short subsections which links to longer documentations. Reporter here adds extra 100 lines which changes look of the vignette so much. IMO describing possible values of `init(reporter)` can be shorter just bullet-points instead of sections. Maybe we should change "Adding Support for Reporting to Custom Modules" to "Creating reproducible reports" (name tbd) and describe everything related to report there?

m7pr:
@gogonzo about

> Maybe we should change "Adding Support for Reporting to Custom Modules" to "Creating reproducible reports" (name tbd) and describe everything related to report there?

Do you think this should be a document in `teal` or in `teal.reporter`? I also have a feeling that this needs a special document that covers only reporting, which is now very extensive : )

gogonzo:
> @gogonzo about
> 
> > Maybe we should change "Adding Support for Reporting to Custom Modules" to "Creating reproducible reports" (name tbd) and describe everything related to report there?
> 
> Do you think this should be a document in `teal` or in `teal.reporter`? I also have a feeling that this needs a special document that covers only reporting, which is now very extensive : )

We can do this:
- change a content "Adding Support for Reporting to Custom Modules" to fit a new design. This vignette is also linked in "Creating custom modules", so I guess it is ok to keep it, but update the content
- Create a new vignette called "Managing Reproducible Report Documents in Teal" where we describe:
  a. `init(reporter)` to control initial state of the reporter
  b.  Managing teal_module(s) report cards (with `disable_report` and `after`)

m7pr:
@gogonzo make sense!
Do we need special vignette to cover `teal_report`? Or can it be included in "Adding Support for Reporting to Custom Modules"?

gogonzo:
> @gogonzo make sense! Do we need special vignette to cover `teal_report`? Or can it be included in "Adding Support for Reporting to Custom Modules"?

We for sure need `teal_report` vignette in `teal.report`. I don't know where in teal should we mention `teal_report`. Maybe in "Include Data in Teal Application"? I wouldn't spend too much time about `teal_report` as it serves the same role as `teal_data` and also it should be possible to specify `init(data = <teal_report>)`.

m7pr:
Ok, I:

- simplified and shortened `teal_report Class` in `teal.reporter`
- simplified and shortened `Getting started with teal.modules.general` in `teal`
- updated `Adding Support for Reporting to Custom Modules` in `teal`
- created `Managing Reproducible Report Documents in teal` in `teal`

Left 3 TODO in `Managing Reproducible Report Documents in teal` in `teal`

vignettes/managing-reproducible-report-documents-in-teal.rmd
```diff
@@ -0,0 +1,194 @@
...
+library(teal.reporter)
+
+# Create a custom reporter with specific configuration
+custom_reporter_template <- Reporter$new()
+custom_reporter_template$set_template(
+  # Need to return named `teal_card`
+  function(card) {
+    tcard <- c("Templated information", card)
+    attributes(tcard) <-  attributes(card)
+    tcard
```
m7pr: This is nasty. `c` removes all classes and attributes. We can substitute this example to use `edit_teal_card`.
So it's

```
custom_reporter_template$set_template(
  function(card) {
    edit_teal_card(card, append = "Templated information", after = 0)
  }
)
```

vignettes/adding-support-for-reporting.Rmd
```diff
@@ -93,23 +93,24 @@ knitr::include_url(url, height = "800px")
 
 ## Add support for reporting
 
-### Modify the declaration of the server function
+### Modify the body of the server function
 
-The first step is to add an additional argument to the server function declaration - `reporter`.
-This informs `teal` that the module requires `reporter`, and it will be included when the module is called.
+To inform `teal` that the module requires reporting, return interactive `teal_data` from the `server` function.
```
gogonzo:
```suggestion
To inform `teal` that the module requires reporting, return `reactive` containing `teal_report` from the `server` function.
```

vignettes/adding-support-for-reporting.Rmd
```diff
@@ -93,23 +93,24 @@ knitr::include_url(url, height = "800px")
...
+Outputs of the module will be included in the report when the module is called.
```
gogonzo: Card is added to the report on demand, not when the module is clicked 
```suggestion
Current outputs of the modules and a reproducible code will be included in the module when "add to report" button is clicked.
```

vignettes/adding-support-for-reporting.Rmd
```diff
@@ -93,23 +93,24 @@ knitr::include_url(url, height = "800px")
...
 See below:
 
 ```{r module_2}
 my_module_with_reporting <- function(label = "example teal module") {
   module(
     label = label,
-    server = function(id, data, reporter) {
+    server = function(id, data) {
```
gogonzo: This module returns unmodified data while output shows the selected dataset.

vignettes/getting-started-with-teal.Rmd
```diff
@@ -126,9 +126,18 @@ For further details see [Filter Panel vignette](filter-panel.html).
...
 
-If any of the `modules` in your `teal` application support reporting (see [`teal.reporter`](https://insightsengineering.github.io/teal.reporter/) for more details), users of your application can add the outputs of the modules to a report.
-This report can then be downloaded and a special _Report Previewer_ module will be added to your application as an additional tab, where users can view and configure their reports before downloading them. See more details in [this vignette](adding-support-for-reporting.html).
+The `init` function includes a `reporter` parameter that controls the global reporting functionality of your `teal` application. This parameter determines whether and how users can generate reports from the analyses performed in your application.
 
+The `reporter` parameter in `init` accepts three types of values:
+
+* `parameter is omitted`: reporting is enabled by default globally.
+* `NULL`: reporting is disabled globally.
+* `teal.reporter::Report$new()`: predefined reporter object.
```
gogonzo:
```suggestion
* `teal.reporter::Report$new()`: predefined reporter object, possibly containing some initial cards.
```

vignettes/managing-reproducible-report-documents-in-teal.rmd
```diff
@@ -0,0 +1,195 @@
...
+output: rmarkdown::html_vignette
+vignette: >
+  %\VignetteIndexEntry{Managing Reproducible Report Documents in teal}
+  %\VignetteEngine{knitr::rmarkdown}
+  %\VignetteEncoding{UTF-8}
+---
+
+## Introduction
+
+`teal` provides comprehensive reporting capabilities that allow users to generate reproducible documents from their analysis modules. This vignette covers the key components for managing report documents in `teal` applications.
```
gogonzo:
```suggestion
`teal` provides reporting capabilities that allow app users to compose reproducible documents from their analysis modules. This vignette covers the key components for managing report documents in `teal` applications.
```

vignettes/managing-reproducible-report-documents-in-teal.rmd
```diff
@@ -0,0 +1,195 @@
...
+
+The reporting system offers three levels of control:
+
+1. **Global reporting control** via the `reporter` parameter in `init()`
+2. **Module-level control** using `disable_report()` to exclude specific modules  
+3. **Report customization** with the `after` function to enhance report content
```
gogonzo: Three levels suggests that you go deeper or higher in the hierarchy. We operate here only on global/module (local) level.
```suggestion
The reporting system offers two types of control:

1. **Global reporting control** via the `reporter` parameter in `init()`
2. **Module-level control** via `after()` and `disable_report()`
```

vignettes/managing-reproducible-report-documents-in-teal.rmd
```diff
@@ -0,0 +1,195 @@
...
+
+This vignette demonstrates how to use these features effectively to create flexible and customized reporting workflows.
+
+## Global and Local Reporting Control
```
gogonzo: I think it's worth to split Global and Local according to the bullet points from above

```suggestion
## Global Reporting Control
```

vignettes/managing-reproducible-report-documents-in-teal.rmd
```diff
@@ -0,0 +1,195 @@
...
+Custom initial reporter can also include a template function that will be applied to every card:
+
+```{r custom_reporter_template}
+library(teal.reporter)
+
+# Create a custom reporter with specific configuration
+custom_reporter_template <- Reporter$new()
+custom_reporter_template$set_template(
+  # Need to return named `teal_card`
+  function(card) {
```
gogonzo, line 103: Can you show some practical information, like setting a Rmd template or adding some header/footer etc.?

vignettes/managing-reproducible-report-documents-in-teal.rmd
gogonzo: You should use `append(card, <something to add>, after = 0)` - then it doesn't loose a class because a first argument is a `card`

vignettes/managing-reproducible-report-documents-in-teal.rmd
```diff
@@ -0,0 +1,195 @@
...
+  ),
+  reporter = custom_reporter_template
+)
+# Start app with
+# if (interactive()) shinyApp(app_custom$ui, app_custom$server)
+```
+
+#### Module-Level Reporting Control
+
+While the `reporter` parameter controls reporting globally, you can also disable reporting for specific modules using `disable_report()`. This allows you to have fine-grained control over which modules support reporting:
```
gogonzo, line 124: Both, `after` and `disable_report` can control module's report. After is more powerful and it doesn't restrict to the report only but to the whole `teal_report/teal_data` object. We can showcase how to modify the report using `after` 
 
```suggestion
While the `reporter` parameter controls reporting globally, you can also manage a report document for a specific module using `after()` `disable_report()`:
```

vignettes/managing-reproducible-report-documents-in-teal.rmd
m7pr: Yes, I think I'll go with this one

vignettes/managing-reproducible-report-documents-in-teal.rmd
m7pr: Make sense, thanks!

vignettes/managing-reproducible-report-documents-in-teal.rmd
m7pr: @gogonzo `append` doesn't loose the class, but it looses `metadata` attribute, so there is no name to be displayed in Reporter previewer.

```r
> custom_card <- teal_card()
> metadata(custom_card, "title") <- "The title"
> custom_card
An object of class "teal_card"
Slot "metadata":
Slot "metadata":$title
[1] "The title"

> append(custom_card, "x")
[[1]]
[1] "x"

> append(custom_card, teal_card("x"))
[[1]]
[1] "x"

attr(,"class")
[1] "teal_card"
> metadata(append(custom_card, teal_card("x")))
list()
```

What do you think about exposing `append.teal_card`  method that also keeps attributes?
We can actually just change `edit_teal_card` to `append.teal_card` and remove the part used to `modify`

image

vignettes/adding-support-for-reporting.Rmd
m7pr: Thanks @gogonzo for pointing this out. I wanted to keep the example easy, so we just show that it is enough to `return(data)` to turn reporting on.

We can extend this example module by adding

```r
rdata <- reactive({
  eval_code(
    data(), 
    substitute(head(dataname), env = list(dataname = as.name(input$dataname))), 
    keep_output = TRUE
  )
})
return(rdata)
```
    
or

```r    
rdata <- reactive({
  req(input$dataname)
  obj <- data()
  teal_card(obj) <- c(teal_card(obj), "## Glimpse of the dataset", head(obj[[input$dataname]]))
  obj
})

return(rdata)
```

depending on whether we would like to build the report with `eval_code` or with `teal_card`.
I think `eval_code` is more appropriate here, even though we have `substitute` usage in it (that I think is advanced for regular R users). We can also extend this example by explaining the `keep_output = TRUE`

vignettes/managing-reproducible-report-documents-in-teal.rmd
m7pr: I was rather thinking about setting some templated text for disclaimers, data sources, data privacy information, etc as an information added on top or at the bottom of the card. This sentence can also be added to this example.

The `Rmd` template could be useful, if we create some text that can be interpreted as `rmarkdown`. Maybe including a logo that is taken from the web?

vignettes/adding-support-for-reporting.Rmd
m7pr: However this very same code is added in the next section `Add content to the report card`

So maybe keep `Modify the body of the server function` as it is, and extend `Add content to the report card` to use `eval_code`

vignettes/managing-reproducible-report-documents-in-teal.rmd
m7pr: @gogonzo there is a section following this sentence that presents the usage of `after`:
`Using the `after` Function to Customize Module Report Cards`

Do you think it's enough? It shows that you can adjust the `teal_card` of `teal_report`, but we can apply `join_keys` modifications as well, and also an `eval_code` to show that the whole `teal_data/teal_report` can be changed in any direction, not only the `@teal_report` field.

vignettes/adding-support-for-reporting.Rmd
gogonzo: I don't know if this vignette needs to have two examples. Previously adding reporter was more complicated now Reporter is supported by default as `teal_report` is passed to the teal_module's server. Only thing which user has to do is `return(<teal_report>)`. I'd just mention `keep_output` and add one markdown-heading element.

m7pr: Ok, I think I can remove the whole first example and just use the last one

m7pr: I simplified adding-support-for-reporting.Rmd vignette https://github.com/insightsengineering/teal/pull/1545/commits/2cca04bca1ab9d0f40034ae193ed6b68ad2ac124

vignettes/managing-reproducible-report-documents-in-teal.rmd
```diff
@@ -0,0 +1,195 @@
...
+
+##### 3. Custom Reporter 
+For advanced use cases, you can provide a custom initial reporter object:
+
+```{r custom_reporter}
+library(teal.reporter)
+
+# Create a custom reporter with specific configuration
+custom_reporter <- Reporter$new()
+custom_card <- teal_card("## Welome", "Welcome message", head(iris))
```
pawelru:
```suggestion
custom_card <- teal_card("## Welcome", "Welcome message", head(iris))
```

vignettes/managing-reproducible-report-documents-in-teal.rmd
```diff
@@ -0,0 +1,195 @@
...
+  reporter = custom_reporter                        # TODO: doesn't work yet, app do not start with a predefined card
+)
+# Start app with
+# if (interactive()) shinyApp(app_custom$ui, app_custom$server)
+```
+
+##### 4. Custom Reporter with a template
+
+Custom initial reporter can also include a template function that will be applied to every card.
+With templated approach you can included disclaimers, data sources, data privacy information, etc.
```
pawelru:
```suggestion
With the templated approach, you can included disclaimers, data sources, data privacy information, etc.
```

R/dummy_functions.R
```diff
@@ -11,6 +11,16 @@
...
+#'
+#' This module returns an object of class `teal_module`, that contains a `server` function.
+#' Since the server function returns a `teal_report` object, this makes this module reportable, which means that
+#' the reporting functionality will be turned on automatically by the `teal` framework.
+#'
+#' For more information on reporting in `teal`, see the vignettes:
+#' - `vignette("managing-reproducible-report-documents-in-teal", package = "teal")`
+#' - `vignette("adding-support-for-reporting-to-custom-modules", package = "teal")`
 #'
 #' @return A `teal` module which can be included in the `modules` argument to [init()].
```
averissimo, line 23: idea from https://github.com/insightsengineering/teal.modules.general/pull/884#discussion_r2357631517

Consider extending the `@return` tag to describe what the `srv` function may return (which in turn, activates the `Add to reporter` and `Show R code`

NEWS.md
```diff
@@ -24,6 +24,13 @@ used for storing the content of the report. You can also globally disable report
     - Reporter: The Report previewer is no longer displayed as yet another teal module, it is placed inside a "Report" drop-down right next to the "Module" drop-down. The Report previewer is a modal that displays the added report cards. The "Report" drop-down also contains other global report options like download/load/reset the Report.
     - Teal Data Module: The UI created by the `teal_data_module()` is now placed inside a modal that can only be closed when it returns `teal_data`.
 
+### New features
+
+* `init` and `srv_teal` have new `reporter` parameter, that allows to pre-define `teal.reporter::Reporter` object to be
+used for storing the content of the report. You can also globally disable reporting by setting `reporter = NULL` 
+(and `disable = TRUE` in `ui_teal` for cases when `ui_teal` is used as shiny module).
+* TODO: verify if we need to clone/deep_clone reporter in `srv_teal/init`.
```
m7pr:
```suggestion
```

tests/testthat/test-shinytest2-reporter.R
```diff
@@ -53,6 +53,10 @@ testthat::test_that("e2e: reporter card can be customized", {
   skip("TODO")
 })
 
+testthat::test_that("e2e: reporter card can be customized", {
+  skip("TODO")
+})
+
```
m7pr:
```suggestion
```

vignettes/adding-support-for-reporting.Rmd
```diff
@@ -16,81 +16,16 @@ The `teal` package offers an integrated reporting feature utilizing the `teal.re
 For a comprehensive explanation of the reporting functionality itself, please refer to the documentation therein.
 
 This article is intended for module developers and aims to provide guidance on enhancing a custom `teal` module with an automatic reporting feature.
-This enhancement enables users to incorporate snapshots of the module outputs into a report which can then be reviewed in another module automatically provided by `teal`.
+This enhancement enables users to incorporate snapshots of the module outputs into a report, which can be viewed in another module called `Reporter previewer` (automatically provided by `teal`).
```
m7pr:
```suggestion
This enhancement enables users to incorporate snapshots of the module outputs into a report, which can be viewed in another modal called `Reporter previewer` (automatically provided by `teal`).
```

vignettes/managing-reproducible-report-documents-in-teal.rmd
```diff
@@ -0,0 +1,195 @@
...
+custom_reporter$append_cards(list(WelcomeCard = custom_card))
+
+# Use the custom reporter in your app
+app_custom <- init(
+  data = teal_data(IRIS = iris, MTCARS = mtcars),
+  modules = modules(
+    example_module("Analysis 1"),
+    example_module("Analysis 2")
+  ),
+  reporter = custom_reporter                        # TODO: doesn't work yet, app do not start with a predefined card
```
m7pr:
```suggestion
  reporter = custom_reporter
```

vignettes/managing-reproducible-report-documents-in-teal.rmd
```diff
@@ -0,0 +1,195 @@
...
+library(teal.reporter)
+
+# Create a custom reporter with specific configuration
+custom_reporter_template <- Reporter$new()
+custom_reporter_template$set_template(
+  # Need to return named `teal_card`
+  function(card) {
+    tcard <- c("Templated information", card)
+    attributes(tcard) <-  attributes(card)
+    tcard
```
m7pr:
```suggestion
  function(card) {
    disclaimer <- teal.reporter::teal_card("Here comes disclaimer text")
    c(disclaimer, card)
```

m7pr:
Fixed in here https://github.com/insightsengineering/teal/pull/1593

PR #1559, opened 2025-07-14
WIP: Fixes TealAppDriver & implement regression tests

### LINKED ISSUE #1535: Implement regression test
### Feature description

Related with https://github.com/insightsengineering/coredev-tasks/issues/589

### Scope of work

teal (40 hours)
- teal_data_module interaction
- teal_transform_module
- common data warnings and errors in teal app
- Check if the module can run, use example_module()
- Snapshot manager (check the modal shows with the right design)

### Code of Conduct

### Contribution Guidelines

### Security Policy

averissimo:
# Pull Request

Fixes #1535 

### Changes description

- [ ] Remove `testthat::skip("chromium")` from e2e tests
- [ ] Ensure all e2e pass
- [ ] Add regression tests

### DIFF
--- a/R/TealAppDriver.R
+++ b/R/TealAppDriver.R
@@ -452,7 +452,7 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
             replacement = "",
             self$get_text(
               sprintf(
-                "#%s-filters-%s .filter-card-varname",
+                "#%s-filters-%s-container .filter-card-varname",
                 self$active_filters_ns(),
                 x
               )
@@ -694,7 +694,7 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
       tab_ns <- unlist(lapply(names(active_tab_inputs), function(name) {
         gsub(
           pattern = "-active_module_id$",
-          replacement = sprintf("-%s", active_tab_inputs[[name]]),
+          replacement = sprintf("-nav-%s", .label_to_id(active_tab_inputs[[name]])),
           name
         )
       }))
@@ -707,19 +707,9 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
           }
         }
       }
-      private$ns$module <- sprintf("%s-%s", active_ns, "module")
-
-      components <- c("filter_panel", "data_summary")
-      for (component in components) {
-        if (
-          !is.null(self$get_html(sprintf("#%s-%s-panel", active_ns, component))) ||
-            !is.null(self$get_html(sprintf("#%s-%s-table", active_ns, component)))
-        ) {
-          private$ns[[component]] <- sprintf("%s-%s", active_ns, component)
-        } else {
-          private$ns[[component]] <- sprintf("%s-module_%s", active_ns, component)
-        }
-      }
+      private$ns$module <- sprintf("%s-module", active_ns)
+      private$ns$filter_panel <- sprintf("%s-filter_panel", active_ns)
+      private$ns$data_summary <- sprintf("%s-data_summary", active_ns)
     },
     # @description
     # Get the active filter values from the active filter selection of dataset from the filter panel.


tests/testthat/test-shinytest2-data_summary.R
```diff
@@ -85,62 +75,51 @@ testthat::test_that("e2e: data summary table is displayed with 3 columns for dat
...
 
-    app <- TealAppDriver$new(
-      data = data,
-      modules = example_module()
-    )
+  teal.data::join_keys(data) <- teal.data::join_keys(
+    teal.data::join_key("mtcars2", "mtcars1", keys = c("am"))
+  )
+
+  app <- TealAppDriver$new(data = data, modules = example_module())
```
llrs-roche, line 99: Not sure what is happening, but I'm seeing a warning locally:
```
Warning (test-shinytest2-data_summary.R:99:3): e2e: data summary table does not list unsupported objects
'package:MultiAssayExperiment' may not be available when loading
```

tests/testthat/test-shinytest2-init.R
```diff
@@ -2,18 +2,17 @@ testthat::skip_if_not_installed("shinytest2")
...
 
 testthat::test_that("e2e: teal app initializes with no errors", {
-  testthat::skip("chromium")
   skip_if_too_deep(5)
   app <- TealAppDriver$new(
     data = simple_teal_data(),
     modules = example_module(label = "Example Module")
   )
   app$expect_no_shiny_error()
+  app$expect_screenshot(selector = "#teal-tabpanel_wrapper")
```
llrs-roche, line 11: Locally I got a warning about missing windows snapshot (Not sure how important this is)

```
Warning (test-shinytest2-init.R:11:3): e2e: teal app initializes with no errors
Adding new file snapshot: 'tests/testthat/_snaps/windows-4.5/teal-001.png'
```

tests/testthat/test-shinytest2-filter_panel.R
```diff
@@ -93,12 +85,31 @@ testthat::test_that("e2e: filtering a module-specific filter is not refected in
...
+  app <- TealAppDriver$new(data = data, modules = example_module())
+
+  # Visible by default
+  filterpanel_accordion_selector <- paste(
+    ".teal-filter-panel [id$=\\'filters-main_filter_accordian\\']",
+    "> .accordion-item > .accordion-collapse"
+  )
+  testthat::expect_true(app$is_visible(filterpanel_accordion_selector))
+  app$click(selector = paste(
+    ".teal-filter-panel [id$='filters-main_filter_accordian']",
```
llrs-roche: I fixed the typo `accordian` -> `accordion` in teal.slice regression tests but if you use an id there is no need to specify the class of the element as ids are unique. Start the selector with `#filters-main_filter_accordion` (if https://github.com/insightsengineering/teal.slice/pull/653 gets merged before this PR)

tests/testthat/test-shinytest2-module_bookmark_manager.R
```diff
@@ -56,6 +49,7 @@ testthat::test_that("bookmark_manager_button shows modal with url containing sta
 
   testthat::expect_match(
     rvest::html_text(app$get_html_rvest("div[id$=bookmark_modal] pre")),
-    "_state_id_"
+    "_state_id_=[a-zA-Z0-9]+",
```
llrs-roche: This seems a bit weird. Is this the pattern for ids? Shouldn't then the regex include `-` and `_`? If it is for ids of elements it might accept other characters like `.` or space ` `

tests/testthat/test-shinytest2-reporter.R
```diff
@@ -64,30 +52,20 @@ testthat::test_that("e2e: adding a report card in a module adds it in the report
...
     "Card comment"
   )
-
   app$click(NS(app$active_module_ns(), "reporter-add_report_card_simple-add_card_ok"))
 
+  # Check whether card was added
   app$navigate_teal_tab("Report previewer")
-
   accordian_selector <- sprintf("#%s-pcards .accordion-toggle", app$active_module_ns())
   app$click(selector = accordian_selector)
```
llrs-roche: Her locally it FAILS I'm using the main branch of the dependencies for testing this:
```
Error (test-shinytest2-reporter.R:60:3): e2e: adding a report card in a module adds it in the report previewer tab
Error in `app_find_node_id(self, private, input = input, output = output, 
    selector = selector)`: Cannot find HTML element with selector #teal-teal_modules-nav-temporary_label-module-pcards .accordion-toggle
```

tests/testthat/test-shinytest2-wunder_bar.R
```diff
@@ -35,7 +33,30 @@ testthat::test_that("wunder_bar_srv clicking snapshot icon opens snapshot-manage
...
+
+testthat::test_that("e2e: collapsing wunderbar hides filter and data summary", {
+  skip_if_too_deep(5)
+
+  data <- teal.data::teal_data(mtcars1 = mtcars, mtcars2 = data.frame(am = c(0, 1), test = c("a", "b")))
+  app <- TealAppDriver$new(data = data, modules = example_module())
+  app$click(selector = ".teal_module.active > .bslib-sidebar-layout > button.collapse-toggle")
+  testthat::expect_false(app$is_visible(".teal-filter-panel"))
+  testthat::expect_false(app$is_visible(".teal-active-data-summary-panel"))
+  app$expect_screenshot(selector = ".bslib-sidebar-layout .main")
```
llrs-roche, line 59: Missing windows snapshot too:

```
Warning (test-shinytest2-wunder_bar.R:49:3): e2e: collapsing wunderbar hides filter and data summary
Adding new file snapshot: 'tests/testthat/_snaps/windows-4.5/teal-001.png'
```

tests/testthat/test-shinytest2-show-rcode.R
```diff
@@ -1,61 +1,78 @@
...
+      app$get_text("#shiny-modal div.modal-header > h4"),
+      "Example Code"
+    )
+
+    # There are two Dismiss buttons with similar id and the same label.
+    dismiss_text <- app$get_text("#shiny-modal button[data-dismiss]")
+    testthat::expect_length(dismiss_text, 2)
+    testthat::expect_setequal(dismiss_text, "Dismiss")
+
+    # Check for Copy buttons.
```
llrs-roche, line 40: This is side comment: but on my computer the `Copy to Clipboard` button doesn't work: The clipboard doesn't collect te R code. Probably it shouldn't be tested but just letting you know in case you know an easy way to test it really works (and then I should check with local IT why it fails on my computer).

tests/testthat/test-shinytest2-teal_data_module.R
```diff
@@ -2,148 +2,72 @@ testthat::skip_if_not_installed("shinytest2")
...
   app$click("teal-data-teal_data_module-submit")
   app$click(selector = "#teal-close_teal_data_module_modal button")
-  testthat::expect_true(
-    app$get_js("document.querySelector('#teal-close_teal_data_module_modal button.disabled') === null")
+
+  testthat::expect_null(
+    app$get_html("#teal-close_teal_data_module_modal button.disabled")
   )
-  testthat::expect_true(app$get_js("document.querySelector('.teal-data-module-popup') === null"))
+  testthat::expect_null(app$get_html(".teal-data-module-popup"))
```
llrs-roche: This is already tested, no need to duplicate tests if the app is the same (Might have missed some other duplicate tests, please check if there are any).

tests/testthat/test-shinytest2-teal_data_module.R
```diff
@@ -282,37 +189,23 @@ testthat::test_that("e2e: teal_data_module gets removed after successful data lo
...
-          data <- within(
-            teal_data(),
-            {
-              dataset1 <- iris
-              dataset2 <- mtcars
-            }
-          )
-
-          data
+          data <- within(teal_data(), {
```
llrs-roche:
```suggestion
          within(teal_data(), {
```

tests/testthat/test-shinytest2-teal_data_module.R
```diff
@@ -325,74 +218,8 @@ testthat::test_that("e2e: teal_data_module is still visible after successful dat
...
-          rvest::html_nodes(
-            app$get_html_rvest("#teal-teal_modules-active_tab"),
-            "a[data-value*='example_module']"
-          ),
-          "disabled"
-        )
-      )
-    )
-  )
+  app$click(selector = "#teal-close_teal_data_module_modal button[data-dismiss='modal']")
```
llrs-roche: Not tested but if you use the id why select with attributes properties too?
```suggestion
  app$click(selector = "#teal-close_teal_data_module_modal")
```

tests/testthat/test-shinytest2-wunder_bar.R
```diff
@@ -35,7 +33,30 @@ testthat::test_that("wunder_bar_srv clicking snapshot icon opens snapshot-manage
...
+})
+
+testthat::test_that("e2e: collapsing wunderbar hides filter and data summary", {
+  skip_if_too_deep(5)
+
+  data <- teal.data::teal_data(mtcars1 = mtcars, mtcars2 = data.frame(am = c(0, 1), test = c("a", "b")))
+  app <- TealAppDriver$new(data = data, modules = example_module())
+  app$click(selector = ".teal_module.active > .bslib-sidebar-layout > button.collapse-toggle")
+  testthat::expect_false(app$is_visible(".teal-filter-panel"))
+  testthat::expect_false(app$is_visible(".teal-active-data-summary-panel"))
```
llrs-roche, line 58: I think it is best to test also that previously the panel(s) were visible. The teal filter panel could disappear and we wouldn't notice.

R/reporter_previewer_module.R
```diff
@@ -43,6 +43,7 @@ reporter_previewer_module <- function(label = "Report previewer", server_args =
   # This is to prevent another module being labeled "Report previewer".
   class(module) <- c(class(module), "teal_module_previewer")
   module$label <- label
+  module$path <- label
```
llrs-roche, line 57: Not sure what this change does, the commit message does not provide any clue and I find surprising that the same information is on two parts of the same object. Should label be in just one place?

tests/testthat/test-module_teal.R
```diff
@@ -466,7 +466,7 @@ testthat::describe("srv_teal teal_modules", {
           `teal_modules-active_module_id` = "module_1"
         )
         out <- modules_output$module_1()
-        testthat::expect_true(!is.null(out))
+        testthat::expect_type(out, "double")
```
llrs-roche, line 469: What is this magic number and how it is related to "srv_teal teal_modules are not called again when data changes"? All the previous tests expect it to be NULL?

tests/testthat/helper-shinytest2.R
```diff
@@ -35,3 +35,26 @@ report_module <- function(label = "example teal module") {
     }
   )
 }
+
+example_teal_data_module <- function(with_submit = FALSE, once = TRUE) {
```
llrs-roche: I only see it being used on `teal_data_module` did you want to reuse it in other tests too?

llrs-roche:
I checked the tests added and I only found 2 bslib regression tests added about collapsing the data summary and the filters panel (no one about collapsing filters of an individual dataset). Not sure if we need to test for other things that might be tested on teal.slice but are modified by teal or some that are teal exclusive regarding bslib.

tests/testthat/test-shinytest2-init.R
averissimo: I uploaded the snap for `linux` using the insightsengineering's rstudio container, I don't see any other window snaps, so I think we should skip this one

tests/testthat/test-shinytest2-data_summary.R
averissimo: This is a `rlang::hash` issue that I also get locally, but not on the tests.

I'm not sure about the best way to address this as it may be a conditional warning, but I'm trying to find a proper solution

tests/testthat/test-shinytest2-filter_panel.R
averissimo: Awesome, thanks!

I think this is one of the tests that should be redundant, as it looks for the main filter accordion.

tests/testthat/test-shinytest2-module_bookmark_manager.R
averissimo: I don't think it can start with those as it's based on an hex id:

https://github.com/rstudio/shiny/blob/08383ad8b978ed4add7dedcb3a16929c2610d2d8/R/shiny.R#L26

llrs-roche: Then a comment on what this `_state_id_` is would be appreciated by me and possibly future readers of the code.

tests/testthat/test-shinytest2-data_summary.R
averissimo: I invested a bunch of time on this _(not the first time it popups)_ and best solution seems to suppress warning at the `rlang::hash` call

This is not the first time it popups, with https://github.com/insightsengineering/teal.data/issues/351 and https://github.com/insightsengineering/teal.code/issues/194

As well as other times.

The best I can assertain is that `rlang::hash` tries to serialize the namespace included in the `{MAE}` dataset in C++ and it outputs this warning as it is not a base package.

tests/testthat/test-module_teal.R
averissimo: This is the output of the module (`runif(1)`) defined above.

I changed it from `not null` to an explicit type for the test to be more specific and less prone to keep passing if it starts to return something else (reactive, silent error, etc...)

R/reporter_previewer_module.R
averissimo: The reporter is created in a particular way, with a temporary label and then it is changed manually.

This affects the label, but not the path of it. The path appears as a data attribute and this change ensures that it is the same as the label and not `temporary label`

averissimo: It was introduced by the UX/UI redesing, maybe @vedhav can give more context on why it is needed?

My understanding is that is 2-fold, ensuring a unique name while keeping label intact and practically on navigation

tests/testthat/helper-shinytest2.R
averissimo: It was used several times there and reduced code. I was refraiming from creating a single `helper` file for this function, but there is no real drawback.

`testthat` encourages to keep this kind of functions/constants/etc.. on a helper file.

WDYT about creating `helper-shinytest2-teal_data_module`?

tests/testthat/test-shinytest2-show-rcode.R
averissimo: Are you saying it doesn't work on any browser or on the rstudio viewer?

tests/testthat/test-shinytest2-teal_data_module.R
averissimo: I believe it's because the button itself needs to be clicked, but I'll confirm

tests/testthat/test-shinytest2-show-rcode.R
llrs-roche: I can't install Firefox to test it but on Rstudio viewer I couldn't, and on Chrome I sometimes can't and sometimes I can (just tested on teal.gallery with _stable and _dev and I could in both). But marking as resolved as it might be a RStudio issue

tests/testthat/test-shinytest2-filter_panel.R
llrs-roche: Mmh, so should it be removed from here?

tests/testthat/test-shinytest2-data_summary.R
llrs-roche: I'll close this comment (and maybe document your finding on one of these issues).

R/reporter_previewer_module.R
llrs-roche: Oh yes, now I recall some comments. Nothing to do then (except a comment to document this?)

tests/testthat/helper-shinytest2.R
llrs-roche: As Dawid commented recently, it would make the tester to chase to multiples files to get a picture of what is being tested. I prefer to have it here or on the file it is used (which is why I pointed out that it wasn't reused elsewhere). I'll leave it up to you but resolving this.

tests/testthat/test-shinytest2-filter_panel.R
averissimo: I meant that it should be also here. Regardless, of the tests on `teal.slice`.

As here it looks for its presence and whether it can be collapsed

tests/testthat/test-shinytest2-show-rcode.R
averissimo: `rstudio` doesn't allow for this to work, regardless of the operating system.

Good to know that you tested on the `teal.gallery`, those are the "truth" :-)

tests/testthat/helper-shinytest2.R
averissimo: Ok, unless I can find a use case opportunity, I'll move it back to the file.

Keeping the comment unresolved until it's commited

R/reporter_previewer_module.R
averissimo: Comment then! For both lines :sweat_smile:  (I updated the comment above that chunk) cfbfd844d

tests/testthat/test-shinytest2-teal_data_module.R
averissimo: We need the `button` in the selector, but good catch on not needing the `data-dismiss`. We should only have 1 button there.

```suggestion
  app$click(selector = "#teal-close_teal_data_module_modal button")
```

tests/testthat/test-shinytest2-reporter.R
averissimo: Can you try again? The `accordian` -> `accordion` typo was corrected here.

Maybe you had the `teal.slice` branch loaded? I'm hoping it's that, because I've run this multiple times in a row and found no problems :see_no_evil:

llrs-roche: I tested again and it failed with the same error message. But I think is because I am using `main`.

tests/testthat/test-shinytest2-wunder_bar.R
averissimo: I agree, but it should live on its own test.

Added 1 test above with d86b5141a

tests/testthat/test-shinytest2-teal_data_module.R
averissimo: It is, great eye!! I further improved so that the same selector is used instead of duplicating the string

tests/testthat/test-shinytest2-filter_panel.R
```diff
@@ -93,12 +85,31 @@ testthat::test_that("e2e: filtering a module-specific filter is not refected in
...
+testthat::test_that("e2e: filter panel UI can be collpased and expanded (`bslib` regression)", {
+  skip_if_too_deep(5)
+
+  data <- teal.data::teal_data(mtcars1 = mtcars, mtcars2 = data.frame(am = c(0, 1), test = c("a", "b")))
+  app <- TealAppDriver$new(data = data, modules = example_module())
+
+  # Visible by default
+  filterpanel_accordion_selector <- paste(
+    ".teal-filter-panel [id$=\\'filters-main_filter_accordion\\']",
+    "> .accordion-item > .accordion-collapse"
```
llrs-roche: Selector could be simplified:
```suggestion
    "#filters-main_filter_accordion > .accordion-item > .accordion-collapse"
```

tests/testthat/test-shinytest2-filter_panel.R
```diff
@@ -93,12 +85,31 @@ testthat::test_that("e2e: filtering a module-specific filter is not refected in
...
+  )
+  testthat::expect_true(app$is_visible(filterpanel_accordion_selector))
+  app$click(selector = paste(
+    ".teal-filter-panel [id$='filters-main_filter_accordion']",
+    "> .accordion-item > .accordion-header button"
```
llrs-roche:
```suggestion
    "#filters-main_filter_accordion > .accordion-item > .accordion-header button"
```

DESCRIPTION
```diff
@@ -81,7 +81,7 @@ RdMacros:
     lifecycle
 Remotes:
     insightsengineering/teal.reporter@main,
-    insightsengineering/teal.slice@main,
+    insightsengineering/teal.slice@651_regr_test@main,
```
averissimo: Works as a blocking comment for PR merge until https://github.com/insightsengineering/teal.slice/pull/653 is merged

```suggestion
    insightsengineering/teal.slice@main,
```

tests/testthat/test-shinytest2-filter_panel.R
averissimo: That `id` is not valid, the full id of the element is `"#teal-teal_modules-nav-example_teal_module-filter_panel-filters-main_filter_accordion"`

I've deliberately used the `[id$='...']` just to match the termination of the id, as it becomes very hardcoded and the test and sensitive to `teal` HTML namespace.

@llrs-roche Do you think we should use it fully?

To simplify it a bit, we could also remove the `.teal-filter-panel` prefix

The reason for including in the first place is to have a full match in the selector (to complement the partial id match)

llrs-roche: Sorry, realized after I made the comments that here the selector used `$=` and `\\` to scape the quotes but I did't look into it deeper.  
I think it is good to keep full id as it is easier to reason. All the nested structure (`>... > ...` ) is already very sensitive to teal structure so I don't think that depending on the namespace would be worse. 
In other tests there was a helper to create the id within the namespace 

IMHO all or most of the selector needed should have an id (or at most on the parent div).

DESCRIPTION
llrs-roche: PR on teal.slice is merged. Revert to main and merge?

tests/testthat/test-shinytest2-filter_panel.R
averissimo: Check the latest commit: a8ce7e6fe

```diff
diff --git a/tests/testthat/test-shinytest2-filter_panel.R b/tests/testthat/test-shinytest2-filter_panel.R
index c38d236ad..592e2fd95 100644
--- a/tests/testthat/test-shinytest2-filter_panel.R
+++ b/tests/testthat/test-shinytest2-filter_panel.R
@@ -100,15 +100,10 @@ testthat::test_that("e2e: filter panel UI can be collpased and expanded (`bslib`
   app <- TealAppDriver$new(data = data, modules = example_module())
 
   # Visible by default
-  filterpanel_accordion_selector <- paste(
-    ".teal-filter-panel [id$=\\'filters-main_filter_accordion\\']",
-    "> .accordion-item > .accordion-collapse"
-  )
+  filter_panel_id <- "#teal-teal_modules-nav-example_teal_module-filter_panel-filters-main_filter_accordion"
+  filterpanel_accordion_selector <- paste(filter_panel_id, "> .accordion-item > .accordion-collapse")
   testthat::expect_true(app$is_visible(filterpanel_accordion_selector))
-  app$click(selector = paste(
-    ".teal-filter-panel [id$='filters-main_filter_accordion']",
-    "> .accordion-item > .accordion-header button"
-  ))
+  app$click(selector = paste(filter_panel_id, "> .accordion-item > .accordion-header button"))
   testthat::expect_false(app$is_visible(filterpanel_accordion_selector))
 
   app$stop()
```

Issue #1539, opened 2025-05-28
[Question]: When to show "Report Previewer" tab?


gogonzo:
### What is your question?

Teal adds "Report Previewer" only when it is needed, currently on main it is a following condition:
- When any of the module(s) have a `reporter` argument in the formals

With the new design, none of none of the `teal_module`(s) has `report` argument as it is not needed. Now, teal_module returns reactive containing teal_data. This indicates that module is reportable. This means that in `teal` we need to detect whether any of the module returns a reactive containing `teal_data`. The point of the problem is that we DON'T CALL A MODULE UNTIL ITS TAB IS CLICKED. It means that "Report Previewer" might not been shown until the first "reportable" module is clicked.

I think we need a better idea how to handle this. Here is some primitive mockup:

<img width="988" alt="Image" src="https://github.com/user-attachments/assets/5c774422-ad01-4708-9f04-7562771b97ec" />

 

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [x] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [x] I agree to follow this project's Security Policy.

donyunardi:
Initial thoughts:

* crawl through the code returned by `module()`.
  If we're consistent with the naming convention, maybe we can crawl through the returned module() code before being evaluated and look for specific pattern. I'm not too keen on this because it relies on how user write their custom module but just want to mention it.

* Leave the decision to show the "Report Previewer" module to user.
  ```r
    app <- init() |> show_report_previewer(TRUE)
  ```
  The default is to always show the report previewer. If none of the modules is reportable, then this module will show no cards. We can then encourage user to hide it by setting it to FALSE.

* Maybe it's okay that "Report Previewer" is not shown in the beginning until there is something to preview. 

Will think some more.

gogonzo:
> * crawl through the code returned by `module()`.
>   If we're consistent with the naming convention, maybe we can crawl through the returned module() code before being evaluated and look for specific pattern. I'm not too keen on this because it relies on how user write their custom module but just want to mention it.

Not that easy and robust naming convention is something we can follow, not module-developers out there. We would have to parse for the possible `return` in the middle of the `moduleServer` code. Also if there is a function call at the end which returns modified `teal_report` then it will be very hard to parse recursively. Static code analysis doesn't guarantee anything.



> * Leave the decision to show the "Report Previewer" module to user.
[...]

PR #1562, opened 2025-07-14
Redesign and change position of reporter options

vedhav:
Companion of https://github.com/insightsengineering/teal.reporter/pull/360
Closes https://github.com/insightsengineering/teal/issues/1539

In this PR:
- reporter menu accessible as a dropdown with preview, downlaod, load and reset buttons
- reporter_previewer_module to be deprecated and download args like `global_knitr` `rmd_yaml_args`, `rmd_output` etc. will be controlled by options 
- Introduced `.teal_navbar` to create a navigation bar and `.teal_navbar_menu` to create a dropdown menu.

### DIFF
--- a/R/reporter_previewer_module.R
+++ b/R/reporter_previewer_module.R
@@ -24,7 +24,7 @@ reporter_previewer_module <- function(label = "Report previewer", server_args = 
   checkmate::assert_true(all(names(server_args) %in% names(formals(teal.reporter::reporter_previewer_srv))))
 
   lifecycle::deprecate_soft(
-    when = "",
+    when = "1.0.0",
     what = "reporter_previewer_module()",
     details = paste(
       "Please use `options()` to customize the reporter options:\n",
--- a/R/init.R
+++ b/R/init.R
@@ -219,6 +219,7 @@ init <- function(data,
           ),
           tags$header(
             id = "teal-header",
+            style = "margin: 0.5em",
             tags$div(id = "teal-header-content")
           ),
           ui_teal(
@@ -227,6 +228,7 @@ init <- function(data,
           ),
           tags$footer(
             id = "teal-footer",
+            style = "margin: 0.5em",
             tags$div(id = "teal-footer-content"),
             ui_session_info("teal-footer-session_info")
           )


averissimo:
Not sure if it fits here, but the `module$path` of reporter is not updated so the `data-value` is set to `"temporary label"`

image

I'm applying the following patch on #1559 as this is picked by e2e tests, but it also fits in this PR

```diff
diff --git a/R/reporter_previewer_module.R b/R/reporter_previewer_module.R
index 19b707b34..758c1736c 100644
--- a/R/reporter_previewer_module.R
+++ b/R/reporter_previewer_module.R
@@ -43,6 +43,7 @@ reporter_previewer_module <- function(label = "Report previewer", server_args =
   # This is to prevent another module being labeled "Report previewer".
   class(module) <- c(class(module), "teal_module_previewer")
   module$label <- label
+  module$path <- label
   attr(module, "teal_bookmarkable") <- TRUE
   module
 }
```

R/module_teal.R
```diff
@@ -103,6 +102,39 @@ srv_teal <- function(id, data, modules, filter = teal_slices()) {
       shinyjs::showLog()
     }
 
+    insertUI(
```
gogonzo: why to `insertUI` when elements don't use any objects created in the server. I think this can be moved to UI directly

vedhav: These ui elements need to be placed in a different level than the current shiny module where they are called. The need to be placed inside the `ui_teal_modules_nav()` but since it is called in a different shiny module we are inserting by referencing the selector of that element.

NAMESPACE
```diff
@@ -1,19 +1,17 @@
 # Generated by roxygen2: do not edit by hand
 
+S3method(.srv_teal_module,default)
```
gogonzo, line 3: @averissimo please have a look here. I've prefixed these functions as there is an extra `ui/srv_teal_module` called before, which creates a navigation bar etc.
I wonder if we can remove this from the package NAMESPACE but still have this as a S3 method internally?

NEWS.md
```diff
@@ -1,5 +1,20 @@
...
+    - `teal.reporter.rmd_outputs` to customize the R Markdown outputs types for the report.
+    - `teal.reporter.rmd_yaml_args` to customize the widget inputs in the download report modal.
+    - `teal.reporter.knitr_global` to customize the global `knitr` options for the report.
+
+### Enhancement
+
+* General repositioning of key navigation components across the app:
+    - Modules: The module navigation is moved from a nested tab selection to a "Module" drop-down selection. The module selection can be done from the nested button of modules.
+    - Reporter: The Report previewer is no longer displayed as yet another teal module, it is placed inside a "Report" drop-down right next to the "Module" drop-down. The Report previewer is a modal that displays the added report cards. The "Report" drop-down also contains other global report options like download/load/reset the Report.
+    - Teal Data Module: The UI created by the `teal_data_module()` is now placed inside a modal that can only be closed when the `teal_data` is resolved.
```
gogonzo:
```suggestion
    - Teal Data Module: The UI created by the `teal_data_module()` is now placed inside a modal that can only be closed when it returns `teal_data`.
```

NAMESPACE
llrs-roche: Even if S3 methods are only to be used internally they should be on the NAMESPACE file. See [WRE](https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#Registering-S3-methods-1)

NEWS.md
```diff
@@ -1,5 +1,20 @@
 # teal 0.16.0.9015
 
+### Breaking changes
+
+* The `reporter_previewer_module()` is deprecated and will be removed in the future release. The custom `server_args` values can be set using `options()`:
```
llrs-roche, line 5: Should NEWS have a link at the end to the issue or PR that provides this? Something like `... in the future release (#1562).`.
Same for the enhancement. It will help if down the line we need to check why we took some decisions.

R/module_teal.R
```diff
@@ -158,46 +187,44 @@ srv_teal <- function(id, data, modules, filter = teal_slices()) {
...
+              class = "teal teal-data-module-popup",
+              modalDialog(
+                id = session$ns("teal_data_module_ui"),
+                size = "xl",
+                tags$div(
+                  ui_init_data(session$ns("data")),
+                  validate_ui
+                ),
+                easyClose = FALSE,
+                footer = tags$div(id = session$ns("close_teal_data_module_modal"), modalButton("Dismiss"))
```
llrs-roche, line 215: Didn't we decide to change the Dismiss text  to something else on the delayed data modal to make it more in line with the other modals? I remember that the `easyClose` should remain FALSE but I thought we decided to rename the button.

NAMESPACE
gogonzo: Yes, but it is for a generic which is outside of the package. Generic which is in base for example, uses `UseMethod` and won't be able to locate sub-methods if they are not exported. 

> The standard method for S3-style UseMethod dispatching might fail to locate methods defined in a package that is imported but not attached to the search path

To skip the namespace it is possible to use `registerS3method("foo", "teal_module",  foo.teal_module)` or `#' @method` instead

R/reporter_previewer_module.R
```diff
@@ -1,29 +1,40 @@
...
 #'
 #' @export
-#'
 reporter_previewer_module <- function(label = "Report previewer", server_args = list()) {
   checkmate::assert_string(label)
   checkmate::assert_list(server_args, names = "named")
   checkmate::assert_true(all(names(server_args) %in% names(formals(teal.reporter::reporter_previewer_srv))))
 
+  lifecycle::deprecate_soft(
+    when = "",
```
llrs-roche: Shouldn't that be:
```suggestion
    when = "1.0.0",
```
To signal it will be deprecated on this release?

NAMESPACE
averissimo: What was the problem?

The app still runs when I remove the `#' @export`  tag of  `.srv_teal_module` and `.ui_teal_module`

IIRC the other tags are just "informative"

I have:

- Change code and run `devtools::document()`
- Installed the packages locally
- Restarted the R session
- Run 2 example apps

Part of the diff below

```diff
diff --git a/NAMESPACE b/NAMESPACE
index a9563f1a8..1dbdc26e6 100644
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -1,11 +1,5 @@
 # Generated by roxygen2: do not edit by hand
 
-S3method(.srv_teal_module,default)
-S3method(.srv_teal_module,teal_module)
-S3method(.srv_teal_module,teal_modules)
-S3method(.ui_teal_module,default)
-S3method(.ui_teal_module,teal_module)
-S3method(.ui_teal_module,teal_modules)
 S3method(c,teal_slices)
 S3method(format,teal_module)
 S3method(format,teal_modules)
diff --git a/R/module_nested_tabs.R b/R/module_nested_tabs.R
index 5ecf07dbb..3d1f4119f 100644
--- a/R/module_nested_tabs.R
+++ b/R/module_nested_tabs.R
@@ -158,13 +158,11 @@ srv_teal_module <- function(id,
 }
 
 #' @rdname module_teal_module
-#' @export
 .ui_teal_module.default <- function(id, modules, active_module_id) {
   stop("Modules class not supported: ", paste(class(modules), collapse = " "))
 }
```


https://github.com/user-attachments/assets/24be303b-6ace-4c26-9651-02d7c373d258

averissimo: Using `R 4.5.1`

gogonzo: This is what I get after running exploratory app

image

llrs-roche: This function `registerS3method` (and others) is documented as:
>  Internal namespace support functions. Not intended to be called directly, and only visible because of the special nature of the base namespace.

Not saying it doesn't work ...

averissimo: Conclusion is to merge it as it is, and if we find a solution that works for all, we can clean the `NAMESPACE` in the future.

I'm resolving this comment.

NEWS.md
vedhav: Thanks Lluis, I'll add it.

R/module_teal.R
vedhav: We debated over "Dismiss" or "Close" and decided to go with "Dismiss" - It is also the [default footer by modalDialog](https://shiny.posit.co/r/reference/shiny/latest/modaldialog.html).
Only in this case we have `easyClose = FALSE` because we want to explicitly wait for the user to resolve the data before closing it.

R/reporter_previewer_module.R
vedhav: Good catch, I think `teal.reporter` also needs a version. We have this in the checklist before the CRAN release so we know for sure the version we need to put here.

PR #1568, opened 2025-07-23
refurnish

gogonzo:


R/module_teal.R
```diff
@@ -114,12 +97,7 @@ ui_teal <- function(id, modules) {
     ui_snapshot_manager_panel(ns("snapshot_manager_panel")),
     ui_filter_manager_panel(ns("filter_manager_panel"))
   )
-
-  navbar <- .teal_navbar(
-    id = ns("active_module_id"),
-    nav_items = nav_elements,
-    tab_content = module_items$tab_content
-  )
+  navbar <- .teal_navbar_append(navbar, nav_elements)
```
gogonzo, line 105: @vedhav this is the most important change in this commit. `navbar` is created in `ui_teal_module` instead of here and child is appended instead.

R/module_teal.R
```diff
@@ -206,10 +184,8 @@ srv_teal <- function(id, data, modules, filter = teal_slices()) {
 
     if (inherits(data, "teal_data_module")) {
       setBookmarkExclude(c("teal_data_module_ui", "open_teal_data_module_ui"))
-      insertUI(
-        selector = c(".teal-modules-wrapper .nav-item-custom"),
-        where = "beforeBegin",
-        .expand_button(
+      .teal_navbar_insert_ui(
```
gogonzo, line 192: introduced a wrapper to make sure that it inserts in the correct place.

R/module_nested_tabs.R
```diff
@@ -59,23 +62,112 @@
 #' @keywords internal
 NULL
 
-#' @rdname module_teal_modules_nav
-ui_teal_module <- function(id, modules, active_module_id) {
+
+#' @rdname module_teal_module
+ui_teal_module <- function(id, modules) {
```
gogonzo, line 67: This is a wrapper module which then calls modules recursively to obtain nav-link(s) and tab-pane(s) with consistent it

PR #1569, opened 2025-07-24
Update gif instead of shinylive

llrs-roche:
# Pull Request

Replacement for https://github.com/insightsengineering/teal/pull/1490

Tried to create a `.gif` but it either was too small on the screen or it weighted much more than the video.
The raw `.mp4` is not displayed on the github branch so I went with the `.webp` that was previously used.


vedhav (approved):
LGTM

PR #1573, opened 2025-08-05
Remove extra space when the data tab is removed when `once = TRUE`

vedhav:
When using a `teal_data_module` as a data source and having `once = TRUE` to remove the ui after the data is resolved we need to remove the data load tab. Right now we remove the inner button and the span is still present so the flex gap adds some extra space that looks inconsistent. In this PR we remove the complete span element so the spacing is just the same as a non-teal_data_module app.

### Before
Screenshot 2025-08-05 105313

### After
image

Example app for testing:

```r
devtools::load_all("../teal")

tdm_resolve_once <- teal_data_module(
  ui = function(id) {
    fluidPage(
      numericInput(NS(id, "iris_rows"), "iris rows", min = 0, max = 150, step = 1, value = 10),
      actionButton(NS(id, "submit"), "Submit")
    )
  },
  server = function(id, ...) {
      eventReactive(input$submit, {
        teal_data(iris = head(iris, input$iris_rows), mtcars = mtcars)
      })
    })
  },
  once = TRUE
)

tdm_resolve_not_once <- teal_data_module(
  ui = function(id) {
    fluidPage(
      numericInput(NS(id, "iris_rows"), "iris rows", min = 0, max = 150, step = 1, value = 10),
      actionButton(NS(id, "submit"), "Submit")
    )
  },
  server = function(id, ...) {
      eventReactive(input$submit, {
        teal_data(iris = head(iris, input$iris_rows), mtcars = mtcars)
      })
    })
  },
  once = FALSE
)

app <- init(
  data = tdm_resolve_once, # Getting extra gap when using `once = TRUE`
  # data = tdm_resolve_not_once,
  modules = example_module()
)

shinyApp(app$ui, app$server)
```


averissimo (dismissed):
~Shouldn't the `data` tab stay if parameter is `ONCE = FALSE`?~

~It seems that the previous implementation as well as this PR doesn't support keeping the tab.~

~https://github.com/user-attachments/assets/65dcfc7e-1809-40c2-8991-d4d1183b591d~

vedhav:
@averissimo please check that there are two `teal_data_module` objects: `tdm_resolve_not_once` and `tdm_resolve_once` defined and one of them is commented.

R/module_teal.R
```diff
@@ -234,7 +234,12 @@ srv_teal <- function(id, data, modules, filter = teal_slices()) {
               session$ns("close_teal_data_module_modal")
             )
           )
-          removeUI(sprintf("#%s", session$ns("open_teal_data_module_ui")))
+          shinyjs::runjs(
+            sprintf(
+              "document.querySelector('#%s')?.parentElement.remove();",
+              session$ns("open_teal_data_module_ui")
+            )
+          )
```
averissimo: Crawling to remove the element might come to haunt us in the future

Did you consider having an explicit way of removing the `.expand_button()` UI, such as using `data-for` attribute or something similar? (I'm taking inspiration from the `<label for="xxx">` in forms)

I think this could be done with shiny alone, WDYT?

```diff
diff --git a/R/module_teal.R b/R/module_teal.R
index b05360ecf..06bce5903 100644
--- a/R/module_teal.R
+++ b/R/module_teal.R
@@ -228,18 +228,8 @@ srv_teal <- function(id, data, modules, filter = teal_slices()) {
         observeEvent(data_signatured(), once = TRUE, {
           logger::log_debug("srv_teal@2 removing data tab.")
           # when once = TRUE we pull data once and then remove data tab
           shinyjs::runjs(
             sprintf(
               "document.querySelector('#%s button').click();",
               session$ns("close_teal_data_module_modal")
             )
           )
-          shinyjs::runjs(
-            sprintf(
-              "document.querySelector('#%s')?.parentElement.remove();",
-              session$ns("open_teal_data_module_ui")
-            )
-          )
+          shiny::removeUI(sprintf("[data-for='%s']", session$ns("open_teal_data_module_ui")))
         })
       }
     } else {
diff --git a/R/utils.R b/R/utils.R
index 0528d8284..707218301 100644
--- a/R/utils.R
+++ b/R/utils.R
@@ -514,6 +514,7 @@ pluralize <- function(x, singular, plural = NULL) {
 .expand_button <- function(id, label, icon) {
   tags$span(
     class = "teal expand-button",
+    "data-for" = id,
     htmltools::htmlDependency(
       name = "teal-busy-disable",
       version = utils::packageVersion("teal"),
```

averissimo (commented):
Quick question

R/module_teal.R
vedhav: I think it was just a preference to use runjs as it was quick and easy for me to test in browser and implement here.
But, if you think we should have `shiny::removeUI` we can use it without changing `.expand-button()` too.

```suggestion
          shiny::removeUI(
            selector = sprintf(
              ".teal.expand-button:has(#%s)",
              session$ns("teal_data_module_ui")
            )
          )
```

vedhav: And, I like your suggstion to use `shiny::removeModal` instead of clicking the dismiss button using selector in the line above. I will commit it, tested that it works fine.

averissimo (approved):
:+1: awesome and simple

PR #1576, opened 2025-08-12
obtain a named list from environment

gogonzo:
On teal.code@main `get_var` is hard-deprecated. Here is the easier way to obtain qenv objects as a list


gogonzo:
@chizapoth get_var is fixed here. Depending if you use `teal@main` (changes available instantly) from github directly or r-universe/pharmaverse repository (we need to wait for a built https://pharmaverse.r-universe.dev/builds)

PR #1592, opened 2025-09-10
Add show r code to main ui of teal

### LINKED ISSUE #1534: [Research] Automatic Show R Code button creation
### Feature description

This issue is related with https://github.com/insightsengineering/teal/issues/1526

If a module can return teal_data object, then we might be able to automate the creation of Show R Code button. 

### Code of Conduct

### Contribution Guidelines

### Security Policy

averissimo:
# Pull Request

- Fixes https://github.com/insightsengineering/teal/issues/1534
- Companion of:
  - https://github.com/insightsengineering/teal/pull/1592
  - https://github.com/insightsengineering/teal.widgets/pull/319
  - https://github.com/insightsengineering/teal.reporter/pull/401

### Changes description

- Adds "Show R code" to main teal UI
- Supports `disable_show_r_code()` function that is applied to modules to hide it
- Adds support to hide "+ Add to Report" without affecting this new functionality
  - It previously changes the return object to `NULL` fully disabling any post-module functionality
- Adds new global option to disable the "Show R Code" button
  - `+ Add to Report` doesn't need it as it can be disabled in `init`

image

### Example app

```r
# Show R Code PR example app

# - teal_reporter: 1534-show_r_code@main
# - teal: 1534-show_r_code@main
pkgload::load_all("../teal.logger")
pkgload::load_all("../teal.widgets")
pkgload::load_all("../teal.transform")
pkgload::load_all("../teal")

library(teal.logger)
library(teal.code)
library(teal.data)
library(teal.reporter)
library(teal.slice)
library(teal.widgets)
library(teal.transform)
library(teal)

do.call(
  options,
  rlang::list2(
    "teal.bs_theme" = bslib::bs_theme(version = "5"),
    teal.log_level = "TRACE",
    teal.show_js_log = TRUE,
    teal.show_r_code = TRUE,
  )
)

example_extended <- function(label = "example teal module",
                             datanames = "all",
                             transformators = list(),
                             decorators = list()) {
  checkmate::assert_list(decorators, "teal_transform_module")

  mod <- example_module(label, datanames, transformators, decorators)

  module(
    label,
    server = function(id, data, decorators) {
        result <- mod$server("example", data, decorators)

        reactive({
          q <- result()
          attr(q, "teal.enable_src") <- FALSE
          q
        })
      })
    },
    ui = function(id, decorators) mod$ui(shiny::NS(id, "example"), decorators),
    ui_args = mod$ui_args,
    server_args = mod$server_args,
    datanames = mod$datanames,
    transformators = mod$transformators
  )
}

teal::init(
  data = within(teal_data(), {
    iris <- iris
    require(nestcolor)
    CO2 <- CO2
  }),
  modules = modules(
    example_module(label = "Module (full)"),
    example_extended(label = "📓 Module (manual)"),
    example_module(label = "📓 Module (pipe disable show r code)") |> 
      disable_src(),
    example_module(label = "🗒️ Module (pipe disable report)") |> 
      disable_report(),
    example_module(label = "🗒️📓 Module (pipe disable both)") |> 
      disable_src() |> 
      disable_report()
  )
) |> shiny::runApp()
```


R/source_code.R
```diff
@@ -0,0 +1,85 @@
...
+      teal_data_handled <- tryCatch(mod_out_r(), error = function(e) e)
+      tcard <- if (inherits(teal_data_handled, "qenv")) {
+        teal.code::get_code(teal_data_handled)
+      }
+    })
+
+    is_button_showed_r <- shiny::reactive({
+      getOption("teal.show_r_code_ui", TRUE) &&
+        inherits(mod_out_r(), "qenv") &&
+        !isFALSE(attr(mod_out_r()@code, "teal.show_r_code")) # Only hide when value is explicitly FALSE
```
m7pr: Just double checking, so there is a global option set via options called `teal.show_r_code_ui` to disable globally, and a local attribute of the code `teal.show_r_code` to disable locally?

It would be the best if we could get similar behavior for both Show R Code and Add to the Reporter.

Maybe just:
- disable_report/disable_show_r_code for local disable
- init(reporter = NULL / init(show_r_code = FALSE) for global disable
?

And then forget about `attr(mod_out_r()@code, "teal.show_r_code")`?

R/teal_data_utils.R
```diff
@@ -42,7 +42,10 @@ NULL
 }
 
 #' @rdname teal_data_utilities
-.collapse_subsequent_chunks <- function(report) {
+.collapse_subsequent_chunks <- function(card) {
```
m7pr, line 45: Is this related to show R code or some adjustment/upgrade of the reporter part?

R/teal_reporter.R
```diff
@@ -167,7 +173,7 @@ srv_add_reporter <- function(id, module_out, reporter) {
 disable_report <- function(x) {
   checkmate::assert_class(x, "teal_module")
   after(x, server = function(data) {
-    teal.reporter::teal_card(data) <- teal.reporter::teal_card()
-    NULL
+    attr(teal.reporter::teal_card(data), "teal.show_report") <- TRUE
+    data
```
m7pr: ah, so it now operates on the attribute and returns the data so that it can be used in another piped function?

vignettes/teal-options.Rmd
```diff
@@ -156,6 +156,12 @@ list(
 )
 ```
 
+### `teal.show_r_code` (`logical`)
```
m7pr: I'm leaning towards of including this as an `init` parameter.

R/source_code.R
```diff
@@ -0,0 +1,85 @@
...
+#' app <- init(
+#'   data = within(teal_data(), iris <- iris),
+#'   modules = modules(
+#'     example_module(label = "example teal module") |> disable_show_r_code()
+#'   )
+#' )
+#' if (interactive()) {
+#'   shinyApp(app$ui, app$server)
+#' }
+disable_show_r_code <- function(x) {
```
gogonzo: What do you think about calling this "src". In base `src` refers to source code (e.g. `srcfile`, `srcref`)
```suggestion
disable_src <- function(x) {
```

R/source_code.R
```diff
@@ -0,0 +1,85 @@
...
+  checkmate::assert_class(x, "teal_module")
+  after(x, server = function(data) {
+    attr(data@code, "teal.show_r_code") <- FALSE
+    data
+  })
+}
```
gogonzo: Proposition how to disable globally.

 
```suggestion
  useMethod("disable_src")
}

disable_src.teal_modules <- function(x) {
  lapply(x$children, function(child) {
    disable_src(child)
  })
}

disable_src.teal_module <- function(x) {
  checkmate::assert_class(x, "teal_module")
  after(x, server = function(data) {
    attr(data@code, "teal.show_r_code") <- FALSE
    data
  })
}

```

Alternatively, I thought that `after` could be an `S3` so that `disable_report` can also disable globally for `modules(...) |> disable_report()`

inst/css/custom.css
```diff
@@ -134,8 +134,14 @@ body > div:has(~ #shiny-modal-wrapper .blur_background) {
...
+  color: var(--bs-primary);
+  border: solid 1px var(--bs-primary);
+  padding: 0.3em;
+  border-radius: 5px;
+  background: transparent;
 }
+
+.outline-button.primary:hover:not(:disabled) {
+  background: var(--bs-primary-bg-subtle);
+}
```
gogonzo:
```suggestion
}

```

R/source_code.R
averissimo: I believe we want to keep `init` arguments minimal, hence why I used an option

R/teal_data_utils.R
averissimo: The argument name is misleading, it's a small cleanup

R/teal_reporter.R
averissimo: Yup!

R/source_code.R
averissimo: Sounds good, I matched the terminology we use in the UI, but this is more in line with R

vignettes/teal-options.Rmd
averissimo: I thought our goal with init was to keep it simple and minimal.

Checkout @gogonzo suggestion below.

R/source_code.R
averissimo: Interesting proposal.

We avoided this with datanames because of the order of operations, here it would just override the disabling as we do not support the `enable_XXXXX`

I like the alternate proposal of `after` being an S3, it would make the whole thing cleaner

R/teal_data_utils.R
averissimo: Given the function is changed

R/source_code.R
averissimo: However, I would name it as `disable_show_scr` to be a bit more explicit.

WDYT?

gogonzo: I would prefer `disable_show_rcode`, `disable_show_r_code` if  `disable` and `show` together doesn't sound weird. Some could ask whether it disables or hides?

I like `disable_src` (as an acronym of Show-R-Code), WDYT?

averissimo:
Applied the feedback and updated the testing app.

image

R/source_code.R
averissimo: All good! Made the change

R/source_code.R
averissimo: @gogonzo see the latest commits that implements the recursive `after` _(sample app has nested modules)_

inst/css/custom.css
```diff
@@ -134,8 +134,14 @@ body > div:has(~ #shiny-modal-wrapper .blur_background) {
   background: var(--bs-primary-bg-subtle);
 }
 
-.teal.add-reporter-container {
-  display: flex;
-  justify-content: flex-end;
-  padding: 10px;
+.outline-button.primary {
```
gogonzo: This will be very hard to track in the future. Here the is the button created using private function in `teal.reporter`. Can we instead change a style of the container instead?

R/source_code.R
m7pr: @averissimo getting back to this

> I believe we want to keep init arguments minimal, hence why I used an option

Even though we want to keep them minimal, the show R code (managing code reproducibility) is as big feature as reporter, so it could benefit it's own parameter.

From another hand, you could also pass `reporter` object as an option, and remove it from the `init`.

But right now we are mixing the session option and the function parameter for very similar usecase (even though show r code is a bit smaller)

R/source_code.R
m7pr: also a fan of `disable_src`

R/module_source_code.R
```diff
@@ -0,0 +1,85 @@
...
+#'   shinyApp(app$ui, app$server)
+#' }
+#' @export
+disable_src <- function(x) {
+  checkmate::assert_multi_class(x, c("teal_module", "teal_modules"))
+  after(x, server = function(data) {
+    attr(data@code, "teal.show_src") <- FALSE
+    data
+  })
+}
```
m7pr, line 113: @averissimo should this be done as `after`? so it's dispatched for teal_modules that lapply over a dispatch for teal_module like suggested in here https://github.com/insightsengineering/teal/pull/1592#discussion_r2338768547?

Or we don't need to implement `disable_src` like this, because `after` already handles the dispatches?

llrs-roche:
I was using this PR to prepare a vignette on teal and I noticed that the button on the Encoding panel is still shown:

image

I guess this button will be automatically removed or it will require modifications on all modules? Sorry I didn't saw any comment/feedback on this.

inst/css/custom.css
averissimo: Agree, I'm modifying

R/module_source_code.R
averissimo: The idea is that `after` already handles the dispatch so that we can create "simpler" functions that contain the functionality

inst/css/custom.css
averissimo: I'm convinced we need the rule, as such I made the rule more specific to `teal`.

Hardcoded `:hover:not(:disabled)` is not possible though.

If we want to have some consistent UI AND for each package to stand on its own, we need this repetition. Same reasoing to why `disable-busy` is repeated..

R/source_code.R
averissimo: I think this is acceptable because we want to encourage reproducible code on `teal`, hence hiding this global option doesn't seem strange for me.

A similar approach as `teal.lockfile.mode` option

We could not allow a method of disabling it globally, given the helper `disable_src` is available.

averissimo: I'm not completely opposed to adding it to `init`, it just feels a very small option to deserve first-class access to that function

R/source_code.R
```diff
@@ -0,0 +1,85 @@
...
+    )
+  )
+}
+
+#' @noRd
+srv_source_code <- function(id, module_out) {
+  moduleServer(id, function(input, output, session) {
+    mod_out_r <- reactive({
+      req(module_out)
+      if (is.reactive(module_out)) {
```
gogonzo:
```suggestion
      if (is.reactive(module_out)) {
        req(module_out())
```

R/source_code.R
```diff
@@ -0,0 +1,85 @@
...
+        module_out()
+      }
+    })
+
+    doc_out <- reactive({
+      req(mod_out_r())
+      teal_data_handled <- tryCatch(mod_out_r(), error = function(e) e)
+      tcard <- if (inherits(teal_data_handled, "qenv")) {
```
gogonzo:
```suggestion
      if (inherits(teal_data_handled, "qenv")) {
```

R/source_code.R
```diff
@@ -0,0 +1,85 @@
...
+        teal.code::get_code(teal_data_handled)
+      }
+    })
+
+    is_button_showed_r <- shiny::reactive({
+      getOption("teal.show_src", TRUE) &&
+        inherits(mod_out_r(), "qenv") &&
```
gogonzo:
```suggestion
        inherits(!is.null(doc_out()) &&
```

R/source_code.R
```diff
@@ -0,0 +1,85 @@
...
+          ui_source_button(session$ns(NULL))
+        }
+      })
+      teal.widgets::verbatim_popup_srv(
+        id = "source_code", verbatim_content = doc_out, title = "Show R Code for Response"
+      )
+    })
+    observeEvent(doc_out(), ignoreNULL = FALSE, {
+      shinyjs::toggleState("source_code_container", condition = !is.null(doc_out()))
+    })
```
gogonzo: This observer doesn't seem necessary as it could be moved to `renderUI`. `is_button_showed` should determine if button is showed, we don't need this extra observer just to hide empty container.   
```suggestion
```

R/source_code.R
```diff
@@ -0,0 +1,85 @@
...
+      }
+    })
+
+    is_button_showed_r <- shiny::reactive({
+      getOption("teal.show_src", TRUE) &&
+        inherits(mod_out_r(), "qenv") &&
+        !isFALSE(attr(mod_out_r()@code, "teal.show_src")) # Only hide when value is explicitly FALSE
+    })
+
+    .call_once_when(!is.null(doc_out()), {
```
gogonzo: do we really need `call_once_when`?

R/source_code.R
```diff
@@ -0,0 +1,85 @@
...
+  shiny::tagList(
+    htmltools::htmlDependency(
+      name = "teal-reporter-busy-disable",
+      version = utils::packageVersion("teal.reporter"),
+      package = "teal.reporter",
+      src = "js",
+      script = "busy-disable.js"
+    ),
+    shiny::actionButton(
+      shiny::NS(id, shiny::NS("source_code", "button")),
```
gogonzo: Nice work-around to avoid `verbatim_popup_ui` but not accepting. This is not documented nor commented at all. Because of this `srv_source_code` is confusing where `ui_source_button(session$ns(NULL))` is linked with `teal.widgets::verbatim_popup_srv(id = "source_code")`.

You can consider renaming  `ui_source_button` -> `verbatim_popup_ui` and call it with the same namespace in the renderUI


```suggestion
      shiny::NS(id, "button"),
```

R/source_code.R
averissimo: Good point, I copied it blindly from `reporter` and corrected in both places with updated code

R/source_code.R
averissimo: I don't agree with this one, we only want to show the button if it's a `qenv`, otherwise the next clause will crash if it's not a qenv

`isFALSE(attr(mod_out_r()@code, "teal.show_src"))`

`NULL` case is "included" in the inherits

```r
inherits(NULL, "qenv")
#> [1] FALSE
```

R/source_code.R
averissimo: I believe the answer is no

It's both here and in `reporter` to avoid starting a server function and extra `render` unless it's really necessary

With that in mind, WDYT? remove only here, both, or keep it as is.

I would favor removing it to keep the code simpler, even if there's a small penalty

R/module_teal_reporter.R
```diff
@@ -109,15 +109,13 @@ srv_add_reporter <- function(id, module_out, reporter) {
   } # early exit
   moduleServer(id, function(input, output, session) {
     mod_out_r <- reactive({
-      req(module_out)
-      if (is.reactive(module_out)) {
-        module_out()
+      if (!is.null(module_out) && is.reactive(module_out)) {
+        tryCatch(module_out(), error = function(e) e)
```
averissimo, line 130: Alternatively, if we want to break the reactivity chain and don't show anything if output is null (use @llrs-roche suggestion in #1597):

```r
      if (req(is.reactive(module_out))) {
```

note: @llrs-roche flagged that the previous code would trigger the `after` function being called twice.

R/module_nested_tabs.R
```diff
@@ -225,9 +225,16 @@ srv_teal_module <- function(id,
...
-      .modules_breadcrumb(modules),
+      tags$div(
+        style = "display: flex; justify-content: space-between; align-items: center; width: 100%;",
+        .modules_breadcrumb(modules),
+        tags$div(
+          style = "display: flex; gap: 0.5em;",
+          ui_add_reporter(ns("add_reporter_wrapper")),
+          ui_source_code(ns("source_code_wrapper"))
+        )
+      ),
```
averissimo, line 237: Discussion for @insightsengineering/nest-core-dev (from coredev corner)

Q: Should the buttons on top right `+ Add to report` and `Show R code`, always be shown in the UI?

Consider:
- Disabling creates a consistent UI look (no mysterious buttons appearing or disappearing while using teal app)
    - It also gives an explanation why button is disabled
- Global option would completely hide either buttons
- We could add an additonal `after` function that also hides them in case modules use old API

image

R/module_source_code.R
```diff
@@ -0,0 +1,114 @@
...
+#'   )
+#' )
+#' if (interactive()) {
+#'   shinyApp(app$ui, app$server)
+#' }
+#' @export
+disable_src <- function(x) {
+  checkmate::assert_multi_class(x, c("teal_module", "teal_modules"))
+  after(x, server = function(data) {
+    attr(data, "teal.enable_src") <- FALSE
```
averissimo, line 110: Discussion for @insightsengineering/nest-core-dev (from coredev corner)

Using attributes set on the S4 object instead of the slots. The attribute names uses a namespace `prefix` to indicate the context.

In the future we may use `shared.xxxx` to pass options to other modules (think `mdr` composite modules). This makes it easier to maintain over time.

R/source_code.R
averissimo: Good call, see the new changes in 4bb8ecd42

It "overrides" the `verbatim_popup_ui`

R/source_code.R
averissimo: The problem with moving it to `renderUI` is that it re-renders it _(graying out the UI while shiny is computing)_.

Plus. This `toggleState` doesn't control the visibility of the element, it disables/enables the button _(disables when module has error)_

R/source_code.R
```diff
@@ -0,0 +1,122 @@
...
+        session$sendCustomMessage("disable-tooltip", session$ns("source_code_wrapper"))
+      } else {
+        session$sendCustomMessage("enable-tooltip", session$ns("source_code_wrapper"))
+        bslib::update_tooltip(id = "source_code_tooltip", reason)
+      }
+
+      shinyjs::toggleState(
+        "source_code_wrapper",
+        condition = !is.null(code_out()) &&
+          checkmate::test_multi_class(mod_out_r(), c("qenv", "qenv.error")) &&
```
averissimo: Allowing `qenv.error` that has an attribute that disables functionality

R/source_code.R
averissimo: This may be ignored if we keep the buttons always showing in the UI

R/module_nested_tabs.R
averissimo: @kumamiao @donyunardi your opinion is welcome here as this is a product feature.

We can implement any of the choices.

kumamiao: Thank you @averissimo for the proposal. I personally prefer the "Keep & disable" option, which encourages our users to utilize more of these features, and makes it more consistent like you said. A few other considerations points:

1. With our recent refactors, it has become much easier for customized modules to enable these features. So, adding them in newly created customized modules should not place too much extra burden on developers.
2. My only concern is with legacy modules, which may have been built with the old customization approach. Updating those may not be straightforward - or the developer may simply not have time. In such cases, app users might question why a feature is “disabled” and request it to be enabled, which could lead to complaints from module developers. I suggest changing the wording from "is disabled" to "is not enabled".
4. Regarding your second point on the global option: will it still display the two buttons but keep them greyed out, or will it completely hide them across the app? If it’s the latter, how does the implementation look like for app developers, similar to disabling them on the module level or via options?

R/module_teal_reporter.R
llrs-roche: I investigated this more adding some logging/ records. It turns out that this reactive is called twice:

```
[INFO] 2025-09-23 14:11:53.7255 pid:23792 token:[f4d8f90a] teal Logging general .call_once_when: 0
[INFO] 2025-09-23 14:11:54.0441 pid:23792 token:[f4d8f90a] teal Logging specific .call_once_when: 0
[INFO] 2025-09-23 14:11:54.1674 pid:23792 token:[f4d8f90a] teal srv_add_reporter 0
[INFO] 2025-09-23 14:11:54.3699 pid:23792 token:[f4d8f90a] teal srv_add_reporter 1
[INFO] 2025-09-23 14:11:54.3801 pid:23792 token:[f4d8f90a] teal Logging after function: 0 
```

I'm not sure why or how `srv_add_reporter` reactive `mod_out_r` is executed twice: it is under two `.call_once_when` functions.  The logging message was at:

```r
mod_out_r <- reactive({
      if (!is.null(module_out) && is.reactive(module_out)) {
            message("srv_add_reporter ", id, " ", x)
            module_out()
```
The bad news is that neither the `!is.null` or `req(is.reactive(` helped to just execute `module_out()` once. 
This might be worth of a different issue to to increase the scope of this PR.

R/module_nested_tabs.R
averissimo: > 2. My only concern is with legacy modules, which may have been built with the old customization approach. Updating those may not be straightforward - or the developer may simply not have time. In such cases, app users might question why a feature is “disabled” and request it to be enabled, which could lead to complaints from module developers. I suggest changing the wording from "is disabled" to "is not enabled".

:+1: this reads better and more diplomatic. Thanks!

Do you think this deals with most legacy modules? I'm worried with some that have `Show R Code` in the encoding panel and then disabled in the top-right

> 3. Regarding your second point on the global option: will it still display the two buttons but keep them greyed out, or will it completely hide them across the app? If it’s the latter, how does the implementation look like for app developers, similar to disabling them on the module level or via options?

We implement either way, my intuition is that it should not appear at all with global option.

Re: how does it look: it's being done differently for `reporter` and `show r code` (something that @m7pr flagged)

- `reporter` as an option in `teal::init` call when this argument is passed as null
- `show r code` as an R option set before shiny call `teal.enable_src = FALSE`

It is not part of `init` for 2 reasons: `1.` keeping the function arguments minimal (after we removed title, footer, etc..), and `2.` we want to discourage disabling this functionality.

R/module_teal_reporter.R
averissimo: The culprit might be the module itself, as it outputs 2 different values (depending on the module) it might start with an error state and move towards good one.

R/module_nested_tabs.R
kumamiao: > Do you think this deals with most legacy modules? I'm worried with some that have Show R Code in the encoding panel and then disabled in the top-right

I have a similar concern now. My naive intuition before was that their existing SRC will be directly translated to the SRC on the top-right, but it looks like with the new approach, we no longer have the argument below in `teal.widgets::standard_layout()`, so there will be scenarios like you mentioned - the SRC is still in the encoding, but the upper right showing SRC is not enabled. Any good proposals here? Could there be some utility function to make the conversion? Do we have some documentation on making the updates from legacy modules? Maybe the wording could add something like"..., or has not been updated with the new approach of creating SRC"? tagging @donyunardi here in case any other ideas.
```
forms = tagList( teal.widgets::verbatim_popup_ui(ns("rcode"), button_label = "Show R code") )
```

Re: implementation, I think the proposal makes sense for it to be not a part of `init`, just wanted to make sure that they're not implemented the same way as the module level, which could lead to confusion.

R/source_code.R
```diff
@@ -0,0 +1,122 @@
...
+    })
+
+    output$source_code_container <- renderUI({
+      if (is_button_showed_r()) {
+        bslib::tooltip(
+          id = session$ns("source_code_tooltip"),
+          trigger = shiny::tags$div(
+            id = session$ns("source_code_wrapper"),
+            class = "cursor-helper",
+            ui_source_button(session$ns(NULL))
```
gogonzo: `ui_source_button` is redundant

```suggestion
            verbatim_popup_ui(session$ns("source_code"))
```

R/source_code.R
```diff
@@ -0,0 +1,122 @@
...
+        "The show source code functionality is disabled for this module."
+      } else if (inherits(mod_out_r(), "error")) {
+        "The module returned an error, check it for errors."
+      } else if (is.null(code_out())) {
+        "The module does not support source code functionality"
+      }
+    })
+
+    output$source_code_container <- renderUI({
+      if (is_button_showed_r()) {
```
gogonzo: option is not reactive, doesn't need to make it so - it will never trigger anything.

```suggestion
      if (getOption("teal.show_src", TRUE)) {
```

R/source_code.R
```diff
@@ -0,0 +1,122 @@
...
+      src = "js",
+      script = "busy-disable.js"
+    ),
+    shiny::actionButton(
+      shiny::NS(id, "button"),
+      "Show R code",
+      class = "primary teal outline-button teal-busy-disable",
+      title = title
+    )
+  )
```
gogonzo: isn't it easier to use `.outline_button` (imported via getFromNamespace?)

R/source_code.R
```diff
@@ -0,0 +1,122 @@
...
+}
+
+#' @noRd
+verbatim_popup_srv <- teal.widgets::verbatim_popup_srv
+
+#' @noRd
+ui_source_button <- function(id, title = NULL) {
+  verbatim_popup_ui(shiny::NS(id, "source_code"), title)
```
gogonzo: `verbatim_popup_srv` will become unused after removal from teal.modules.* - maybe it is worth to not hack it and do the modal here from scratch?

R/teal_reporter.R
```diff
@@ -129,20 +127,54 @@ srv_add_reporter <- function(id, module_out, reporter) {
...
+        )
+      }
+    })
+    teal.reporter::add_card_button_srv("reporter_add", reporter = reporter, card_fun = doc_out)
 
     observeEvent(doc_out(), ignoreNULL = FALSE, {
-      shinyjs::toggleState("reporter_add_container", condition = inherits(doc_out(), "teal_card"))
+      reason <- trimws(reason_r() %||% "")
+      if (is.null(reason) || identical(reason, "")) {
+        session$sendCustomMessage("disable-tooltip", session$ns("report_add_wrapper"))
```
gogonzo: `add_card_button_ui` is called in reactive context anyway - isn't it easier to render this button with/without tooltip depending on a reason in renderUI?
Looks like the whole process can be easily encapsulated within shiny module

R/source_code.R
averissimo: I initially was going after a pattern, but that was obsolete. Changed and removed

R/source_code.R
averissimo: I think we should avoid repeated code for the same purpose on the framework. Once that the modules no longer use it, we could eventually move that to `teal`.

I've improved on the implementation.

Alternatively, I could make the necessary change on `teal.widgets` and avoid this UI override.

R/teal_reporter.R
averissimo: Adding this logic in the rendering will gray out the button when the module itself is computing.

It's creates a distraction from my point of view.

Whether that warrants this extra complexity is another thing.

Alternatively, we could add a tooltip that describes the action more in detail

- `+ Add to report`: Click here to add the current module to the Report"
- `Show R code`: "Click here to see the R code used in the module"

R/source_code.R
averissimo: Fixed

R/teal_reporter.R
averissimo: If we go with the `renderUI` approach, then the `enabled`/`disabled` toggle, must also be controlled there as `shinyjs` will not be consistent

R/module_nested_tabs.R
```diff
@@ -225,9 +225,16 @@ srv_teal_module <- function(id,
...
-    ui_add_reporter(ns("add_reporter_wrapper")),
     tagList(
-      .modules_breadcrumb(modules),
+      tags$div(
+        style = "display: flex; justify-content: space-between; align-items: center; width: 100%;",
+        .modules_breadcrumb(modules),
+        tags$div(
+          style = "display: flex; gap: 0.5em;",
+          ui_add_reporter(ns("add_reporter_wrapper")),
+          ui_source_code(ns("source_code_wrapper"))
```
gogonzo, line 235: @averissimo since button is always there, and you don't want to link it with the "reason", maybe it should look like this?

```suggestion
          if (getOption("teal.show_src", TRUE)) {
            bslib::tooltip(
              id = ns("source_code_tooltip"),
              trigger = shiny::tags$div(
                id = session$ns("source_code_wrapper"),
                class = "cursor-helper",
                ui_source_button(session$ns(NULL))
              ),
              uiOutput(ns("reason"))
          )
```

R/teal_reporter.R
gogonzo: Got it, what about this https://github.com/insightsengineering/teal/pull/1592/files#r2382268043

R/module_nested_tabs.R
averissimo: Ok, let's make it shiny based with:

The visual glitch is very hard to see and this avoids the observer and the extra JS code to handle `bslib` limitation

```r
  output$source_code_container <- shiny::renderUI({
      if (getOption("teal.show_src", TRUE)) {
        reason <- trimws(reason_r() %||% "")
        new_ui <- verbatim_popup_ui(session$ns("source_code"))
        if (!identical(reason, "")) {
          new_ui <- bslib::tooltip(
            id = session$ns("source_code_tooltip"),
            trigger = shinyjs::disabled(
              shiny::tags$div(
                id = session$ns("source_code_wrapper"),
                class = "cursor-helper",
                new_ui
              )
            ),
            reason
          )
        }
        new_ui
      }
    })
```

R/source_code.R
gogonzo: > Alternatively, I could make the necessary change on teal.widgets and avoid this UI override.


Makes more sense, there is a `disabled` argument (`verbatim_popup_srv(disabled)`), what if the "reason" was implemented in `verbatim_popup_srv`? 
Maybe we can sneak-in reason somehow?

R/module_nested_tabs.R
averissimo: Here's the visual glitch that will show: as it takes some time for the `Show R code` button to correctly render.

It's fast, but looking at the video below you can see it (screenshots of the incorrect state after main module is in error state)

Context:
- "Add to report" button has current implementation
- "Show R code" has the code in the previous comment

### While shiny is computing module

image

### After error on module

image

On this clip:

https://github.com/user-attachments/assets/43956532-fae4-436e-b0dd-6f75fb950d98

averissimo:
@gogonzo I've updated the PRs and added https://github.com/insightsengineering/teal.widgets/pull/319

- Static UI function _(only renders the reason for failure or description of button)_
- Render button moved to `teal.widgets` and is shared between `reporter` and `teal`

<details>

<summary>Visual glitch was solved in eb67ae9</summary>

Only problem with this approach: ~_(small visual glitch with delayed reason rendering)_~

- ~I think this is because shiny only renders `reason` when it's visible (that is: hovered)~

https://github.com/user-attachments/assets/83a16975-f52c-4e51-aecf-5fb44cfc4fe3

<details>

llrs-roche:
It might be premature, but I was testing the feature and the app works well except the manual module one that the show R code isn't disabled (it takes a while but the button gets activated):

image

It is difficult to read but the blurred letters of the bread crumb say  "Home / 📓 Module (manual)", app and branches used

<details>
<summary>Code used</summary>

```r
# Show R Code PR example app

# - teal_reporter: 1534-show_r_code@main
# - teal: 1534-show_r_code@main
# - teal.widgets: 1534-show_r_code@main
pkgload::load_all("../teal.widgets")
pkgload::load_all("../teal")
# Main branch
library(teal.logger)
library(teal.code)
library(teal.data)
library(teal.reporter)
library(teal.slice)
library(teal.widgets)
library(teal.transform)

do.call(
  options,
  rlang::list2(
    "teal.bs_theme" = bslib::bs_theme(version = "5"),
    teal.log_level = "TRACE",
    teal.show_js_log = TRUE,
    teal.show_r_code = TRUE,
  )
)

example_extended <- function(label = "example teal module",
                             datanames = "all",
                             transformators = list(),
                             decorators = list()) {
  checkmate::assert_list(decorators, "teal_transform_module")

  mod <- example_module(label, datanames, transformators, decorators)

  module(
    label,
    server = function(id, data, decorators) {
        result <- mod$server("example", data, decorators)

        reactive({
          q <- result()
          attr(q, "teal.enable_src") <- FALSE
          q
        })
      })
    },
    ui = function(id, decorators) mod$ui(shiny::NS(id, "example"), decorators),
    ui_args = mod$ui_args,
    server_args = mod$server_args,
    datanames = mod$datanames,
    transformators = mod$transformators
  )
}

teal::init(
  data = within(teal_data(), {
    iris <- iris
    require(nestcolor)
    CO2 <- CO2
  }),
  modules = modules(
    example_module(label = "Module (full)"),
    example_extended(label = "📓 Module (manual)"),
    example_module(label = "📓 Module (pipe disable show r code)") |> 
      disable_src(),
    example_module(label = "🗒️ Module (pipe disable report)") |> 
      disable_report(),
    example_module(label = "🗒️📓 Module (pipe disable both)") |> 
      disable_src() |> 
      disable_report()
  )
) |> shiny::runApp()
```

</details>

averissimo:
Thanks @llrs-roche that was a problem with the actual example, as the attribute that hints the disable has changed from the `code` slot to the object itself.

However, I'm finding a problem when not using the `result` of the inner module, that is with: 

It should give the original data and disable the button, but the button is enabled

```r
        reactive({
          q <- data() # instead of the "correct" `q <- result()`
          attr(q, "teal.enable_src") <- FALSE
          q
        })
```

This is limited to this use case, which is very weird.

vignettes/teal-options.Rmd
```diff
@@ -156,6 +156,12 @@ list(
 )
 ```
 
+### `teal.show_src` (`logical`)
```
averissimo, line 159: This name is deliberately different from the property to disable inside a module as it has a different goal (hides/removes it)

inst/css/custom.css
averissimo: No longer relevant

R/module_source_code.R
```diff
@@ -0,0 +1,112 @@
+#' @noRd
+verbatim_popup_srv <- teal.widgets::verbatim_popup_srv
+
+#' @noRd
+verbatim_popup_ui <- function(id) {
```
gogonzo, line 5:
```suggestion
verbatim_popup_ui <- function(id) {
  # we replace teal.widgets::verbarbatim_popup_ui with own button. `inputId="button"` is necessary
```

inst/css/custom.css
```diff
@@ -134,8 +134,7 @@ body > div:has(~ #shiny-modal-wrapper .blur_background) {
...
 
-.teal.add-reporter-container {
-  display: flex;
-  justify-content: flex-end;
-  padding: 10px;
-}
+[data-bs-toggle="tooltip"].disabled {
+  cursor: not-allowed;
+  color: var(--bs-gray);
+}
```
gogonzo:
```suggestion
}

```

gogonzo (approved):
👍

R/source_code.R
averissimo: We need to implement the tooltip in a parent element of `verbatim_popup_button`.

I would prefer to have an identical setup and eventually deprecate this function (moving it internally to `teal`). It might be easier to maintain if `add to report` and `show r code` are as similar as possible.

PR #1595, opened 2025-09-16
Cheatsheet

chizapoth:
Content changes


llrs-roche (approved):
Nice addition. It passes my local checks. 

I thought we would add a link to the file on the README or somewhere were users can find it.

m7pr:
What do you think about making the dataname consistent across the whole cheatsheet?
So it's always either `IRIS <- iris`, so that we can refer to `IRIS` everywhere

image

chizapoth:
> Nice addition. It passes my local checks.
> 
> Two comments: I thought we would add a link to the file on the README or somewhere were users can find it.

@donyunardi @kumamiao We should definitely add the link (and the thumbnail picture), the question is where: my suggestions are: 
- README.md, right under (or before) installation
- [teal vignette](https://insightsengineering.github.io/teal/latest-tag/), right under (or before) installation

I would also suggest to remove the picture of the video, instead **keep the link to the video**. 

> One minor note about the PDF, on the first page there is some space between the background image and the limits of the document. See the space marked with the red lines:
> 
> image
> I'm not sure if this is an artifact for printing or something else.

This will be fixed!

chizapoth:
> What do you think about making the dataname consistent across the whole cheatsheet? So it's always either `IRIS <- iris`, so that we can refer to `IRIS` everywhere
> 
> image

Good point! Will update this detail.

m7pr:
image
The main code example uses `example_module()`. While simple, it would be more instructive to use one of the `real` modules listed on the second page, such as e.g. `tm_variable_browser`. This would better connect the app-building example with the module reference section.

m7pr:
It might be just me, but I think it would look better if the sentence `Build your own modules` could start from a new line
image

So it's
```
Can't find what you need?
Build your own modules!
```

instead

```
Can't find what you need? Build your own 
modules!
```

m7pr:
Hey, this is excellent work. Just few things to consider, but don't feel obliged to incorporate the feedback.
I get the impression that print screens of the teal app are blurry and hard to read. Is there anyway to improve the quality of the screenshots?

While taking new screenshots, maybe you can substitute the app from the first page with a real use case app. Maybe something from teal gallery? So that we can show a powerful setup with some cool plots and tables, instead of an empty demo. So that we can spark and trigger the imagination when showcasing the full capacity of teal apps!?

chizapoth:
I have done the following modifications: 

- aligned the thumbnail logo part so it looks a bit better
- modified example data name IRIS so it's consistent
- made a new line for 'Build your own module'

I have not replaced the screenshot for the example module. I think given that we will have a user guide that are specifically designed to teach people what functionalities there are, maybe this cheatsheet should remain the 'app developer' guide. Also I find that **example modules** are simple enough, while the tmc modules can be very rich in information that can be overwhelming at the first glance. 

I do think we can add one line like 'Please check the user guide for features of teal module' when we have the user guide!

I will add the link to the thumbnail and download after this PR is merged.

donyunardi:
> @donyunardi @kumamiao We should definitely add the link (and the thumbnail picture), the question is where: my suggestions are:
> README.md, right under (or before) installation

I would say just in README.md after the installation and before the usage section. I checked couple posit packages and this seems to be to common place to put the cheatsheet. 
Can you please add it with this PR? 

> I would also suggest to remove the picture of the video, instead keep the link to the video.

I would prefer to keep the video jpg. 

> I get the impression that print screens of the teal app are blurry and hard to read. Is there anyway to improve the quality of the screenshots?

Please use this repo to get the high quality images for NEST products:
https://github.com/insightsengineering/hex-stickers

Here is teal:
https://github.com/insightsengineering/hex-stickers/blob/main/PNG/teal.png

PR #1593, opened 2025-09-12
Update report in custom modules vignette for new teal.reporter

llrs-roche:
# Pull Request

Fixes #1591

Update the vignette to explain how teal.reporter can be used to generate the report on modules.

There might be room for improvement on the vignette as this is my first module using this new features.
Specially this section of code doesn't seem self explanatory:

```r
 teal.reporter::teal_card(obj) <-
            c(
              teal.reporter::teal_card("# Module with reporting"),
              teal.reporter::teal_card(obj),
              teal.reporter::teal_card("## Module's code")
            )
```

It is an object of class `teal_report` (of a single module!) being modified by the function `teal_card` to add `teal_card`s. But I might have got this wrong. 

Another thing I'm not confident is on having to use `module()[[input$dataname]]` for the `renderPrint`. After having to add a line to print the table selected for the report.

I removed lot's of apps that showed intermediate work, as I think now it can be done in two steps:
 1. Adding a card
 2. Returning it to be used by teal

Awesome work that went into this! 
In general, it felt very easy to set it up once I got around the `teal_reporter` and `teal_card` classes seeing other modules on TMG.  But let me know if I missed a clearer description.

Issue #1591, opened 2025-09-08
[Documentation]: Update adding-support-for-reporting.Rmd vignette in teal to use the new approach


donyunardi:
### Feature description

Update adding-support-for-reporting.Rmd vignette in teal to use the new approach

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [x] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [x] I agree to follow this project's Security Policy.

PR #1596, opened 2025-09-17
Add app example for disable_src()

llrs-roche:
# Pull Request

Part of #1591 
Companion of #1593

### DIFF
--- a/vignettes/adding-support-for-reporting.Rmd
+++ b/vignettes/adding-support-for-reporting.Rmd
@@ -260,6 +260,35 @@ As the module writer, you can also add any other content to the report you'd lik
 `teal.reporter` supports the addition of tables, charts, and more.
 For more information, explore the API of `teal_report()` to learn about the supported content types.
 
+
+## Removing support for Show R code
+
+If your module supports a report but you want to disable the show R code functionality, use `disable_src()`
+
+```{r app_4}
+app <- init(
+  data = teal_data(IRIS = iris, MTCARS = mtcars),
+  modules = my_module_with_reporting() |> disable_src()
+)
+
+if (interactive()) {
+  shinyApp(app$ui, app$server)
+}
+```
+
+```{r shinylive_iframe_4, echo = FALSE, out.width = '150%', out.extra = 'style = "position: relative; z-index:1"', eval = requireNamespace("roxy.shinylive", quietly = TRUE) && knitr::is_html_output() && identical(Sys.getenv("IN_PKGDOWN"), "true")}
+code <- paste0(c(
+  knitr::knit_code$get("as_interactive"),
+  knitr::knit_code$get("setup"),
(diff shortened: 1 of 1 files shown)


vignettes/adding-support-for-reporting.Rmd
```diff
@@ -260,6 +260,35 @@ As the module writer, you can also add any other content to the report you'd lik
 `teal.reporter` supports the addition of tables, charts, and more.
 For more information, explore the API of `teal_report()` to learn about the supported content types.
 
+
+## Removing support for Show R code
```
averissimo:
```suggestion
## Removing support for displaying reproducible code
```

or 

```suggestion
## Removing support for Show R code button
```

vignettes/adding-support-for-reporting.Rmd
```diff
@@ -260,6 +260,35 @@ As the module writer, you can also add any other content to the report you'd lik
...
+
+If your module supports a report but you want to disable the show R code functionality, use `disable_src()`
```
averissimo: Is "Show R code" functionality self-explanatory?

Looking at this with app/mod developer goggles it might not be completely clear what we mean with this

averissimo:
```suggestion
If your module supports a report but you want to disable the button that allows to display the module's reproducible code ("Show R code"), use `disable_src()`
```

llrs-roche: Yes, this is more explanatory. Added!

vignettes/adding-support-for-reporting.Rmd
llrs-roche: I went with the first, as the button is a secondary consequence.

vignettes/adding-support-for-reporting.Rmd
```diff
@@ -260,6 +260,35 @@ As the module writer, you can also add any other content to the report you'd lik
 `teal.reporter` supports the addition of tables, charts, and more.
 For more information, explore the API of `teal_report()` to learn about the supported content types.
 
+
+## Removing support for displaying reproducible code
+
+If your module supports a report but you want to disable the button that allows to display the module's reproducible code ("Show R code"), use `disable_src()`:
+
```
averissimo, line 267: Can you add a note that this can be applied to both an individual module and multiple at once _(when nested)_

(which disables to all modules recursively)

```r
app <- init(
  modules = modules(
    example_module(label = "1"),
    example_module(label = "2"),
    example_module(label = "3")
  ) |> disable_src()
)
```

llrs-roche: Sure! I saw your testing app and I saw it worked but I thought it would be better if we kept things simple. But sure it will help for those with complex apps.

PR #1597, opened 2025-09-17
Improve after function

llrs-roche:
# Pull Request

Fixes comments from [#1593](https://github.com/insightsengineering/teal/pull/1593#issuecomment-3296483031) and https://github.com/insightsengineering/teal/pull/1592#discussion_r2351702950

- [x] add examples to `after` function
- [ ] have a conversation on whether we should rename `after` function to something else, like `edit_server`, `edit_server_data`
  - [ ]  Check the vignettes too. 
- [ ] Add tests to after function

### DIFF
--- a/R/after.R
+++ b/R/after.R
@@ -8,6 +8,18 @@
 #' @param ... additional argument passed to `ui` and `server` by matching their formals names.
 #' @return A `teal_report` object with the result of the server function.
 #' @export
+#' @examples
+#' app <- init(
+#'   data = teal_data(IRIS = iris, MTCARS = mtcars),
+#'   modules = after(example_module(), server = function(input, output, session, data) {
+#'     teal_card(data) <-
+#'       c(teal_card(data), teal_card("## New title", "text"))
+#'     data
+#'   })
+#' )
+#' if (interactive()) {
+#'   shinyApp(app$ui, app$server)
+#' }
 after <- function(x,
                   ui = function(id, elem) elem,
                   server = function(input, output, session, data) data,
--- a/man/after.Rd
+++ b/man/after.Rd
@@ -27,3 +27,16 @@ A \code{teal_report} object with the result of the server function.
 Primarily used to modify the output object of module to change the containing
 report.
 }
+\examples{
+app <- init(
+  data = teal_data(IRIS = iris, MTCARS = mtcars),
+  modules = after(example_module(), server = function(input, output, session, data) {
+    teal_card(data) <-
+      c(teal_card(data), teal_card("## New title", "text"))
+    data
+  })
+)
+if (interactive()) {
+  shinyApp(app$ui, app$server)
+}
+}


llrs-roche:
We can duplicate the module UI! As a consequence the show R code works in both but only the first input of one updates the output of the other:

```r
library("teal.reporter")
app <- init(
  modules = example_module() |> 
    after(
      ui = function(id, elem) {c(elem,elem)},
      server = function(data) {
        teal_card(data) <- c(teal_card(data), teal.reporter::teal_card("Modification"))
        within(data, object[, 1:2])
      })
)
if (interactive()) {
  shinyApp(app$ui, app$server)
}
```

image

I think if we only want to modify the server part we should not allow to update the UI (at least without some more checks).

gogonzo:
> ```r
> library("teal.reporter")
> app <- init(
>   data = teal_data(IRIS = iris, MTCARS = mtcars),
>   modules = example_module() |> 
>     after(
>       ui = function(id, elem) {c(elem,elem)},
>       server = function(data) {
>         teal_card(data) <- c(teal_card(data), teal.reporter::teal_card("Modification"))
>         within(data, object[, 1:2])
>       })
> )
> if (interactive()) {
>   shinyApp(app$ui, app$server)
> }
> ```
> 

Example doesn't show capabilities of using `ui`. Here is an example which we can include in a conversation

```r
library("teal.reporter")
app <- init(
  modules = example_module() |> 
    after(
      ui = function(id, elem) {
        ns <- NS(id)
        check_box <- checkboxInput(ns("src"), "Include R Code in the report", TRUE)
        htmltools::tagAppendChild(elem, check_box, .cssSelector = ".standard-layout .sidebar .sidebar-content")
      },
      server = function(input, output, session, data) {
        teal_card(data) <- c(teal_card(data), teal.reporter::teal_card("Modification"))
        if (!input$`wrapper-src`) {
          teal_card(data) <- Filter(function(x) !inherits(x, "code_chunk"), teal_card(data))
        }
        data
      }
    )
)

runApp(app)
```

llrs-roche:
I'll replace the example with yours it is more complete and show a) how the ui is namespaced and can be used on the server b) it is actually useful.

gogonzo:
> a) how the ui is namespaced and can be used on the server

I think we can fix the namespace so that ui<->server. I think we don't need this "wrapper" namespace. Fun fact, it is also possible to refer to teal_module's inputs, by calling `input$wrapped-<module input>` (you can test with `input$wrapped-dataname` of `example_module`)

llrs-roche:
I focused on testing and small modifications. I didn't manage to remove the wrapped namespace and keep apps working: probably it will require deeper modifications. 

While exploring how it works I found that it won't work with multiple modules: `after(modules(example_module(), example_module()))`.
Do we want it to work with nested modules or multiple ones?
This behavior would match the `disable_src()` and `disable_report()` but I think it introduces more complexity.
What would happen if we have: `after(modules(example_module() |> after(), example_module()))` ?

Suggested names:
- `teal_modify_module`: This is similar to `teal_transform_module`, it also modifies the UI but only modifies the output
- `modify_module`: We have `modify_header`, `modify_title`, on `?teal_modifiers`. This function can modify everything so it makes sense to be more general (and be on its own help page).
- `after`: makes reference to when is this function executed, after the original module is run (via `srv_add_reporter`)

R/teal_reporter.R
```diff
@@ -109,8 +109,7 @@ srv_add_reporter <- function(id, module_out, reporter) {
   } # early exit
   moduleServer(id, function(input, output, session) {
     mod_out_r <- reactive({
-      req(module_out)
-      if (is.reactive(module_out)) {
+      if (req(is.reactive(module_out))) {
```
gogonzo: new code is not equivalent of the previous one. I suspect also that it should rather be (does it? @averissimo )
```
if (is.reactive(module_out)) {
  req(module_out)
  ...
```

R/after.R
```diff
@@ -1,24 +1,50 @@
...
+#' runApp(app)
 after <- function(x,
                   ui = function(id, elem) elem,
                   server = function(input, output, session, data) data,
                   ...) {
   # todo: make a method for teal_app and remove teal_extend_server?
   checkmate::assert_multi_class(x, "teal_module")
-  if (!is.function(ui) || !all(names(formals(ui)) %in% c("id", "elem"))) {
-    stop("ui should be a function of id and elem")
+  if (!is.function(ui) || !all(c("id", "elem") %in% names(formals(ui)))) {
```
gogonzo: New code is not equivalent of the previous code and allows `ui` to have more arguments that just `id` and `elem`.

tests/testthat/test-after.R
```diff
@@ -0,0 +1,20 @@
...
+
+test_that("after requires functions for server", {
+  expect_error(after(example_module(), server = "a"), "server should be a function")
+  expect_error(after(example_module(), server = function(data){data}), "server should be a function")
+  expect_no_error(after(example_module(), server = function(input, output, data, session){data}))
+})
+
+test_that("after returns a module", {
+  expect_s3_class(after(example_module()), "teal_module")
+})
```
gogonzo: Would be nice to add a test in `test-module_teal.R` to see if after can change the data returned from the module

R/after.R
```diff
@@ -28,64 +54,64 @@ after <- function(x,
...
-        wrapper_args$data <- original_out()
-        do.call(`_y`, wrapper_args[names(formals(`_y`))], quote = TRUE)
+        output <- if (is.reactive(original_out)) {
+          original_out()
+        } else {
+          original_out
+        }
+        wrapper_args$data <- output
+        do.call(`_new_server`, wrapper_args[names(formals(`_new_server`))], quote = TRUE)
       })
```
gogonzo: Please consult with @averissimo who is checking out these areas recently. I think that final lines of the `.after` should be (conclusion is just from code analysis)
```suggestion
      wrapper_args$data <- reactive({
        if (is.reactive(original_out)) {
          original_out()
        } else {
          original_out
        }
      })

      do.call(`_new_server`, wrapper_args[names(formals(`_new_server`))], quote = TRUE)
```

gogonzo (commented):
Tests don't cover module's reactive output.

gogonzo:
> While exploring how it works I found that it won't work with multiple modules: after(modules(example_module(), example_module())).

This change is pending in other [PR](https://github.com/insightsengineering/teal/pull/1592) 

> What would happen if we have: after(modules(example_module() |> after(), example_module())) ?

It would wrap module around module. BTW question should be asked in mentioned [PR](https://github.com/insightsengineering/teal/pull/1592)

R/after.R
llrs-roche: This is to force the functions to have the required arguments so they don't miss some arguments.
Previously using something like `ui =function(id){}` would silently pass: https://github.com/insightsengineering/teal/pull/1597/files#diff-b5a2f9463d5d57838cb6ee86ae054d43335895a21e92f598cb310abe59137638R8

You can see there are some attempts to use other arguments: https://github.com/insightsengineering/teal/pull/1597/files#diff-a5c0a1711563afbdaac3ed2683887be7ae554566b927661df204b5e234b53254R64
That's why I opened the checks on arguments used.

R/teal_reporter.R
llrs-roche: My understanding of the code was: it checked if `module_out` was truthy, if it was it checked that it was reactive. 
Now it checks if it is available and reactive in a single step.

tests/testthat/test-after.R
llrs-roche: Yes, I was considering doing this I'll add it there then.

R/after.R
llrs-roche: Will do! Thanks for the feedback

R/teal_reporter.R
averissimo: I believe that the `req(module_out)` was a way to throw a `shiny.silent.error` and break the reactivity after.

The new code seems to be doing the same thing, but this is also being addressed in #1592

averissimo: (I've added a review comment there to remind. (pending until I submit the other comments)

image

R/after.R
m7pr: I also talked about similar testing for srv with @averissimo and I think those test should remain as they were.

It's better to allow `ui = f(id)` to pass, and also `ui = f(id, elem)` to pass, but it prevented if you had too much arguments.
This would fail `ui = f(id, elem, x, y)`. Now `ui = f(id, elem, x, y)` would not fail, and it's not the intention. It's the intention to limit the arguments to max `id` and `elem`.

R/teal_reporter.R
llrs-roche: Removed this change as per discussion with André.

R/after.R
llrs-roche: I didn't find any comment on what was the intention of the check and it was not obvious. I think in a call I had with @averissimo he said this was fine. But I'll keep the old behavior as it is more restrictive.

llrs-roche:
For the reviewer, see the comment about naming of the function here: https://github.com/insightsengineering/teal/pull/1597#issuecomment-3311535019

Once a name is decided I'll go to update it

R/after.R
averissimo: @m7pr in what cases do we have/want `function(id) {...}` as the `ui` here?

As this would make the module's UI completely disappear.

averissimo:
@llrs-roche do you want to bring the S3 methods for after into this PR? or should we do it on a separate one

Setting several in pipes works on the `show r code` PR #1592 (with some modification to the `ui` to add some headers)

- UI is composed one on top of the other
- Data is modified with both `reporter` and `show r code` disabled

```r
# Show R Code PR example app
# - teal_reporter: 1534-show_r_code@main
# - teal: 1534-show_r_code@main

pkgload::load_all("../teal.logger")
pkgload::load_all("../teal.widgets")
pkgload::load_all("../teal.transform")
pkgload::load_all("../teal")

teal::init(
  data = within(teal_data(), {
    iris <- iris
    require(nestcolor)
    CO2 <- CO2
  }),
  modules =
    #
    # Modules
    #
    modules(
      label = "📓 All mods disabled with pipe src",
      example_module(label = "📓 Module 1") |> disable_src(),
      example_module(label = "📓 Module 2")
  ) |> disable_report()
) |> shiny::runApp()
```


https://github.com/user-attachments/assets/ee67f359-5d6e-433b-ba78-55522c8bafc6

R/after.R
```diff
@@ -1,24 +1,59 @@
...
+#'           teal_card(data) <- Filter(function(x) !inherits(x, "code_chunk"), teal_card(data))
+#'         }
+#'         data
+#'       }
+#'     )
+#' )
+#' if (interactive()) {
+#'   runApp(app)
+#' }
 after <- function(x,
```
llrs-roche: About renaming the function, I suggests the following comments:
- `teal_modify_module`: This is similar to `teal_transform_module`, it also modifies the UI but `teal_transform_module` only modifies the output
- `modify_module`: We have `modify_header`, `modify_title`, on `?teal_modifiers`. This function can modify everything so it makes sense to be more general (and be on its own help page).
- `after`: makes reference to when is this function executed, after the original module is run (via `srv_add_reporter`)

averissimo: I'm a fan of `after` due to the simplicity of the name. This is also a mostly internal function and shouldn't be directly used by App developers. Instead, wrapping the code in a `<action>_<what>` such as `disable_report`.

If not I vote for `modify_module`, but if we go this way we should allow it to be a top-level modifier (that is, changes the `teal_app` object -- on a new issue of course)

llrs-roche:
@averissimo I think it would make more sense to have this function and methods in a single branch. I only realized after creating this branch that part of the functionality was on that PR. But if the methods are needed on 1592 we can keep them there and wait until that PR is merged to review this branch (and the vignettes).

R/after.R
llrs-roche: > This is also a mostly internal function and shouldn't be directly used by App developers.

The moment we export it and mention on the vignettes it stops being internal function. Are you suggesting to not mention it on the vignette but keep it exported? 
Note that if we want to modify the `teal_app` I would suggest `modify_app`. But maybe if it could work with apps but only modify the modules we could keep the name.

m7pr: @averissimo I don't think it's internal. And will probably be used by a lot of people, once mentioned in the vignettes.
I don't like `after` as it doesn't contain neither a `verb` or a `noun` so you can't really tell what it's doing. I'm also not a fan of short generic names, that might be similar to base functions, becuase sooner or later we will get a collision with the same name in the other package. Becuase of the second argument, we have `teal_card` instead of `card` and we have `teal_data` instead of `data`. I think prepending with `teal_` makes it directly connected with the package, and also allows to overcome any collisions. 

`teal_modify_module` would be my top pick, however it's similar to `teal_transform_module` so maybe we could go with `teal_update_module` so there is a distinction that one changes the input data, and the other changes the output data. 

- `teal_update_module`
- `teal_modify_module`
- `teal_modify_output`

gogonzo: What about `transform.teal_module(..., when = "after")`?

llrs-roche: I like the suggested names of @m7pr `update_module` is maybe clearer on the intention while `modify_output` focus more on when/where the modifications apply. 

@gogonzo mixing the dot and `_` pattern doesn't seem within teal's naming style.  But I see that the new `"when"` argument would open it up to transform modules `"before"`. I'm not sure this is desirable or how it would work but it could be added as an argument to `teal_update_module`.

gogonzo: > mixing the dot and _ pattern doesn't seem within teal's naming style

@llrs-roche `transform` is an S3 is base, `.teal_module` is just a method variant. `teal_update_module(when = c("before", "after"))` (or `transform`) is a good point in discussion about `teal_transform_module`

R/after.R
m7pr: @averissimo I don't know. I never used the `after` with `ui`. I am just explaining the logic after those checks, that are the same as for the `server` function

R/after.R
m7pr: Oh yeah, I like `transform.teal_module(..., when = "after")` if we can have `before` and `after` as `when` and a nice explanation :)

averissimo:
@llrs-roche :+1:  I moved the code here with 4928f4d

R/after.R
averissimo: I like the idea of `transform.teal_module`... the man page is not the best, but I think we can work with that

image

averissimo: @m7pr @llrs-roche my use of `internal` was misguided :facepalm:, I meant as: "probably never going to be used directly, but instead via wrappers".

Exporting whatever we end up naming is mandatory!

llrs-roche: Thanks all @averissimo, @gogonzo @m7pr. Let's use the `transform` method (Sorry I somehow missed this yesterday). However, I think for now I'll leave the `when = "before"` without implementing it.

R/after.R
gogonzo: I was wrong in the first time. `new_server` is not a module but a `function(input, output, session` which can (and have to) be executed in a reactive context. I'm resolving this conversation 👍

R/after.R
gogonzo: Quick note from the meeting:
We could have a `transform` methods to modify `teal_module` its `ui` and `server`
```r
transform.teal_module <- function(`_data`, ui, server, label, when = c("before", "after"), ...) {
  if (when == "before") {
    attr(`_data`, "transformators") <- c(
      attr(`_data`, "transformators"), 
      teal_transform_module(label, ui, server)
    )
  } else {
    `_data`$ui <- transform(ui)
    `_data`$server = transform(server)
  }
}

transform.teal_ui <- function(`_data`, ui, ...) { # because ui is a function we need probably to set the class 
  .after_ui(`_data`, ui)
}

transform.teal_server <- function(`_data`, server, ...) {
  ??
}
```

llrs-roche: I realized during the call that we could probably reuse the transformators and decorators existing infrastructure. Thanks for showing how it could be done.
Regarding having specific methods for server and ui, I'm not so sure. Why have a method when we only have one type of input and no class like on the `teal_ui` case? On the server case there it might help dispatching based on what is the server function (an expression, a reactive, ...)

gogonzo: > Why have a method when we only have one type of input and no class like on the teal_ui case?

Because both `ui` and `server` are functions, so it means they will fall into the same method. Method itself will have to distinguish wether it is a `server` function or `ui`.

Please note that it could be possible to `transform` only `mod$ui <- transform(mod$ui)` or `mod$server <- transform(mod$server)` (by an app developer).

R/transform.R
```diff
@@ -0,0 +1,151 @@
...
+  if (!is.function(server) || !all(names(formals(server)) %in% c("input", "output", "session", "data"))) {
+    stop("server should be a function of `input`, `output`, `session` and `data`")
+  }
+
+  additional_args <- list(...)
+  # overwriting `_data`$ui/server causes infinite recursion
+  tm <- `_data` # nolint: object_name
+
+  if (identical(when, "before")) {
+    # NOTE: Attribute is not used for transformators, or it would be needed to have more support
```
gogonzo: What is the plan for `before`?
What "Attribute is not used for transformators" means?

llrs-roche: I was about to comment on this PR to request a review. You got it faster 😄.

The [suggested code](https://github.com/insightsengineering/teal/pull/1597#discussion_r2381451197) suggests using an attribute on a module. But the current transformators feature doesn't use attributes, instead it uses the arguments to transform the inputs. Using attributes would require adding support for this feature on the `print` and other functions/methods. This comment is to address that approach to implement the `before` option. 

I am not sure how to implement the `before` strategy ( even how `after` and decorators are working). I don't think this can be done by Friday. I'll comment on https://github.com/insightsengineering/teal/issues/1603 with my findings and then if you think this can be done before the release I'll be happy to work on this.

gogonzo: > The https://github.com/insightsengineering/teal/pull/1597#discussion_r2381451197 suggests using an attribute on a module. But the current transformators feature doesn't use attributes

Yup, suggested code was wrong, instead of `attr` transformators are `<teal_module>$transformators` (list element).

> instead it uses the arguments to transform the inputs

What do you mean?

> I don't think this can be done by Friday. I'll comment on https://github.com/insightsengineering/teal/issues/1603 with my findings and then if you think this can be done before the release I'll be happy to work on this.

It doesn't have to be rushed before Friday. We don't need to release `after/transform` if it is not finished. Shape of `transform` could be affected by `when = "before"` so I'm not sure whether we want to deal with breaking changes.

llrs-roche: @gogonzo I have written some thoughts on https://github.com/insightsengineering/teal/issues/1603. Happy to have a call and see how do you think we should proceed

llrs-roche: Sorry, I didn't see [the message](https://github.com/insightsengineering/teal/pull/1597#discussion_r2390708239) above before my reply. I also tried to modify the list, I think the current assertions prevented me to modify it directly.

R/transform.R
```diff
@@ -0,0 +1,179 @@
...
+  tm <- `_data` # nolint: object_name_lintr
+
+  if (identical(when, "before")) {
+    ttm <- teal_transform_module(ui = ui, server = server)
+    # add more assertions
+    if ("elem" %in% names_ui) {
+      warning("This allows to modify the UI. But requires to return the modified UI")
+    }
+    tm$transformators <- c(tm$transformators, list(ttm))
+  } else if (identical(when, "decorator")) {
```
gogonzo, line 91: You need to make sure that `_data$ui/srv` have `decorators` in formals - at least to throw the warning.

R/transform.R
```diff
@@ -0,0 +1,179 @@
...
+    ttm <- teal_transform_module(ui = ui, server = server)
+    tm$server_args$decorators <- c(tm$server_args$decorators, list(ttm))
+    tm$ui_args$decorators <- c(tm$ui_args$decorators, list(ttm))
+    browser()
```
gogonzo, line 95:
```suggestion
```

R/transform.R
```diff
@@ -0,0 +1,179 @@
...
+
+  additional_args <- list(...)
+  # overwriting `_data`$ui/server causes infinite recursion
+  tm <- `_data` # nolint: object_name_lintr
+
+  if (identical(when, "before")) {
+    ttm <- teal_transform_module(ui = ui, server = server)
+    # add more assertions
+    if ("elem" %in% names_ui) {
+      warning("This allows to modify the UI. But requires to return the modified UI")
```
gogonzo, line 88: Not really clear message about what went wrong.

tests/testthat/test-shinytest2-transform.R
```diff
@@ -0,0 +1,141 @@
+testthat::test_that("e2e: transform before decorators", {
```
gogonzo, line 1:
```suggestion
testthat::test_that("e2e: transform(when = 'before') doesn't fail the app", {
```

tests/testthat/test-shinytest2-transform.R
```diff
@@ -0,0 +1,141 @@
...
+  skip_if_too_deep(5)
+  expect_no_error(TealAppDriver$new(
+    data = simple_teal_data(),
+    modules = transform(example_module(), when = "before")
+  ))
+  app_transform$stop()
+})
+
+testthat::test_that("e2e: transform as decorators", {
```
gogonzo, line 10:
```suggestion
testthat::test_that("e2e: transform(when = 'decorator') doesn't fail the app", {
```

tests/testthat/test-shinytest2-transform.R
```diff
@@ -0,0 +1,141 @@
...
+  skip_if_too_deep(5)
+  expect_no_error(TealAppDriver$new(
+    data = simple_teal_data(),
+    modules = transform(example_module(), when = "decorator")
+  ))
+  app_transform$stop()
+})
+
+testthat::test_that("e2e: transform after decorators", {
```
gogonzo, line 19:
```suggestion
testthat::test_that("e2e: transform(when = 'after') doesn't fail the app", {
```

tests/testthat/test-shinytest2-transform.R
```diff
@@ -0,0 +1,141 @@
...
+  expect_no_error(TealAppDriver$new(
+    data = simple_teal_data(),
+    modules = transform(example_module(), when = "after")
+  ))
+  app_transform$stop()
+})
+
+testthat::test_that("e2e: transform adds wrapped to ids of modules", {
+  skip_if_too_deep(5)
+  app <- TealAppDriver$new(
```
gogonzo, line 30: Id of the module is predictable, we don't need to run two apps to compare identifiers. However this test doesn't need TealAppDriver - this can be tested using testServer, please have a look at `srv_teal` where most of the framework is tested

tests/testthat/test-shinytest2-transform.R
```diff
@@ -0,0 +1,141 @@
...
+  expect_true(any(module_2), label = "Is 'Module 2' modified?")
+  expect_true(all(module_1 | module_2), label = "Are all modules modified?")
+
+  app$stop()
+  app_transform$stop()
+})
+
+
+
+testthat::test_that("e2e: transform modifies arbitrary modules structure", {
```
gogonzo, line 105: This test title is imprecise. It should probably say:
- `transform.teal_modules` adds "wrapped" to all modules server id

llrs-roche:
After discussions I'm closing this PR in favor of #1630 which has a narrower scope

PR #1601, opened 2025-09-24
`disable_sidebar` function

m7pr:
Built on top of an `after` so it mimics `disable_reporter` and `disable_src`.
Applied globally. Removes everything besides the module output.

Maybe we can go one step further and create functions that just hide filter panel, or active data summary or just encodings?

```r
devtools::load_all()

data <- teal_data(IRIS = iris, MTCARS = mtcars)

app_without_sidebar <- init(
  data = data,
  modules = modules(
    example_module(label = "Module 1"),
    example_module(label = "Module 2")
  ) |> disable_sidebar()
)

runApp(app_without_sidebar)
```


gogonzo:
@m7pr Is there any issue linked to this?

gogonzo (changes requested):
Wrong function name - it hides the sidebar not disables. What's the point of hiding sidebar which can be collapsed (generic `bslib::sidebar` feature).

m7pr:
@gogonzo sorry I missed your previous comment. There are no cards related to this. This was just a show off of applications of after function. I think this can be closed as it was rather a POC

PR #1602, opened 2025-09-25
delete wrong @return

gogonzo:
The description is wrong after latest release. I'm removing the description of the `teal_modules` fields as it might inspire users to use it.


llrs-roche:
The description says that it will return a class object but not what this class structure and the help page for teal_modules goes to this same help page. So users won't get more information without exploring the object themselves or the methods available to that class.

PR #1606, opened 2025-09-30
Show/Hide reporter UI when teal UI is generated by `renderUI`

### LINKED ISSUE #1604: [Question]: Reporter Preview Not Included When Using Teal as a Shiny Module in New Version
### What is your question?

With the latest version of teal, when creating an app with teal as a Shiny module, the reporter preview is not included. 
Is there an easy way to add the reporter to the app please?

**Steps to Reproduce:**

Follow the example provided in the vignette: https://insightsengineering.github.io/teal/latest-tag/articles/teal-as-a-shiny-module.html
Create an app using teal as a Shiny module.
Observe that the reporter preview is missing.

### Code of Conduct

### Contribution Guidelines

### Security Policy

vedhav:
Closes #1604

### DIFF
--- a/R/module_teal.R
+++ b/R/module_teal.R
@@ -238,16 +238,26 @@ srv_teal <- function(id, data, modules, filter = teal_slices(), reporter = teal.
         ui = tags$div(validate_ui)
       )
     }
-    if (!is.null(reporter)) {
-      shinyjs::show("reporter_menu_container")
-      reporter$set_id(attr(filter, "app_id"))
-      teal.reporter::preview_report_button_srv("preview_report", reporter)
-      teal.reporter::report_load_srv("load_report", reporter)
-      teal.reporter::download_report_button_srv(id = "download_report", reporter = reporter)
-      teal.reporter::reset_report_button_srv("reset_reports", reporter)
-    } else {
-      removeUI(selector = sprintf("#%s", session$ns("reporter_menu_container")))
-    }
+
+    # Wait for UI to be rendered before showing/hiding reporter UI elements
+    # See: https://github.com/rstudio/shiny/issues/3348
+    observeEvent(
+      once = TRUE,
+      reactiveValuesToList(session$input),
+      {
+        if (!is.null(reporter)) {
+          shinyjs::show("reporter_menu_container")
+          reporter$set_id(attr(filter, "app_id"))
+          teal.reporter::preview_report_button_srv("preview_report", reporter)
+          teal.reporter::report_load_srv("load_report", reporter)
+          teal.reporter::download_report_button_srv(id = "download_report", reporter = reporter)
+          teal.reporter::reset_report_button_srv("reset_reports", reporter)
+        } else {
+          removeUI(selector = sprintf("#%s", session$ns("reporter_menu_container")))
+        }
+      },
+      ignoreInit = TRUE
+    )
 
     datasets_rv <- if (!isTRUE(attr(filter, "module_specific"))) {
       eventReactive(data_signatured(), {
--- a/R/module_teal.R
+++ b/R/module_teal.R
@@ -243,6 +243,7 @@ srv_teal <- function(id, data, modules, filter = teal_slices(), reporter = teal.
     # See: https://github.com/rstudio/shiny/issues/3348
     observeEvent(
       once = TRUE,
+      ignoreInit = TRUE,
       reactiveValuesToList(session$input),
       {
         if (!is.null(reporter)) {
@@ -255,8 +256,7 @@ srv_teal <- function(id, data, modules, filter = teal_slices(), reporter = teal.
         } else {
           removeUI(selector = sprintf("#%s", session$ns("reporter_menu_container")))
         }
-      },
-      ignoreInit = TRUE
+      }
     )
 
     datasets_rv <- if (!isTRUE(attr(filter, "module_specific"))) {
--- a/R/module_teal.R
+++ b/R/module_teal.R
@@ -71,28 +71,26 @@ ui_teal <- function(id, modules) {
   navbar <- ui_teal_module(id = ns("teal_modules"), modules = modules)
   nav_elements <- list(
     withr::with_options(reporter_opts, { # for backwards compatibility of the report_previewer_module$server_args
-      shinyjs::hidden(
-        tags$div(
-          id = ns("reporter_menu_container"),
-          .teal_navbar_menu(
-            label = "Report",
-            icon = "file-text-fill",
-            class = "reporter-menu",
-            if ("preview" %in% getOption("teal.reporter.nav_buttons")) {
-              teal.reporter::preview_report_button_ui(ns("preview_report"), label = "Preview Report")
-            },
-            tags$hr(style = "margin: 0.5rem;"),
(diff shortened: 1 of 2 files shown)


R/module_teal.R
```diff
@@ -238,16 +238,26 @@ srv_teal <- function(id, data, modules, filter = teal_slices(), reporter = teal.
...
+          reporter$set_id(attr(filter, "app_id"))
+          teal.reporter::preview_report_button_srv("preview_report", reporter)
+          teal.reporter::report_load_srv("load_report", reporter)
+          teal.reporter::download_report_button_srv(id = "download_report", reporter = reporter)
+          teal.reporter::reset_report_button_srv("reset_reports", reporter)
+        } else {
+          removeUI(selector = sprintf("#%s", session$ns("reporter_menu_container")))
+        }
+      },
+      ignoreInit = TRUE
```
llrs-roche: A minor comment, maybe place `ignoreInit` argument also at the top with `once`, as I think it is also important for the correct behavior we want.

vedhav: Sure 👍🏽

vedhav:
@llrs-roche  I'll keep the `bindEvent` change for later. We use `observeEvent` in many places and not a single `bindEvent` yet. We can simply change all of them. But, I like the consistency that we use the same thing in all the places right now. And there is no benefit to this migration and `shiny` will always support this function.

R/module_teal.R
vedhav: Done in 1e22dcfbd1d781b12fc20adcda683bf75974f4fc

vedhav:
Let me add a test for this now.

llrs-roche:
The benefit of using `bindEvent` is that it can be cached:
"However, bindEvent() is more flexible: it can be combined with `bindCache()`, and it can also be used with render functions". teal framework speed is a growing concern, being able to use cache could enable faster responsiveness. 

But I agree with your point of not introducing this here 👍(🤔 maybe I should open a more general issue for this).

vedhav:
> The benefit of using `bindEvent` is that it can be cached: "However, bindEvent() is more flexible: it can be combined with `bindCache()`, and it can also be used with render functions". teal framework speed is a growing concern, being able to use cache could enable faster responsiveness.
> 
> But I agree with your point of not introducing this here 👍(🤔 maybe I should open a more general issue for this).

I disagree. `bindEvent` cannot be cached. Only `reactives` can be cached using `bindCache`. Events are always greedy in nature and will always trigger. So, in case of `reactiveEvent` it makes sense to use `bindEvent` so you can chain `bindCache`.


Edit: I see that technically you can use `bindCache` on events. But, I don't see practical use in many cases. For reactives it makes more sense. But, I agree to your point that it becomes more flexible and we should plan to migrate to this in the near future.

vedhav:
@llrs-roche I created an issue for the reporter dropdown issue that exists right now. We can only add test after resolving that: https://github.com/insightsengineering/teal/issues/1607 as it will break the test logic.

averissimo:
~This is very similar to the `data` tab~ _(in a way that there is no need for JS with `removeUI` working fine)_, maybe we should just show the UI by default and remove it if `reporter` is `NULL`.

WDYT?

```diff
diff --git a/R/module_teal.R b/R/module_teal.R
index 75600c9f7..869c94c86 100644
--- a/R/module_teal.R
+++ b/R/module_teal.R
@@ -71,28 +71,26 @@ ui_teal <- function(id, modules) {
   navbar <- ui_teal_module(id = ns("teal_modules"), modules = modules)
   nav_elements <- list(
     withr::with_options(reporter_opts, { # for backwards compatibility of the report_previewer_module$server_args
-      shinyjs::hidden(
-        tags$div(
-          id = ns("reporter_menu_container"),
-          .teal_navbar_menu(
-            label = "Report",
-            icon = "file-text-fill",
-            class = "reporter-menu",
-            if ("preview" %in% getOption("teal.reporter.nav_buttons")) {
-              teal.reporter::preview_report_button_ui(ns("preview_report"), label = "Preview Report")
-            },
-            tags$hr(style = "margin: 0.5rem;"),
-            if ("download" %in% getOption("teal.reporter.nav_buttons")) {
-              teal.reporter::download_report_button_ui(ns("download_report"), label = "Download Report")
-            },
-            if ("load" %in% getOption("teal.reporter.nav_buttons")) {
-              teal.reporter::report_load_ui(ns("load_report"), label = "Load Report")
-            },
-            tags$hr(style = "margin: 0.5rem;"),
-            if ("reset" %in% getOption("teal.reporter.nav_buttons")) {
-              teal.reporter::reset_report_button_ui(ns("reset_reports"), label = "Reset Report")
-            }
-          )
+      tags$div(
+        id = ns("reporter_menu_container"),
+        .teal_navbar_menu(
+          label = "Report",
+          icon = "file-text-fill",
+          class = "reporter-menu",
+          if ("preview" %in% getOption("teal.reporter.nav_buttons")) {
+            teal.reporter::preview_report_button_ui(ns("preview_report"), label = "Preview Report")
+          },
+          tags$hr(style = "margin: 0.5rem;"),
+          if ("download" %in% getOption("teal.reporter.nav_buttons")) {
+            teal.reporter::download_report_button_ui(ns("download_report"), label = "Download Report")
+          },
+          if ("load" %in% getOption("teal.reporter.nav_buttons")) {
+            teal.reporter::report_load_ui(ns("load_report"), label = "Load Report")
+          },
+          tags$hr(style = "margin: 0.5rem;"),
+          if ("reset" %in% getOption("teal.reporter.nav_buttons")) {
+            teal.reporter::reset_report_button_ui(ns("reset_reports"), label = "Reset Report")
+          }
         )
       )
@@ -239,7 +237,6 @@ srv_teal <- function(id, data, modules, filter = teal_slices(), reporter = teal.
       )
     }
     if (!is.null(reporter)) {
-      shinyjs::show("reporter_menu_container")
       reporter$set_id(attr(filter, "app_id"))
       teal.reporter::preview_report_button_srv("preview_report", reporter)
       teal.reporter::report_load_srv("load_report", reporter)
```

vedhav:
@llrs-roche I added a test to check this `e2e: reporter tab is visible when the teal ui creation is delayed` It will fail on the main and pass when you **install** (not load) teal from this branch.

vedhav:
I agree @averissimo now that we know this is the behavior we can embrace it.

tests/testthat/test-shinytest2-reporter.R
```diff
@@ -1,13 +1,49 @@
...
+  testthat::expect_true(
+    unlist(
+      app$get_js(
+        sprintf(
+          "Array.from(document.querySelectorAll('%s')).map(el => el.checkVisibility({}))",
+          "#teal-reporter_menu_container"
+        )
+      )
+    )
+  )
```
llrs-roche, line 46: I'm not sure why `app$is_visible` can't be used here. I guess the appdriver would need to be modified.

vedhav: I see you've resolved the comment. But, for anyone reading this in the future, I'll add the reason. It is because we don't use the `TealAppDriver` for the shinytest2 here because we need to create teal as a shiny module, so we use the `AppDriver` from shinytest2 which does not have this method.

PR #1611, opened 2025-10-03
Always keeps some R code in reports

averissimo:
# Pull Request

- Fixes https://github.com/insightsengineering/teal.reporter/issues/400

### Companion to

- https://github.com/insightsengineering/teal.reporter/pull/414
- https://github.com/insightsengineering/teal/pull/1611
- https://github.com/insightsengineering/teal.modules.general/pull/923

### Open questions

To @donyunardi and @kumamiao 

- Should we always keep in reporter? _(see screenshots below for the sections I mention)_
  - Data pre-processing
  - Filtering

Context:

- The previous version of the reporter had a section that described the active filters.
    - This will be lost if we remove all code.

image

#### Previous filter state

image

### Changes description

- Data pre-processing and Filtering are always kept in report
  - Even when de-selecting option to remove R code

image

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,8 +1,8 @@
 Type: Package
 Package: teal
 Title: Exploratory Web Apps for Analyzing Clinical Trials Data
-Version: 1.0.0.9011
-Date: 2025-10-09
+Version: 1.0.0.9012
+Date: 2025-10-10
 Authors@R: c(
     person("Dawid", "Kaledkowski", , "dawid.kaledkowski@roche.com", role = c("aut", "cre"),
            comment = c(ORCID = "0000-0001-9533-457X")),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,4 @@
-# teal 1.0.0.9011
+# teal 1.0.0.9012
 
 ### New features
 
--- a/R/TealAppDriver.R
+++ b/R/TealAppDriver.R
@@ -1,5 +1,3 @@
-# FilteredData ------
-
 #' Drive a `teal` application
 #'
 #' Extension of the `shinytest2::AppDriver` class with methods for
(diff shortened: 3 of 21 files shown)


m7pr:
This is how it looks for regression module, if you don't include the code (initial data setup code is still visible)

image


and the comparison on how it looks with the code included

image

m7pr (approved):
From the code's perspective, all looks good

kumamiao:
Thanks @m7pr for following up on this one. My comments on the inclusion/exclusion:
1. Whether the app user chooses to include or not include the R code, we should always include the filter state (as we did previously), as this is critical information to inform the app user what is the analysis created. For example, if they did not add in the title/comment, they would not know what population the analysis was applied to. This was specifically requested before if I recall correctly. 
image

2.  If user chooses to not include the R code, let's not show the code for Data pre-processing or Data Filtering. The filter state itself will be sufficient.

gogonzo:
> 2. If user chooses to not include the R code, let's not show the code for Data pre-processing or Data Filtering. The filter state itself will be sufficient.

What if `transformator` is applied with some custom filters? Filter-state won't be sufficient information then about data transformation.  

> This was specifically requested before if I recall correctly.

How important filter state is? Please give us more background what is the meaning of having information which IMO is incomplete. Let's discuss how final report should look like.

m7pr:
Hey @gogonzo we decided together with team today during a catch-up that if code is not showed we will add a small note below the Filter State, that says "further data manipulations and filtering could be applied with transformers for which you did not want to see the code". Stuff like that that suggests that there might be more manipulations. And when code is included, then this message is not shown.

The motivation for Filter State is that people tend to recognize/differentiate various report cards based on the Filter State.

gogonzo:
We decided to do this other way. Thanks @averissimo

PR #1629, opened 2025-10-10
Allow normal shiny app (Teal as a Shiny Module) to use `TealAppDriver`

vedhav:


gogonzo:
@vedha I don't know whether it is needed, maybe we can remove constructor entirely and do this in tests:

```r
app_dr <- shinytest2::AppDriver$new( # or TealAppDriver
  init(
    data = teal_data() |> within(iris <- iris),
    modules = modules(example_module())
  ) |>
    do.call(what = shiny::shinyApp)
)
```

gogonzo:
I also think about `active_ns` - I think we can have a better way to handle this. There is more and more coming to "module container". We have now:

```yaml
base:
  base-wrapper
    - base-reporter_add_container
    - base-src_container
    - base-filter_panel
    - base-data_summary
    - base-module
```  

The list is unlimited. Meanwhile we have: 
- `active_ns()` - returns a list of ids `filter_panel`, `module`, `data_summary`
- `active_module_ns()` - returns character id of `module`
- `active_data_summary_ns()` - returns character id of `data_summary`
- `active_filters_ns()` - returns character id of `filter_panel` (comment: names mismatch) 
- `active_module_element(target)` - returns character id with `#module_ns-target`
... and many more


I think we can limit this to two functions, like: 
- `active_id()` - returns a list of id of all known teal elements (`module`, `wrapper`, `filter_panel`, `data_summary`, `transformators`, `show_r_code`, ...) 
- `active_ns(prefix)` returns `ns` of each known element, so that we can `active_ns()$module("id_in_module")`

----

reason is that `TealAppDriver` has too many methods which are created just to serve particular part of the application. We don't need to create a method to get an `id` or `#selector` of any element in the module-context. Let's make a "factory" method and we will save a lot of code.

gogonzo:
@vedhav to close this one?

m7pr:
still 54 fails in R CMD CHECK

```
  1. └─TealAppDriver$new(...) at test-shinytest2-teal_modifiers.R:28:5
   2.   └─teal (local) initialize(...)
   3.     └─checkmate::assert_class(app$ui, "shiny.tag")
   4.       └─checkmate::makeAssertion(x, res, .var.name, add)
   5.         └─checkmate:::mstop(...)
  ── Error ('test-shinytest2-teal_modifiers.R:43:3'): e2e: modify_header displays custom header in the app ──
  Error in `initialize(...)`: Assertion on 'app$ui' failed: Must inherit from class 'shiny.tag', but has class 'function'.
  Backtrace:
```

https://github.com/insightsengineering/teal/actions/runs/18407812434/job/52451882436?pr=1629#step:44:583

PR #1634, opened 2025-10-14
include teal-slices yaml in the Data filtering section

gogonzo:
Part of:
- https://github.com/insightsengineering/teal.reporter/issues/400

- add teal-slices yaml to the document
- teal-slices are not removed from (downloaded) document output when "Include R code" unchecked
- I've also changed the sections added by teal:
    - `## Code preparation` -> `## Data preparation`
    - `## Data filtering` -> `### Filter settings`, now to include filter-panel-code and teal-slices-yaml

Solution still not perfect when `Include R code` is disabled and not filters added:

Skärmavbild 2025-10-14 kl  15 11 23

### DIFF
--- a/R/teal_data_utils.R
+++ b/R/teal_data_utils.R
@@ -19,7 +19,7 @@ NULL
 #' @rdname teal_data_utilities
 .append_evaluated_code <- function(data, code, filter_states) {
   checkmate::assert_class(data, "teal_data")
-  checkmate::assert_class(filter_states, "teal_slices")
+  checkmate::assert_class(filter_states, "teal_slices", null.ok = TRUE)
   if (length(code) && !identical(code, "")) {
     data@code <- c(data@code, code2list(code))
     teal.reporter::teal_card(data) <- c(


R/teal_data_utils.R
```diff
@@ -17,13 +17,19 @@
...
+  checkmate::assert_class(filter_states, "teal_slices")
   if (length(code) && !identical(code, "")) {
     data@code <- c(data@code, code2list(code))
     teal.reporter::teal_card(data) <- c(
       teal.reporter::teal_card(data),
-      "## Data filtering",
+      "### Filter settings",
+      teal.reporter::code_chunk(
+        .teal_slice_to_yaml(filter_states),
+        lang = "teal-slices",
```
m7pr: Only a minor proposition. End user (app user or report reader) might not be familiar with `teal` internals and might not understand that data filters are called `teal-slices` in teal's implementation,
```suggestion
        lang = "filters",
```

m7pr:
> Solution still not perfect when Include R code is disabled and not filters added:

It's not the end of the world. No code and no filter, so nothing displayed : P empty section. Headline alone indicates that something could be there, if there was a code.

gogonzo:
> > Solution still not perfect when Include R code is disabled and not filters added:
> 
> It's not the end of the world. No code and no filter, so nothing displayed : P empty section. Headline alone indicates that something could be there, if there was a code.

I think the same. We have a simple solution without trying to hack `teal_code` and the way how reporter is rendered

m7pr:
Alrighty, 4 more tests to be fixed and we are gucci.

PR #1635, opened 2025-10-15
Restore reporter cards

gogonzo:
Closes
- insightsengineering/teal.reporter#424 
- insightsengineering/teal.reporter/issues/352

In this PR:
- Reporter restorable with bookmarking (please follow instructions in the issue to reproduce).
- Some tests added to test reporter in teal
- Size of the bookmarking-files minimal 👍 (`tree --du -h shiny_bookmarks`)  
Skärmavbild 2025-10-15 kl  08 33 28

### DIFF
--- a/R/module_teal.R
+++ b/R/module_teal.R
@@ -130,6 +130,10 @@ srv_teal <- function(id, data, modules, filter = teal_slices(), reporter = teal.
       shinyjs::showLog()
     }
 
+    session$onBookmark(function(state) {
+      logger::log_debug("srv_teal@onBookmark: storing report cards")
+      state$values$report_cards <- shiny::isolate(reporter$get_cards())
+    })
     # set timezone in shiny app
     # timezone is set in the early beginning so it will be available also
     # for `DDL` and all shiny modules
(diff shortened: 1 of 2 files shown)


R/module_teal.R
```diff
@@ -246,6 +250,11 @@ srv_teal <- function(id, data, modules, filter = teal_slices(), reporter = teal.
       {
         if (!is.null(reporter)) {
           reporter$set_id(attr(filter, "app_id"))
+          report_cards <- restoreValue(session$ns("report_cards"), NULL)
```
llrs-roche, line 253: While [learning about `onBookmark()` I found `onRestore()`](https://shiny.posit.co/r/reference/shiny/latest/onbookmark.html), should we use that function instead? This works but `onRestore()` seems to be the standard way to do it, specially if we already use `onBookmark()`.

gogonzo: @llrs-roche problem with `onRestore` is that it sets input-values even if they have not been initialized. Some findings are written [here](https://github.com/insightsengineering/coredev-tasks/discussions/532) and in the documentation of `?teal:::restoreValue`

`restoreValue` works in similar way as `shiny::restoreInput` but it can also restore `values`.
From `restoreValue` documentation:
> Bookmarks can store not only inputs but also arbitrary values. These values are stored by onBookmark callbacks and restored by onBookmarked callbacks, and they are placed in the values environment in the session$restoreContext field. Using teal_data_module makes it impossible to run the callbacks because the app becomes ready before modules execute and callbacks are registered. In those cases the stored values can still be recovered from the session object directly.

> Note that variable names in the values environment are prefixed with module name space names, therefore, when using this function in modules, value must be run through the name space function.

PR #1633, opened 2025-10-13
Teal app driver changes

gogonzo:
Please check also:
- https://github.com/insightsengineering/teal.modules.general/pull/927

- Removed `$active_` methods in favour of one `$namespaces()` 
- `namespaces(FALSE)` (default) gives a list of `ns` functions, so that:
      - `namespaces(FALSE)$module("element-id-in-module")`  will return an id of element
      - `namespaces(TRUE)$module("element-id-in-module")`  will return an id of element prefixed by `#`
      - `namespaces()` returns following namespaces: `module`, `filter_panel`, `data_summary`, `wrapper`

PR #1644, opened 2025-10-31
Fixes integration tests

averissimo:
# Pull Request

- Fixes https://github.com/insightsengineering/coredev-tasks/issues/674

### DIFF
--- a/R/TealAppDriver.R
+++ b/R/TealAppDriver.R
@@ -568,7 +568,7 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
       )
       active_base_id <- sub("-wrapper$", "", active_wrapper_id)
 
-      private$ns$module_container <- active_base_id
+      private$ns$base_id <- active_base_id
       private$ns$wrapper <- shiny::NS(active_base_id, "wrapper")
       private$ns$module <- shiny::NS(active_base_id, "module")
       private$ns$filter_panel <- shiny::NS(active_base_id, "filter_panel")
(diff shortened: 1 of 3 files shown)


R/TealAppDriver.R
```diff
@@ -568,6 +568,7 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
       )
       active_base_id <- sub("-wrapper$", "", active_wrapper_id)
 
+      private$ns$module_container <- active_base_id
```
averissimo: @gogonzo @vedhav why was this change implemented?

See the same line in PR https://github.com/insightsengineering/teal/pull/1633/files#diff-5ad41333c93eff54458e60408bd4ee461446d7a846345f9a128d3ed5f85f05d1R571

I see that's it's used in modules, but it breaks a bunch of tests in `teal`.

- `$module_container()` function that no longer exists is used in `tests/testthat/test-shinytest2-reporter.R` below

My proposition is to bring back the `module_container` in cae9d8648 and keep `wrapper` as is

note: I don't understand why CI didn't pick this up as it fails when trying locally with teal `v1.0.0.9015`

gogonzo: > @gogonzo @vedhav why was this change implemented?

`TealAppDriver` was overgrown with methods which were doing the same thing. There was a risk that anything in the "module container" somebody would add another method to test another element in the module-scope.
Replacing id with NS-functions shortened tests where we used to use `paste()` just to find an id of the element in the module-scope. 

> My proposition is to bring back the module_container in cae9d86 and keep wrapper as is

I think removal of `module_container` was by mistake as it was used in [tests in the same PR](https://github.com/insightsengineering/teal/pull/1633/files#diff-5ad41333c93eff54458e60408bd4ee461446d7a846345f9a128d3ed5f85f05d1). Please bring back

PR #1652, opened 2025-11-20
Skip screenshot on non-linux OS

llrs-roche:
# Pull Request

Fixes some internal integration tests that are failing

### DIFF
--- a/tests/testthat/test-shinytest2-init.R
+++ b/tests/testthat/test-shinytest2-init.R
@@ -3,8 +3,10 @@ testthat::skip_if_not_installed("rvest")
 
 testthat::test_that("e2e: teal app initializes with no errors", {
   skip_if_too_deep(5)
-  if (!startsWith(shinytest2::platform_variant(), "linux")) {
-    skip("Skipping test on non-linux platforms.")
+  snaps <- list.dirs(testthat::test_path("_snaps"), recursive = FALSE, full.names = FALSE)
+  platform <- shinytest2::platform_variant()
+  if (!platform %in% snaps) {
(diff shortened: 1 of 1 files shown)


tests/testthat/test-shinytest2-init.R
```diff
@@ -3,6 +3,9 @@ testthat::skip_if_not_installed("rvest")
 
 testthat::test_that("e2e: teal app initializes with no errors", {
   skip_if_too_deep(5)
+  if (!startsWith(shinytest2::platform_variant(), "linux")) {
+    skip("Skipping test on non-linux platforms.")
+  }
```
m7pr: @llrs-roche the platform on integration tests is linux. The R version there 4.4 so `shinytest2::platform_variant()` will return `linux-4.4` on integration tests, and this will be executed. To omit this test on integration tests we need to specifically mark OS distribution and R version
```suggestion
  if (shinytest2::platform_variant() != "linux-4.5")) {
    skip("Skipping test on non-linux platforms.")
  }
```

llrs-roche: Sorry I didn't think on that. I was trying to make a test resilient to changes but I see I missed the point. Thanks I'll modify the logic and update the skip message

m7pr: Or add `app$wait_for_idle()` in this test in between

```
  app$expect_no_shiny_error()
  app$expect_screenshot(selector = "#teal-tabpanel_wrapper")
```

so it's

```
  app$expect_no_shiny_error()
  app$wait_for_idle()
  app$expect_screenshot(selector = "#teal-tabpanel_wrapper")
```

PR #1653, opened 2025-11-24
Follow-up changes to TealAppDriver

m7pr:
Companion to 

- https://github.com/insightsengineering/teal.modules.clinical/pull/1441

Changes
- Changed the way get_active_module_plot_output access namespaces
- Fixed issue where set_active_ns did not set the namespace


R/TealAppDriver.R
```diff
@@ -87,8 +87,8 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
       )
       # end od check
 
-      private$set_active_ns()
       self$wait_for_idle()
+      private$set_active_ns()
```
m7pr, line 91: Maybe this is enough, to first wait and then set the namespace. 
But if not, then the rest of this commit might be needed https://github.com/insightsengineering/teal/pull/1653/commits/f8ea37e384081661d69f16fe69b8adbb02ba060c
but it was produced by cursor, so it needs some review.

osenan: It seems that `wait_for_idle` might not capture alone all the activities of the app. I would keep it like this for the moment. Later on we need to do some performance analysis of e2e tasks.

R/TealAppDriver.R
```diff
@@ -554,18 +553,76 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
+      }
+      wrapper_id
+    },
+    # Helper function to check if wrapper ID is valid
+    is_valid_wrapper_id = function(wrapper_id) {
+      !is.null(wrapper_id) &&
+        length(wrapper_id) > 0 &&
+        !(length(wrapper_id) == 1 && (wrapper_id == "" || is.na(wrapper_id)))
+    },
     set_active_ns = function() {
```
osenan, line 597:
```suggestion
    set_active_ns = function(sleep_time = 0.5) {
```

R/TealAppDriver.R
```diff
@@ -554,18 +553,76 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
+        active_wrapper_id <- private$extract_wrapper_id(
+          sprintf(".teal-modules-tree li a.module-button[data-value='%s']", active_tab_inputs)
+        )
+      }
+
+      # Ensure we have a valid wrapper ID
+      # get_attr returns character(0) when no elements found, or NA_character_ for missing attributes
+      if (!private$is_valid_wrapper_id(active_wrapper_id)) {
+        # Try one more time after a short wait - the page might still be loading
+        Sys.sleep(0.5)
```
osenan:
```suggestion
        Sys.sleep(sleep_time)
```

R/TealAppDriver.R
```diff
@@ -555,17 +555,83 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
+        }
+      }
+
+      # Ensure we have a valid wrapper ID
+      # get_attr returns character(0) when no elements found, or NA_character_ for missing attributes
+      if (is.null(active_wrapper_id) ||
+        length(active_wrapper_id) == 0 ||
+        (length(active_wrapper_id) == 1 && (active_wrapper_id == "" || is.na(active_wrapper_id)))) {
+        # Try one more time after a short wait - the page might still be loading
+        Sys.sleep(0.5)
```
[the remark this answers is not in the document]
osenan: Agree with that comment. Implemented as suggestion

osenan:
@copilot we do not need any new PR, please do not do any action for the moment

R/TealAppDriver.R
```diff
@@ -554,18 +553,77 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
       active_tab_inputs <- all_inputs[grepl("-active_module_id$", names(all_inputs))]
 
-      active_wrapper_id <- sub(
-        "^#",
-        "",
-        self$get_attr(
-          selector = sprintf(".teal-modules-tree li a.module-button[data-value='%s']", active_tab_inputs),
-          attribute = "href"
+      # If no active_module_id input found, find the selected/active tab button directly
+      if (!length(active_tab_inputs) || active_tab_inputs == "") {
```
averissimo: This clause makes more sense if we start with the positive case when `active_tab_inputs` is available

R/TealAppDriver.R
```diff
@@ -554,18 +553,77 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
+      }
+
+      # Ensure we have a valid wrapper ID
+      # get_attr returns character(0) when no elements found, or NA_character_ for missing attributes
+      if (!private$is_valid_wrapper_id(active_wrapper_id)) {
+        # Try one more time after a short wait - the page might still be loading
+        Sys.sleep(sleep_time)
+        active_wrapper_id <- private$extract_wrapper_id(
+          ".teal-modules-tree li a.module-button[href*='-wrapper']:not([href='#'])"
+        )
```
averissimo: It seems to make more sense to sleep and call the method again with `sleep_time = sleep_time - 1` that skips this part if time is negative _(or `NULL` to simply skip it)_.

R/TealAppDriver.R
```diff
@@ -554,18 +553,77 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
+          ),
+          collapse = ", "
+        )
+        stop(sprintf(
+          paste0(
+            "Could not determine active module namespace. ",
+            "Make sure a module tab is selected and the page has finished loading. Found wrapper IDs: %s"
+          ),
+          if (length(found_ids) > 0) found_ids else "none"
+        ))
```
averissimo: This is not good clear and simple code, please refactor and take advantage that `stop` pastes together all its elements.

averissimo:
```suggestion
        stop(
            "Could not determine active module namespace. ",
            "Make sure a module tab is selected and the page has finished loading. Found wrapper IDs: "
            if (length(found_ids) > 0) found_ids else "none"
        ))
```

R/TealAppDriver.R
```diff
@@ -554,18 +553,77 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
       all_inputs <- self$get_values()$input
       active_tab_inputs <- all_inputs[grepl("-active_module_id$", names(all_inputs))]
 
-      active_wrapper_id <- sub(
-        "^#",
-        "",
-        self$get_attr(
-          selector = sprintf(".teal-modules-tree li a.module-button[data-value='%s']", active_tab_inputs),
-          attribute = "href"
+      # If no active_module_id input found, find the selected/active tab button directly
```
averissimo: I don't get why would DOM have the input id if the input does not exist in shiny.

How would that work?

- Is it because of `bslib` doing some magic? This creates additional overhead when/if the navigation selectors are changed.

llrs-roche: I am not sure I follow your thought. Note that `self$get_attr` is also used by `extract_wrapper_id`

R/TealAppDriver.R
```diff
@@ -554,19 +553,50 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
+      length(wrapper_id) >= 1 & wrapper_id != "" & !is.na(wrapper_id)
+    },
     set_active_ns = function() {
+      # Although wait_for_idle() is called before set_active_ns(), it only ensures Shiny is not processing.
+      # wait_for_page_stability() is needed here to ensure the DOM/UI is fully rendered and stable
+      # before trying to extract the namespace.
+      private$wait_for_page_stability()
+
       all_inputs <- self$get_values()$input
       active_tab_inputs <- all_inputs[grepl("-active_module_id$", names(all_inputs))]
```
averissimo: We could use [AppDriver$wait_for_value](https://rstudio.github.io/shinytest2/reference/AppDriver.html?q=get_download#method-wait-for-value-) to make sure we have a value and then discard the rest of the added complexity.

The `ignore = list(NULL, "")` argument will allow us to avoid any of the initialization

spoiler... same can be done with javascript code and `AppDriver$wait_for_js` (let's say for  `is_visible()`)... I'm trying it out in widgets.

averissimo: Done in 041818a

R/TealAppDriver.R
```diff
@@ -554,19 +553,52 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
       filter_panel = character(0)
     ),
     # private methods ----
+    # Helper function to extract wrapper ID from selector and take first match if multiple found
```
llrs-roche: Comment is not up to date. We no longer take the first one
```suggestion
    # Helper function to extract wrapper ID from selector
```

R/TealAppDriver.R
```diff
@@ -554,19 +553,52 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
+    extract_wrapper_id = function(selector) {
+      id <- self$get_attr(selector = selector, attribute = "href")
+      sub("^#", "", id[endsWith(id, "-wrapper")])
+    },
+    # Helper function to check if wrapper ID is valid
+    is_valid_wrapper_id = function(wrapper_id) {
+      length(wrapper_id) >= 1 & wrapper_id != "" & !is.na(wrapper_id)
+    },
     set_active_ns = function() {
-      all_inputs <- self$get_values()$input
```
llrs-roche, line 558: Without `all_inputs` the `active_module_input_id` on line 571 won't work.
I would restore this line

R/TealAppDriver.R
```diff
@@ -554,19 +553,52 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
+        #   ".teal-modules-tree li a.module-button.active, .teal-modules-tree li a.module-button[aria-selected='true']"
+        # ),
+        # private$extract_wrapper_id(
+        #   ".teal-modules-tree li a.module-button[href*='-wrapper']:not([href='#'])"
+        # )
+        # nolint end: commented_code.
       )
-      active_base_id <- sub("-wrapper$", "", active_wrapper_id)
+      validity_ids <- private$is_valid_wrapper_id(unique(ids))
+      valid_ids <- unique(ids[validity_ids])
```
llrs-roche: It is a bit confusing to have two unique call. Maybe we could have an extra line for that:

```suggestion
      ids <- unique(ids)
      validity_ids <- private$is_valid_wrapper_id(ids)
      valid_ids <- ids[validity_ids]
```

But note that this is only required if there are multiple calls to `extract_wrapper_id` on `ids <- c()`

R/TealAppDriver.R
```diff
@@ -554,19 +553,52 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
+      if (length(valid_ids) > 1L) {
+        valid_ids <- valid_ids[1L]
+      } else if (length(valid_ids) < 1L) {
+        stop(
+          "Could not determine valid module namespace. ",
+          "Make sure a module tab is selected and the page has finished loading."
+        )
+      }
+
+      active_base_id <- sub("-wrapper$", "", valid_ids)
```
llrs-roche, line 637: This line is a duplicate of [L559](https://github.com/insightsengineering/teal/pull/1653/changes#diff-5ad41333c93eff54458e60408bd4ee461446d7a846345f9a128d3ed5f85f05d1R559) 

Simple suggestion to not modify L603-L604 that are out of the diff:

```suggestion
      active_base_id <- valid_ids
```

llrs-roche (commented):
I haven't tested yet but I left several comments to simplify the code. I'll test it and update the status of the review

R/TealAppDriver.R
averissimo: Awesome, I made the change with a direct call when setting up the ids

R/TealAppDriver.R
averissimo: Thanks! Too many cooks on this PR :-)

R/TealAppDriver.R
averissimo: I don't think it is, as the line in 559 only removes the `#` from the start of the id

llrs-roche: True, my bad.

llrs-roche:
Let's merge this and tests if the integration tests work with the changes.
@averissimo you'll need to approve it to be merged or are you waiting on further review on the other PR on TMC?

PR #1666, opened 2025-12-09
Adds CSS visibily expecation to `TealAppDriver` that use JS and wait accordingly

averissimo:
…ediate execution

# Pull Request

Implements `expectation` for visibility using `AppDriver$wait_for_js` API that is more robust than immediate execution.


R/TealAppDriver.R
```diff
@@ -649,6 +683,31 @@ TealAppDriver <- R6::R6Class( # nolint: object_name.
...
+                                  visibility_property) {
+
+      testthat::skip_if_not(
+        self$get_js("typeof Element.prototype.checkVisibility === 'function'"),
+        "Element.prototype.checkVisibility is not supported in the current browser."
+      )
+
+      action <- match.arg(action)
+      js_code <- sprintf(
+        "Array.from(document.querySelectorAll('%s')).map(el => el.checkVisibility({%s, %s, %s})).some(Boolean)",
```
osenan, line 706: question, should we print any message to the js console? Maybe in case of error? For debugging purposes. On the other hand, it needs to be consistent with other testing js functions, if we do not have console messages, I guess it does not make sense now

osenan: Probably not, as this is only used in headless browsers in e2e tests

averissimo:
@osenan can you test it again?

I triggered the [CI manually](https://github.com/insightsengineering/teal/actions/runs/20274059637) and all the shinytest2 pass (for both this and `bring_tests` branches). The linter error is from files that this PR does not change.

image

osenan (approved):
Hi, after running the last changes locally, it worked!! Good job!

PR #1667, opened 2025-12-10
Modify Copilot setup to use native R installation

m7pr:
Updated the GitHub Actions workflow to use native R installation instead of a Docker container, ensuring better compatibility for Copilot's setup. Added permissions for repository access and included steps for setting up R and installing dependencies.

Our containers, and other Rocker containers are missing tools that Copilot expects, so it's better to go with a native environment where we install R and R dependencies on our own.

Part of 
- https://github.com/insightsengineering/coredev-tasks/issues/694


m7pr:
@llrs-roche yes, that's basically that but without the check : )

PR #1670, opened 2025-12-12
Add AGENTS.md documentation for the teal framework

m7pr:
Solves

- https://github.com/insightsengineering/coredev-tasks/issues/677

A file with guidelines file for Agents from Cursor / Claude / GitHub Copilot.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -340,6 +340,7 @@ Do not change versions on your own.
 ### GitHub Workflows
 
 Use `r.pkg.template` workflows for consistency:
+
 - **check.yaml**: R CMD check, unit tests, coverage
 - **docs.yaml**: Documentation building and deployment  
 - **audit.yaml**: Security and dependency auditing


m7pr:
Locally, VS Code used the file in GitHub Copilot Chat
image

GitHub Copilot Coding Agent also uses this file
image

Cursor locally also uses it

image

m7pr:
@copilot  do you see AGENT.md file? would this be the file you would pick as agent guidelines?

pawelru:
- for CLAUDE: create a new `CLAUDE.md` (and add to buildignore) with the following content:
```
# Read AGENTS.md

@AGENTS.md
```

(alternatively `.claude/CLAUDE.md` and change into `@../AGENTS.md`)

- for GHCopilot - create a symbolic link from `AGENTS.md` to `.github/instructions/agents.instructions.md`

m7pr:
@pawelru GitHub Copilot says that it will pick .github/instructions/agents.instructions.md but if this file does not exist it will next look for AGENTS.md

https://docs.github.com/en/enterprise-cloud@latest/copilot/tutorials/coding-agent/get-the-best-results

image

m7pr:
Feedback from Pawel (translated from polish to English)

Looks great!

There are many places I would either cut out or greatly simplify: tidyverse style guide, pre-commit, roxygen2
You're explaining a lot of things that pre-commit/auto-formatter would catch. Instead, you can tell it to run pre-commit every time and fix any problems - these errors are quite informative, and it seems you don't need to explain exactly what it does.
I'm missing file extensions - so it can find the right script faster - you only have that at the package level.
I'm also missing design 

https://code.roche.com/llm/rexpert_api/-/blob/56_agents/AGENTS.md?ref_type=heads
Here's my attempt - you can see the file structure.

I also created three documents:
AGENTS.md - always added
AGENTS_ARCHITECTURE.md
AGENTS_DEVELOPMENT.md
These two additional ones are optional - the agent knows this exists and will only open it when it deems it useful.

AGENTS.md
```diff
@@ -0,0 +1,427 @@
+# Teal Ecosystem R Package Development Guide
+
+## Introduction
+
+The teal framework uses shiny to create reproducible environments for analysts. The ecosystem comprises several interconnected packages with specific roles:
+
+### Core Packages
```
osenan, line 7: is there any advantage for coding agents to include the url to repos or are they able to find it by their own?

AGENTS.md
```diff
@@ -0,0 +1,427 @@
...
+# Good: Use explicit returns for clarity
+srv_my_module <- function(id, data) {
+  moduleServer(id, function(input, output, session) {
+    # Module logic
+    return(reactive_result)
+  })
+}
+```
+
+### Line Length and Formatting
```
osenan: cannot we point directly at our .lintr configuration? This is storing information at two different places. If we need to update .lintr we would need to update here as well. I would simply refer as:
### Formatting
Formatting rules are configured in the .lintr file

AGENTS.md
```diff
@@ -0,0 +1,427 @@
...
+  ggplot2::theme_minimal()
+```
+
+## Testing Framework
+
+### Testing Philosophy
+- **Test public functions only**: Internal utilities should be tested through public interfaces
+- **Precise, focused tests**: Each test should verify one specific behavior
+- **High coverage**: Maintain at least 80% test coverage as measured by `covr`
+- **Integration over units**: Test realistic usage patterns
```
osenan, line 220:
```suggestion
- **Integration over units**: Test realistic usage patterns
- **Test Dependencies**.: Add `testthat::skip_if_not_installed(package_name)` only for dependencies in SUGGESTS or related to tests cases
```

AGENTS.md
```diff
@@ -0,0 +1,427 @@
...
+  testthat::expect_error(
+    my_function("not a data frame"),
+    class = "checkmate_error"
+  )
+})
+```
+
+### Shiny Module Testing
+- **Server functions**: Test with `shiny::testServer()`
+- **UI functions**: Test only through `TealAppDriver` for integration
```
osenan: Is this correct? We have many tests for UI functions where we check the class, if there is an error generating the component, snapshot tests, regexp search, all applied to UI functions and outside of e2e tests. For example in `teal.widgets`

AGENTS.md
```diff
@@ -0,0 +1,427 @@
...
+  # Test UI elements are present
+  testthat::expect_true(driver$is_visible("#plot"))
+  
+  driver$stop()
+})
+```
+
+### Test Organization and Naming
+- **One test file per R file**: `test-module_example.R` for `module_example.R`
+- **Descriptive test names**: Clearly describe what is being tested
```
osenan, line 300:
```suggestion
- **Descriptive test names**: Clearly describe what is being tested
-**End to end test names**: `test-shinytest2-module_example.R` for `module_example.R`
```

AGENTS.md
```diff
@@ -0,0 +1,427 @@
...
+Use `r.pkg.template` workflows for consistency:
+- **check.yaml**: R CMD check, unit tests, coverage
+- **docs.yaml**: Documentation building and deployment  
+- **audit.yaml**: Security and dependency auditing
+- **pkgdown.yaml**: Website generation
+
+### Pre-commit Hooks
+Configure `.pre-commit-config.yaml` for code quality:
+
+```yaml
```
osenan: do we need to specify an example yaml here? Could not we refer to the current repo `.pre-commit-config.yaml` file?

osenan (changes requested):
Hi, I really liked this agents.md file. I think it will be very useful for our coding agents to do a better job.
Apart from the specific comments two more comments:

1. Does it make sense to create a section for PR Instructions like in [here](https://agents.md/)?
2. Maybe a section about branch naming conventions
3. Can we put this agent.md in top of project not in specific repo? To be applied to all
4. Does i

AGENTS.md
m7pr: if repos are pulled locally and included in the workspace, then no urls are needed : )

but for github copilot it would be beneficial to have urls!

AGENTS.md
m7pr: good idea and reduces repetition!

AGENTS.md
m7pr: @osenan what do you mean by `or related to tests cases`?

AGENTS.md
m7pr: ok, we can rephrase to try testing basic usage with regular testing, but test UI scenarios with TealAppDriver that is based on shinytest2::AppDriver

AGENTS.md
m7pr: Better to refer to actual pre-commit. yup

AGENTS.md
osenan: I mean if for any test we need specific package like `mockery` or similar that we do not have in Suggests DESCRIPTION

m7pr:
@osenan 

> Can we put this agent.md in top of project not in specific repo? To be applied to all repos

Yeah, still trying to figure out if we should have an extra repository for AGENTS.md and devcontainer file.

AGENTS.md
m7pr: ok, now it's clear. thanks

AGENTS.md
m7pr: urls are included in DESCRIPTION files so should be good

AGENTS.md
m7pr: https://github.com/insightsengineering/teal/pull/1670/commits/8ba62c8d82968c36791c06eac30a945b507f38e4

AGENTS.md
m7pr: https://github.com/insightsengineering/teal/pull/1670/commits/97f02d865235af15970994327bef9476e7e4ddbb

AGENTS.md
m7pr: this was removed

osenan (approved):
Thanks for addressing all the comments. I think now it really looks better.

PR #1675, opened 2025-12-17
Make sure same location of the modules elements

gogonzo:
To make sure that selector will always point to the list of modules.


gogonzo:
> Should a similar comment be added on TealAppDriver?

Not needed. It is hard to determine where in TealAppDriver comment should be added

PR #1676, opened 2025-12-19
Fixes integration tests

averissimo:
# Pull Request

- Quiet verbosity of `reporter_previewer_module` deprecation message
- Update test to shiny new version that outputs elements with whitespace

### DIFF
--- a/tests/testthat/test-reporter_previewer_module.R
+++ b/tests/testthat/test-reporter_previewer_module.R
@@ -1,4 +1,6 @@
 testthat::describe("reporter_previewer_module", {
+  withr::local_options("lifecycle_verbosity" = "quiet")
+
   testthat::it("returns teal_module with previewer class", {
     module <- reporter_previewer_module(label = "test")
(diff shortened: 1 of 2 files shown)


tests/testthat/test-reporter_previewer_module.R
```diff
@@ -1,4 +1,6 @@
 testthat::describe("reporter_previewer_module", {
+  withr::local_options("lifecycle_verbosity" = "quiet")
```
llrs-roche, line 2: Should we use this on other places?

llrs-roche (approved):
Thanks for the quick PR

tests/testthat/test-reporter_previewer_module.R
averissimo: Only when strictly required to reduce repetitve code (`lifecycle::expect_deprecated` wrapper). I think it's safe to use it here as this file will be deleted once the function is deprecated.

We shouldn't use it often as we might miss some deprecations messages from other packages

PR #1679, opened 2026-01-09
Prints `AppDriver` logs during tests when GitHub CI is run with `enable debug logging`

averissimo:
…s when stopping app

# Pull Request

Part of 

### Changes description

- Implements `stop()` method in appdriver that will create a message with the logs when CI is run with `enable debug logging`
- Adds missing documentation


osenan (approved):
The changes look ok to me. It is hard to test it in my machine without the actions in place...
Let's merge it and see what is the result in the CI and if it helps.

PR #1684, opened 2026-01-22
1683 ignore duplicated slice

gogonzo:
another PR https://github.com/insightsengineering/teal.slice/pull/675

closes https://github.com/insightsengineering/teal/issues/1683

Problem arises when trying to simplify teal-app creation from independent modules. Let's image following situation, we have a group of templates where each template has data, slices and module functions. Functions for each template are independent and they are not linked with each other. For example slices for template `labs1` and `labs2` are "independent" set of filters. Sometimes these filters are common filters and they should be shared by the modules - currently teal doesn't provide this, because duplicated filter id-s are raising an error. It makes sense that they will be combined and duplicated ones will be ignored.

```r
ts_labs1 <- function() {
  teal::teal_slices(
    teal.slice::teal_slice("ds1", "var1", id = "labs_filter"),
    mapping = list(
      "lab1_module" = "labs_filter" 
    ),
    module_specific = TRUE
  )
}

ts_labs2 <- function() {
  teal::teal_slices(
    teal.slice::teal_slice("ds1", "var1", id = "labs_filter"),
    teal.slice::teal_slice("ds1", "var2", id = "labs2_filter"),
    mapping = list(
      "lab2_module" = c("labs_filter", "labs2_filter")
    ),
    module_specific = TRUE
  )
}

c(
  ts_labs1(), ts_labs2()
)

# {
#   "slices": [
#     {
#       "dataname"      : "ds1",
#       "varname"       : "var1",
#       "id"            : "labs_filter",
#       "fixed"         : false,
#       "anchored"      : false,
#       "multiple"      : true
#     },
#     {
#       "dataname"      : "ds1",
#       "varname"       : "var2",
#       "id"            : "labs2_filter",
#       "fixed"         : false,
#       "anchored"      : false,
#       "multiple"      : true
#     }
#   ],
#   "attributes": {
#     "allow_add"       : true,
#     "mapping"         : {
#       "lab1_module"   : "labs_filter",
#       "lab2_module"   : ["labs_filter", "labs2_filter"]
#     },
#     "module_specific" : true
#   }
# } ```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -97,7 +97,7 @@ LazyData: true
 Roxygen: list(markdown = TRUE, packages = c("roxy.shinylive"))
 RoxygenNote: 7.3.3
 Remotes:
-    insightsengineering/teal.slice@1683_ignore_duplicated_slice
+    insightsengineering/teal.slice@main
 Collate:
     'TealAppDriver.R'
     'after.R'

### CODE REVIEW

File: R/teal_slices.R [hunk 1]
```diff
@@ -179,3 +142,29 @@ deep_copy_filter <- function(filter) {
...
+#' @keywords internal
+#'
+as.teal_slices <- .copy_from_namespace("as.teal_slices", "teal.slice") # nolint: object_name_linter.
+
+
+#' @rdname teal_slices
+#' @export
+#' @keywords internal
+#'
+c.teal_slices <- .copy_from_namespace("c.teal_slices", "teal.slice")
```
averissimo: Love the removal of duplicated code, but I don't like evaluating code on the fly just to import it, could we do it in a native way instead?

You don't have to accept my suggestion, but I think the namespace should be the `teal` for consistency

Image

```suggestion
#' Function to copy functions from other namespaces
#'
#' Useful when we require function from other namespace
#' where this function calls other functions from `teal` namespace (see `as.teal_slices`, `c.teal_slices`).
.copy_from_namespace <- function(fun, envir = environment(deep_copy_filter)) {
  environment(fun) <- envir
  fun
}

#' @rdname teal_slices
#' @export
#' @keywords internal
#'
as.teal_slices <- .copy_from_namespace(teal.slice::as.teal_slices) # nolint: object_name_linter.

#' @rdname teal_slices
#' @export
#' @keywords internal
#'
c.teal_slices <- .copy_from_namespace(utils::getS3method("c", "teal_slices", envir = asNamespace("teal.slice")))
```

R/teal_slices.R
```diff
@@ -179,3 +142,29 @@ deep_copy_filter <- function(filter) {
     filter_copy
   })
 }
+
+
+#' Function to copy functions from other namespaces
+#'
+#' It differs from `getFromNamespace` as it returns a copy of the function
+#' so it is a part of `teal` namespace. Useful when we require function from other namespace
```
llrs-roche: Why do we need to have these functions here and not in teal.slice ? What are the advantages?

R/teal_slices.R
llrs-roche: Why do we need to have these functions here and not in teal.slice ? I see that it fails without importing c.teal_slices (not sure why) but we can reexport as.teal_slices and the test pass (if c.teal_slices is re-loaded/implemented on teal). 

```
#' @export
teal.slice::as.teal_slices
```

averissimo: `teal_slices()` needs to be called within `teal` environment: `teal::teal_slices()` and not `teal.slice::teal_slices()`

llrs-roche: Ah, I see now. Thanks! Could `teal.slice::teal_slices()` be a method with the default what is on `teal.slices::teal_slices.default <- teal.slices::teal_slices` and with a method on teal `teal::teal_slices.teal_slice <- teal::teal_slices`. I think this could avoid these problems with the namespace and keep both with the same name.

averissimo: That would partially make it work! Then, as soon as `teal` is loaded the `teal_slices` function would be overwritten from the `teal` namespace.

However, the `as.teal_slices` is not an S3 method and would still suffer from this. Or can we register a `as.teal_slices` function that accepts `as.teal_slices.list` and `as.teal_slices.default`?

-------

I think this requires a bigger reflection on this mechanism as a whole.

It seems we need 2 objects, a `slice_raw` and `slices_raw` in `teal.slice` package and then the and `teal_slices` in `teal` (as to not break the most commonly used API).

But I'm sure there's a reason for this (the question is if that is still valid? :thinking:)

llrs-roche: We could make `as.teal_slices` as a S3 method. That's the purpose of methods: having different behavior with similar syntax imho. Yes, method can have multiple classes: we could register `as.teal_slices.list` and `as.teal_slices.default`. But I haven't tested this so there still might be some minor hiccups.

gogonzo: This is a good idea and `as.` should always be an S3 method. I'll have a look in `teal.slice` if it doesn't break anything 👍

gogonzo: @llrs-roche decided not to register `as.teal_slices.list` and `as.teal_slices.default` as I would have to copy both to `teal` and I'm worried about the conflicts in the methods table.

gogonzo:
> Looks good! There's on open comment, that may be moved to a discussion topic

Let's not waste time if it covers problem for now. I just made a small improvement we can improve in the future if somebody touches these lines again

PR #1687, opened 2026-02-06
feat: enable open telemetry to capture the current module

osenan:
# Pull Request

This partially closes [this issue](https://github.com/insightsengineering/coredev-tasks/issues/709).

To test it, please do the following:

1. Setup the docker collector [as described here](https://opentelemetry.io/docs/collector/quick-start/)
2. Run the collector:
```
docker run \
  -p 127.0.0.1:4317:4317 \
  -p 127.0.0.1:4318:4318 \
  -p 127.0.0.1:55679:55679 \
  otel/opentelemetry-collector:0.145.0 \
  2>&1 | tee collector-output.txt # Optionally tee output for easier search later
```
3. Create this `.Renviron` in the same folder you would execute your teal apps
```
OTEL_LOGS_EXPORTER=http
OTEL_LOG_LEVEL=info
SHINY_OTEL_COLLECT=session
```
We tell Open Telemetry to only capture session information until [we fix the issue](https://github.com/insightsengineering/coredev-tasks/issues/720) between teal framework and open telemetry.

Then install this teal branch version.
Run a sample app:

```
library(teal)
library(teal.modules.general)

data <- teal_data()
data <- within(data, {
  data <- CO2
})

app <- init(
  data = data,
  modules = modules(
    tm_data_table(
      label = "Data Table",
      variables_selected = list(data = c("Plant", "Type", "Treatment", "conc", "uptake"))
    ),
    tm_rmarkdown(
      label = "RMarkdown Module",
      rmd_content = c(
        "---",
        "title: \"R Markdown Report\"",
        "output: html_document",
        "---",
        "",
        "```{r}",
        "summary(CO2) |> print()",
        "```"
      )
    )
  )
)

shinyApp(app$ui, app$server)

```
Run the app and try changing the teal module. It should be recorded in the logs.
Then look at the collector output, if you follow the example it will be called `collector-output.txt` or the name you have edited.

Check that in the collector output we have recorded the module.


DESCRIPTION
```diff
@@ -64,11 +64,14 @@ Suggests:
     knitr (>= 1.42),
     mirai (>= 1.1.1),
     MultiAssayExperiment,
+    otel(>= 0.2.0),
+    otelsdk(>= 0.2.2),
```
averissimo, line 68: Do we need this?

DESCRIPTION
```diff
@@ -64,11 +64,14 @@ Suggests:
...
     R6,
     renv (>= 1.0.11),
     rmarkdown (>= 2.23),
     roxy.shinylive,
     rvest (>= 1.0.0),
+    shiny (>= 1.12.1),
```
averissimo, line 74: `shiny` is on depends

DESCRIPTION
llrs-roche: otel yes as it is used

R/module_nested_tabs.R
```diff
@@ -107,6 +107,10 @@ srv_teal_module <- function(id,
                             reporter = teal.reporter::Reporter$new(),
                             data_load_status = reactive("ok")) {
   moduleServer(id, function(input, output, session) {
+    observeEvent(input$active_module_id, {
+      otel::log_info(sprintf("The app current module is %s", input$active_module_id))
```
llrs-roche, line 111: otel is in suggested, code using it should be conditional of otel being loaded:

```suggestion

     if (requireNamespace("otel", quiet = TRUE)) otel::log_info(sprintf("The app current module is %s", input$active_module_id))
```
We  might want to report the id and the label as the label might be more human readable while the id is more for machines.

DESCRIPTION
osenan: Yes, I can change it but will force `teal` to have the last version of shiny

DESCRIPTION
averissimo: `otelsdk` in particular :sweat_smile: it's not being used in the PR

DESCRIPTION
averissimo: I mean it's on line 40

Image

llrs-roche: @averissimo but the version is lower. Oriol is saying that the new requirements would force users to have the latest shiny version. As we know not all environments where teal is used have the latest versions of packages.

R/module_nested_tabs.R
```diff
@@ -107,6 +107,10 @@ srv_teal_module <- function(id,
                             reporter = teal.reporter::Reporter$new(),
                             data_load_status = reactive("ok")) {
   moduleServer(id, function(input, output, session) {
+    observeEvent(input$active_module_id, {
```
averissimo, line 110: ### 1. Where does  `teal.logger` stand, would adding `otel` support to `logger` be a better choice  (controlled by `teal` options). 

We already have a logging infrastructure, and adding this parallel method seems redundant. 

In particular with the question

> What is the criteria for the framework/app/module developer to use `otel` vs. `teal.logger`?

This approach could save all `INFO` level to otel (when otel is enabled).

The verbosity of INFO+ levels isn't that bad as only `messages()` calls generate at this level IIRC _(other than in t.m.helios)_

### 2. Would a new API like `wrap_server()` be a better fit for this use case?

My main question is if we should support this type of logging natively on `teal` and static? I can imagine that there are different use cases to extend past by the module that's being used.

Let's say for an application that is deployed on connect, we may want to also log the username to assess if the modules usage are overrepresented by power users.

I would like to resurface the idea of a API that extends the `init()` call to allow for custom server/ui logic for the whole of teal without the need of using it as module.

```r
my_data <- teal.data::teal_data(IRIS = iris)
teal::init(data = my_data, modules  = teal::example_module()) |>
  teal::transform_init(
    srv = function(id) {
          if (otel::is_tracing_enabled()) {
           ospan <- otel::get_active_span()
           ospan$set_attribute("user_role", "admin")
           ospan$set_attribute("app_version", "1.2.3")
           ospan$set_attribute("app_name", "best_teal_app")
         }
        shiny::observeEvent(session$input[["teal-teal_modules-active_module_id"]], { # namespace depends on where we run this modifier
        otel::log_info(sprintf("The app current module is %s", input$active_module_id))
      })
    },
    ui = NULL # nothing to add here
  )
```

averissimo, line 110: ### 3. How can we trace which application is being logged?

How can we trace the actual application name? Skimming from the documentation I see that there's an API for attributes.

How would that work with this implementation?

Image

DESCRIPTION
averissimo: Can we have duplicate entries on the dependencies with different versions?

DESCRIPTION
osenan: It is not used by teal code by the dependency is [necessary for Open Telemetry](https://shiny.posit.co/r/articles/improve/opentelemetry/#adding-opentelemetry-integration) to work.

R/module_nested_tabs.R
llrs-roche: 1) I agree that we should reuse the infrastructure we already have. Reporting all info to otel would be good. 
2) I like the idea of being able to know what modules each user uses. Isn't this proposed API similar to the `after()` function but applied to the whole app?

DESCRIPTION
llrs-roche: Yes, from the [WRE](https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#Package-Dependencies-1)
> A package or ‘R’ can appear more than once in the ‘Depends’ field, for example to give upper and lower bounds on acceptable versions.

DESCRIPTION
llrs-roche: otelsdk is a dependency of otel and it is only needed to collect telemtry so I am not sure why it is needed here

R/module_nested_tabs.R
gogonzo, line 110: `active_module_id` won't give us a clear picture about what module is used as this is a derivative of a `module$label`. This could be anything. For Example I can use `tm_g_km` and call it "My awesome module" - input will be `my_awesome_module` (or something like this). 
If we want to know how often a module is used I would suggest to implement this in `teal.modules.*` (where teal.modules.* logs are implemented wrongly :D).

We can't accept the fact of having two log-frameworks (`logger` and `otel`)

donyunardi: I agree, it should capture the module's label, the module's function name, and adding more to the scope, the name of the teal app.

PR #1689, opened 2026-02-09
Scope Bootstrap tweaks to teal root + Shiny modal

### LINKED ISSUE #1688: [Bug]:  Host app affected by Bootstrap overrides in custom.css
### What happened?

 We set Bootstrap variables in inst/css/custom.css, example:

https://github.com/insightsengineering/teal/blob/c7bd748f93875fdd47750a0bcda8bd2131c1cd7e/inst/css/custom.css#L33

Because these are global CSS overrides, they leak into any host app that embeds `teal`.

### Code of Conduct

### Contribution Guidelines

### Security Policy

AleKoure:
# Pull Request

Fixes #1688 

- Problem: Bootstrap var tweaks in inst/css/custom.css leaked into host apps.
- Tradeoff: We scope them to .teal-root to avoid host leakage, but Shiny modals render under #shiny-modal-wrapper (outside .teal-root).
- Fix: Apply the same rules to #shiny-modal so teal modals keep consistent styling while limiting impact to the host app (only the modal container).

Longer‑term: we may consider a cleaner way to target teal components and move these values into bs_theme() or more formal app tokens.


AleKoure:
> Hi, thanks for your solution! I was reviewing your PR, you pointed out the right things to fix. However, while testing it I found that customization trough teal options (it calls bslib theming) was not working as expected. [See Scenario 1 in the PR description](https://github.com/insightsengineering/teal/pull/1701). I followed a similar approach than yours, only allowing the teal developer to set if they require a css class for the teal module container to facilitate creation of css selectors. For the package css, I refactored and moved some parts trough the default bslib theming. Once approved it should fix the bug you created

I agree, theming is not working as expected in the current version of **teal**.

The example in my PR used a simple workaround to demonstrate the issue, but the correct approach as I mentioned in the description is to handle this consistently through  `bs_theme()`, as implemented in the PR you linked.

With this change, theming becomes more robust and predictable, which should improve both teal itself and downstream teal embeddings!

PR #1690, opened 2026-02-10
Report which module is in use with teal.logger

osenan:
# Pull Request

This partially will show the information on which is the current active module.
Test this app, switch the teal module and verify that the logger reports which module we are.

```
library(teal)
library(teal.modules.general)
library(teal.modules.mdr)
library(shiny.telemetry)

data <- teal_data()
data <- within(data, {
  data <- CO2
})

logger::log_threshold(logger::INFO, namespace = "teal")

app <- init(
  data = data,
  modules = modules(
    tm_data_table(
      label = "Data Table",
      variables_selected = list(data = c("Plant", "Type", "Treatment", "conc", "uptake"))
    ),
    tm_rmarkdown(
      label = "RMarkdown Module",
      rmd_content = c(
        "---",
        "title: \"R Markdown Report\"",
        "output: html_document",
        "---",
        "",
        "```{r}",
        "summary(CO2) |> print()",
        "```"
      )
    )
  )
)

shinyApp(app$ui, app$server)
```

### DIFF
--- a/R/module_nested_tabs.R
+++ b/R/module_nested_tabs.R
@@ -107,13 +107,17 @@ srv_teal_module <- function(id,
                             reporter = teal.reporter::Reporter$new(),
                             data_load_status = reactive("ok")) {
   moduleServer(id, function(input, output, session) {
+    # Send message to JS to get document title
     session$sendCustomMessage("teal-get-document-title", list(inputId = session$ns("app_title")))
 
-    observeEvent(input$active_module_id, {
+    # Wait for both title and active_module_id to be available
+    observeEvent(list(input$active_module_id, input$app_title), {
+      req(input$app_title)
+
       app_title <- if (!is.null(input$app_title) && nzchar(input$app_title)) {
         input$app_title
       } else {
-        "teal app"
+        "Teal Application"
       }
 
       message(


R/module_nested_tabs.R
```diff
@@ -107,6 +107,10 @@ srv_teal_module <- function(id,
                             reporter = teal.reporter::Reporter$new(),
                             data_load_status = reactive("ok")) {
   moduleServer(id, function(input, output, session) {
+    observeEvent(input$active_module_id, {
+      logger::log_info(sprintf("The app current module is %s", input$active_module_id))
+    })
```
averissimo:
```suggestion
      message(sprintf("The app current module is %s", input$active_module_id))
    })
```

Would it be better to use `message()` here?

That is picked by `teal.logger` [when registers handlers](https://github.com/insightsengineering/teal.logger/blob/1a6dd44bbdc6572ee3d5bf4af62ace1d01612c0b/R/register_handlers.R#L37) and then `logger::log_info` being called

averissimo: We could even use the extra `domain` argument in `message()` to flag it as `otel` if necessary

image

Debugging `teal.logger/R/register_handlers.R` line `79` (under `handler_fun <- function(m) {`)

image

osenan: Hi thanks for the suggestions.
About the domain, it's a good idea. However I prefer to implement anything related Open Telemetry on a separate PR.
About using message, on one hand it's something more common for R to put messages rather than call `log_info`. On the other hand, if we use `log_info` we can control whether to log or not the message with the `log_threshold`. If we use `message` or `warning` function we need to separately control the display of such messages or the inclusion into the logs by R options or functions like `suppressMessages` or suppressWarnings`.

I have tested locally replacing `log_info` by `message` in the branch and then changing the `log_threshold` to `WARN` in the provided example app. If we use `log_info` the current module id is not printed. If we use `message` is printed.

To control everything from a single parameter (log_treshold) I am more inclined to choose `log_info`.

averissimo: The right way of setting the log level with `teal.logger` is with an R option

`options("teal.log_level" = "WARN")`

With this it surppresses all outputs that comes from `message()` or `logger::log_info("something" , namespace = "teal")`

averissimo: Running your sample app with it, it only shows the module initialization (which is run when the modules are created, before the teal app starts)

```diff
diff --git a/R/module_nested_tabs.R b/R/module_nested_tabs.R
index 945d69f20..831610484 100644
--- a/R/module_nested_tabs.R
+++ b/R/module_nested_tabs.R
@@ -108,7 +108,7 @@ srv_teal_module <- function(id,
                             data_load_status = reactive("ok")) {
   moduleServer(id, function(input, output, session) {
     observeEvent(input$active_module_id, {
-      logger::log_info(sprintf("The app current module is %s", input$active_module_id))
+      message(sprintf("The app current module is %s", input$active_module_id))
     })
 
     .srv_teal_module(
```

Image

averissimo: If I'm not mistaken, since https://github.com/insightsengineering/teal/pull/1008 that we primarly use base R for `message()` / `warning()` in the framework, the goal is that module developers don't need to rely/learn a logger package and `namespaces`.

My suggestion to use `message()` is precisely as it's the same as doing `logger::log_info` with the current `teal.logger` framework.

>  About the domain, it's a good idea. However I prefer to implement anything related Open Telemetry on a separate PR.

👍

osenan: I tried again your suggestion but it works differently to me.
the `teal.log_level` option only affects `logger::log_*` functions. And the module is printed to the logger.
I have read the [PR](https://github.com/insightsengineering/teal/pull/1008#issue-2039909626) you mentioned and it says specifically that `warning` or `message` functions are independent of logger levels (which make sense to me).

osenan:
iI've added information about the app title in the last commits. It was requested on similar PR to add module ui i \ function, but that cannot be obtained by changing teal, we should change the module function itself. Besides, I don't think is that helpful

R/module_nested_tabs.R
osenan: Hi, I finally changed logger::log_info to message. It makes sense as we are doing the same with other info calls in modules and teal. Thank you for your effort on explaining why this was done

R/module_nested_tabs.R
```diff
@@ -107,6 +107,24 @@ srv_teal_module <- function(id,
...
                             data_load_status = reactive("ok")) {
   moduleServer(id, function(input, output, session) {
+    session$sendCustomMessage("teal-get-document-title", list(inputId = session$ns("app_title")))
+
+    observeEvent(input$active_module_id, {
+      app_title <- if (!is.null(input$app_title) && nzchar(input$app_title)) {
+        input$app_title
+      } else {
+        "teal app"
+      }
```
averissimo, line 121: This is not working properly,

I'm trying to change the title and still appears as `teal app`, am I missing something?

```r
pkgload::load_all("../teal")
init(
  modules = modules(
    example_module("mod1"),
    example_module("mod2")
  )
) |> modify_title("yada") |> shiny::runApp()
```

averissimo: It's running fine in browser, but `input` is not changing

Firefox console
Image

R console (with browser statetment)
Image

averissimo: This only happens for the first module, maybe it's too fast for input to change

osenan: good catch! It was actually the JS that was not fully loaded. I tried to modify the JS code but it was not working. Then I realized there was a simpler solution by changing the observer and requiring the input title not to be NULL, exactly once the JS is executed.

Please try with this app:
```
options("teal.log_level" = "INFO")  # Set to INFO to see navigation logs
library(teal)
library(teal.modules.general)
getOption("teal.log_level")  # Set to INFO to see navigation logs

data <- teal_data()
data <- within(data, {
  data <- CO2
})

app <- init(
  data = data,
  modules = modules(
    tm_data_table(
      label = "Data Table",
      variables_selected = list(data = c("Plant", "Type", "Treatment", "conc", "uptake"))
    ),
    tm_rmarkdown(
      label = "RMarkdown Module",
      rmd_content = c(
        "---",
        "title: \"R Markdown Report\"",
        "output: html_document",
        "---",
        "",
        "```{r}",
        "summary(CO2) |> print()",
        "```"
      )
    )
  )
 ) |>
   modify_title("My custom title")

shinyApp(app$ui, app$server)
```

averissimo: I think you forgot to push the changes @osenan

osenan: True, pushed now

PR #1697, opened 2026-02-25
Decorators to teal

gogonzo:
Concludes discussion in https://github.com/insightsengineering/coredev-tasks/issues/727

Other PRs:
- https://github.com/insightsengineering/teal.modules.general/pull/987
- https://github.com/insightsengineering/teal.modules.clinical/pull/1467

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: teal
 Title: Exploratory Web Apps for Analyzing Clinical Trials Data
-Version: 1.1.0.9024
+Version: 1.1.0.9025
 Date: 2026-03-02
 Authors@R: c(
     person("Dawid", "Kaledkowski", , "dawid.kaledkowski@roche.com", role = c("aut", "cre"),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,8 +1,9 @@
-# teal 1.1.0.9024
+# teal 1.1.0.9025
 
 ### Enhancements
 
 * Improved print method for `teal_module` object (#1685).
+* Added opt-in URL-based module navigation, controlled by `options(teal.enable_deep_linking = TRUE)` (default: `FALSE`). When enabled, the active module is reflected in the URL as `?active_module=<module path>`, and navigating to such a URL or using the browser's back/forward buttons switches to the corresponding tab.
 
 # teal 1.1.0
 
--- a/R/module_nested_tabs.R
+++ b/R/module_nested_tabs.R
@@ -85,7 +85,7 @@ ui_teal_module <- function(id, modules) {
     tags$ul(
       id = ns("active_module_id"),
       style = "align-items: center; gap: 1em; font-size: large;",
-      class = "teal-navbar nav shiny-tab-input", # to mimic nav and mimic tabsetPanel
+      class = "teal-navbar nav shiny-tab-input", # important! to mimic nav and mimic tabsetPanel
       `data-tabsetid` = "test",
       .teal_navbar_menu(
         !!!module_items$link,
@@ -129,6 +129,32 @@ srv_teal_module <- function(id,
       )
     })
 
+    if (isTRUE(getOption("teal.enable_deep_linking", FALSE))) {
+      observeEvent(input$active_module_id, {
+        current_query <- shiny::parseQueryString(session$clientData$url_search)
+        if (!identical(current_query[["active_module"]], input$active_module_id)) {
+          shiny::updateQueryString(
+            paste0("?active_module=", input$active_module_id),
+            mode = "push",
+            session = session
+          )
+        }
+      })
+      observeEvent(session$clientData$url_search,
+        {
+          query <- shiny::parseQueryString(session$clientData$url_search)
+          if (
+            !is.null(query[["active_module"]]) &&
+              !identical(query[["active_module"]], isolate(input$active_module_id))
+          ) {
+            # updateTabsetPanel works here because the navbar <ul> has class "shiny-tab-input"
+            # (see ui_teal_module), which makes Shiny treat it as a tabsetPanel input.
+            shiny::updateTabsetPanel(session, "active_module_id", selected = query[["active_module"]])
+          }
+        },
+        ignoreNULL = FALSE
+      )
+    }
     .srv_teal_module(
       id = "nav",
       data = data,
--- a/tests/testthat/test-module_teal.R
+++ b/tests/testthat/test-module_teal.R
@@ -3950,3 +3950,136 @@ testthat::describe("teal-src", {
     )
   })
 })
+
+testthat::describe("srv_teal URL navigation", {
+  # local_url_search(): temporarily overrides session$clientData$url_search for the
+  # duration of the calling test, using withr::defer for cleanup.
+  #
+  # `$.mockclientdata` lives only in the shiny namespace (not the S3MethodsTable), so
+  # plain assign() does not affect S3 dispatch. registerS3method() must be used for
+  # both patching and restoration. withr::defer() ties the cleanup to the test frame,
+  # matching the style of local_mocked_bindings().
+  local_url_search <- function(url_val, .local_envir = parent.frame()) {
+    original <- utils::getS3method("$", "mockclientdata")
+    registerS3method("$", "mockclientdata", function(x, name) {
+      if (name == "url_search") {
+        return(url_val)
+      }
+      original(x, name)
+    }, envir = getNamespace("shiny"))
+    withr::defer(
+      registerS3method("$", "mockclientdata", original, envir = getNamespace("shiny")),
+      envir = .local_envir
+    )
+    invisible()
+  }
+
+  withr::local_options(list(teal.enable_deep_linking = TRUE), .local_envir = parent.frame())
+
+  testthat::it("calls updateTabsetPanel with the URL module when active_module differs from current tab", {
+    tab_calls <- list()
+    local_url_search("?active_module=mod2")
+    testthat::local_mocked_bindings(
+      updateTabsetPanel = function(session, inputId, selected = NULL) { # nolint
+        tab_calls[[length(tab_calls) + 1]] <<- list(inputId = inputId, selected = selected)
+      },
+      .package = "shiny"
+    )
+    shiny::testServer(
+      app = srv_teal,
+      args = list(
+        id = "test",
+        data = reactive(teal_data(iris = iris)),
+        modules = modules(
+          module("mod1", server = function(id, data) NULL),
+          module("mod2", server = function(id, data) NULL)
+        )
+      ),
+      expr = session$flushReact()
+    )
+    testthat::expect_length(tab_calls, 1L)
+    testthat::expect_equal(tab_calls[[1]]$selected, "mod2")
+  })
+
+  testthat::it("does not call updateTabsetPanel when URL has no active_module parameter", {
+    tab_calls <- list()
+    local_url_search("?other=foo")
+    testthat::local_mocked_bindings(
+      updateTabsetPanel = function(session, inputId, selected = NULL) { # nolint
+        tab_calls[[length(tab_calls) + 1]] <<- list(inputId = inputId, selected = selected)
+      },
+      .package = "shiny"
+    )
+    shiny::testServer(
+      app = srv_teal,
+      args = list(
+        id = "test",
+        data = reactive(teal_data(iris = iris)),
+        modules = modules(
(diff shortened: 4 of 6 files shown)


R/decorators-helpers.R
```diff
@@ -0,0 +1,98 @@
+#' Check decorators list
+#'
+#' Check if a decorators list is valid and matches the expected output names.
+#'
+#' @param x (named `list`) of `teal_transform_module` objects, or nested lists thereof.
+#' @param names (`character`) optional vector of valid output names. When provided, all names
+#'   in `x` must be either `"default"` or one of these names, and names must be unique.
```
averissimo:
```suggestion
#'   in `x` must be one of these names, and names must be unique.
```

R/decorators-helpers.R
```diff
@@ -0,0 +1,98 @@
...
+#'
+#' @return `TRUE` if valid, otherwise a `character(1)` string describing the problem.
+#'
+#' @export
+check_decorators <- function(x, names = NULL) { # nolint: object_name.
```
averissimo, line 18: To address your concern about usage of flat list of `teal_transform_module`, it's easily done by the module developor with `checkmate::assert_list(x, types = "teal_transform_module)`.

We could change the name of the function to be `check_decorators_named`, `check_decorators_list` or similar

R/module_transform_data.R
```diff
@@ -1,22 +1,83 @@
...
 #' @rdname module_transform_data
-ui_transform_teal_data <- function(id, transformators, class = "well") {
+#' @details
+#' `ui_transform_teal_data` is a thin wrapper around the internal `.ui_transform_teal_data`.
+#' @param class `r lifecycle::badge("deprecated")` No longer used.
+#' @param ... additional arguments passed to `.ui_transform_teal_data` (e.g. `class`).
+#' @return A `list` of `bslib::accordion` UI elements, one per transformator, or `NULL` if
+#'   `transformators` is empty.
+#' @export
+ui_transform_teal_data <- function(id, transformators, class = lifecycle::deprecated(), ...) {
```
averissimo: We could get away without the dot prefix function, as the `ui` wrapper does nothing new. Do you want to keep it for "mirror" reasons (1 srv -> 1 ui)?

R/module_transform_data.R
```diff
@@ -1,22 +1,83 @@
...
+#' @param expr (`expression` or `reactive`) optional expression evaluated on top of the decorated
+#'   output. Useful for post-processing after all transformators have run.
+#' @param modules `r lifecycle::badge("deprecated")` No longer used.
+#' @param is_transform_failed `r lifecycle::badge("deprecated")` No longer used.
 #' @export
+srv_transform_teal_data <- function(id,
+                                    data,
+                                    transformators,
+                                    modules = lifecycle::deprecated(),
+                                    is_transform_failed = lifecycle::deprecated(),
```
averissimo: Does this need an additional check in code to warn users of deprecation?

https://lifecycle.r-lib.org/articles/communicate.html#arguments

R/module_transform_data.R
averissimo: Why not keep the `class` argument here?

If not it may need a deprecation call

```r
  if (!identical(class, "well") {
    lifecycle::deprecate_warn(
      when = "1.2.0", 
      what = "ui_transform_teal_data(class)",
      details = "Ability to choose class will be removed in the next release."
    )
  }
```

averissimo (commented):
Minor comments, looks good

R/decorators-helpers.R
```diff
@@ -0,0 +1,98 @@
...
+  checkmate::assert_string(scope, null.ok = FALSE)
+  if (scope %in% names(decorators)) {
+    result <- decorators[[scope]]
+    if (inherits(result, "teal_transform_module")) {
+      result <- list(result)
+    }
+    result
+  } else {
+    list()
+  }
```
llrs-roche: I think with @averissimo we simplified this on TMG to:
```suggestion
decorators[names(decorators) %in% scope]
```
Do we still think this is the best way?

R/decorators-helpers.R
```diff
@@ -0,0 +1,98 @@
...
+#' @param x (named `list`) of `teal_transform_module` objects, or nested lists thereof.
+#' @param names (`character`) optional vector of valid output names. When provided, all names
+#'   in `x` must be either `"default"` or one of these names, and names must be unique.
+#'
+#' @return `TRUE` if valid, otherwise a `character(1)` string describing the problem.
+#'
+#' @export
+check_decorators <- function(x, names = NULL) { # nolint: object_name.
+
+  check_message <- checkmate::check_list(x, names = "named")
```
llrs-roche, line 20: Should we enforce unique names ? 

```suggestion
  check_message <- checkmate::check_list(x, names = "unique")
```

R/module_transform_data.R
```diff
@@ -1,22 +1,83 @@
...
+      }
+    }
+  })
+}
+
+
 #' @rdname module_transform_data
-ui_transform_teal_data <- function(id, transformators, class = "well") {
+#' @details
+#' `ui_transform_teal_data` is a thin wrapper around the internal `.ui_transform_teal_data`.
```
llrs-roche: This line doesn't explain much to the end user. What is `.ui_transform_teal_data` ?

R/module_transform_data.R
```diff
@@ -1,22 +1,83 @@
...
 #'
 #' @name module_transform_data
 NULL
 
+#' @rdname module_transform_data
+#' @param expr (`expression` or `reactive`) optional expression evaluated on top of the decorated
+#'   output. Useful for post-processing after all transformators have run.
+#' @param modules `r lifecycle::badge("deprecated")` No longer used.
+#' @param is_transform_failed `r lifecycle::badge("deprecated")` No longer used.
 #' @export
```
llrs-roche, line 67: There are no examples of the functionality.

R/decorators-helpers.R
```diff
@@ -0,0 +1,98 @@
...
+#'
+#' Check if a decorators list is valid and matches the expected output names.
+#'
+#' @param x (named `list`) of `teal_transform_module` objects, or nested lists thereof.
+#' @param names (`character`) optional vector of valid output names. When provided, all names
+#'   in `x` must be either `"default"` or one of these names, and names must be unique.
+#'
+#' @return `TRUE` if valid, otherwise a `character(1)` string describing the problem.
+#'
+#' @export
```
llrs-roche, line 11: There aren't any examples of how to use this function.

llrs-roche (commented):
Sorry I didn't see @averissimo reviewed. But I think our comments are complimentary. 

One additional task/extra PR is to modify decorators vignette on teal.

R/decorators-helpers.R
gogonzo: If we do, the line 17 will always be FALSE. We prefer meaningful message instead of `checkmate`'s generic one.

R/decorators-helpers.R
llrs-roche: I simplified, less code to maintain and understand. All checks passed fine with the change.

R/decorators-helpers.R
llrs-roche: Added some examples

llrs-roche:
I updated the code to reflect the renaming from default to all. This should make tests pass.
Other changes:
 - Simplify a bit `select_decorators()`
 - `select_decorators()` now picks also `all` decorators.
 - `select_decorators()` now checks decorators to comply with our convention on `assert_decorators()`
 - The documentation of the decorators is linked now. 
  
 @averissimo I kept the named list so I haven't addressed https://github.com/insightsengineering/teal/pull/1697#discussion_r2852754440. Let us know what you think.

R/module_transform_data.R
```diff
@@ -26,17 +68,33 @@ NULL
...
                                     data,
                                     transformators,
-                                    modules = lifecycle::deprecated(),
-                                    is_transform_failed = lifecycle::deprecated(),
+                                    modules = NULL,
+                                    is_transform_failed = reactiveValues(),
                                     expr) {
   checkmate::assert_class(data, classes = "reactive")
 
+  if (!missing(modules)) {
```
averissimo, line 76:
```suggestion
  if (!is.null(modules)) {
```

R/module_transform_data.R
```diff
@@ -26,17 +68,33 @@ NULL
...
+    lifecycle::deprecate_warn(
+      when = "1.2.0",
+      what = "srv_transform_teal_data(modules)",
+      details = "Argument is used only in the internal of teal and will be removed in the next release."
+    )
+  }
+
+  if (!missing(is_transform_failed)) {
```
averissimo, line 84:
```suggestion
  if(!identical(is_transform_failed, reactiveValues()) {
```

R/module_transform_data.R
```diff
@@ -60,23 +118,38 @@ srv_transform_teal_data <- function(id,
...
-#' @details
-#' `ui_transform_teal_data` is a thin wrapper around the internal `.ui_transform_teal_data`.
 #' @param class `r lifecycle::badge("deprecated")` No longer used.
 #' @param ... additional arguments passed to `.ui_transform_teal_data` (e.g. `class`).
 #' @return A `list` of `bslib::accordion` UI elements, one per transformator, or `NULL` if
 #'   `transformators` is empty.
 #' @export
-ui_transform_teal_data <- function(id, transformators, class = lifecycle::deprecated(), ...) {
+ui_transform_teal_data <- function(id, transformators, class = "well", ...) {
+  if (!missing(class)) {
```
averissimo, line 133:
```suggestion
  if (!identical(class, "well")) {
```

R/decorators-helpers.R
```diff
@@ -2,13 +2,19 @@
...
+#'   in `x` must be one of these names, and names must be unique.
 #'
 #' @return `TRUE` if valid, otherwise a `character(1)` string describing the problem.
-#'
+#' @seealso [module_transform_data()]
 #' @export
+#' @examples
+#' decorator <- teal_transform_module(server = function(id, data) data)
+#' check_decorators(decorator)
+#' check_decorators(list(all = decorator))
```
averissimo, line 15: Are we bring back `all`? Do we ever use it as of now?

llrs-roche: I didn't find any mention to it or default. But it can be handy if a module only generates plots we can then apply a theme to all plots  in a shorter way, than having to `(plot1 = theme, plot2 = theme)`

R/decorators-helpers.R
llrs-roche: Is this something that we need to address? I think we settled on lists.

R/module_transform_data.R
gogonzo: I don't understand why you suggest this? We want to deprecate `class` at all. Your suggestions means that if somebody calls `ui_transform_teal_data(class = "well")` there will be no deprecation warning

R/module_transform_data.R
```diff
@@ -73,15 +73,15 @@ srv_transform_teal_data <- function(id,
                                     expr) {
   checkmate::assert_class(data, classes = "reactive")
 
-  if (!missing(modules)) {
+  if (!is.null(modules)) {
```
gogonzo: do we deprecate an argument or not? It means if somebody explicitly `modules = NULL` then there will be no deprecation message.

R/module_transform_data.R
```diff
@@ -73,15 +73,15 @@ srv_transform_teal_data <- function(id,
...
     lifecycle::deprecate_warn(
       when = "1.2.0",
       what = "srv_transform_teal_data(modules)",
       details = "Argument is used only in the internal of teal and will be removed in the next release."
     )
   }
 
-  if (!missing(is_transform_failed)) {
+  if (rlang::is_missing(is_transform_failed)) {
```
gogonzo: I don't understand why for modules it is `if (!is.null(modules))` and for this `rlang::is_missing(is_transform_failed)`. And why `rlang::is_missing` over `missing`?

llrs-roche: I was trying to follow André suggestions. However, after checking the help page and thinking more carefully I agree with you and I'll revert the changes back to `missing()` on all cases.

R/module_transform_data.R
averissimo: Sorry, my brain froze here. I was under the impression that `!missing(class)` would always be `TRUE` here.

```r
foo <- function(id, transformators, class = "well", ...) {
  if (!missing(class)) {
    print("message shows")
  }
  "end"
}
foo(1, 2)
#> "end"
```

disregard the suggestion

averissimo (approved):
💯

PR #1699, opened 2026-02-27
Enable modules deep linking

### LINKED ISSUE #1698: [Feature Request]: Navigate teal modules
### Feature description

Make teal app to navigate between the modules. How it should work:
- when switching a module url should change to url-encoded module-path. 
- when changing, providing a link app should switch to this module

gogonzo:
closes #1698

Test with `teal_data_module`

```r
library(teal)
options(`teal.enable_deep_linking` = TRUE)

data_module <- teal_data_module(
  ui = function(id) tags$div(actionButton(NS(id, "go"), "Go")),
  server = function(id) {
      reactive({
        req(input$go)
        data <- within(
          teal_data(),
          {
            dataset1 <- iris
            dataset2 <- mtcars
          }
        )
        data
      })
    })
  }
)


app <- init(
  data = data_module,
  modules = modules(
    example_module(label = "mod1"),
    example_module(label = "mod2")
  )
)


shinyApp(app$ui, app$server)
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,8 +1,8 @@
 Type: Package
 Package: teal
 Title: Exploratory Web Apps for Analyzing Clinical Trials Data
-Version: 1.1.0.9023
-Date: 2026-02-27
+Version: 1.1.0.9024
+Date: 2026-03-02
 Authors@R: c(
     person("Dawid", "Kaledkowski", , "dawid.kaledkowski@roche.com", role = c("aut", "cre"),
            comment = c(ORCID = "0000-0001-9533-457X")),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,4 @@
-# teal 1.1.0.9023
+# teal 1.1.0.9024
 
 ### Enhancements
 
--- a/R/teal_data_module.R
+++ b/R/teal_data_module.R
@@ -77,7 +77,14 @@
 #' @seealso [`teal.data::teal_data-class`], [teal.code::qenv()]
 #'
 #' @export
-teal_data_module <- function(ui, server, label = "data module", once = TRUE) {
+teal_data_module <- function(ui = function(id) NULL,
+                             server = function(id) {
+                               moduleServer(id, function(input, output, session) {
+                                 reactive(teal.data::teal_data())
+                               })
+                             },
+                             label = "data module",
+                             once = TRUE) {
   checkmate::assert_function(ui, args = "id", nargs = 1)
   checkmate::assert_function(server, args = "id", nargs = 1)
   checkmate::assert_string(label)
--- a/man/teal_data_module.Rd
+++ b/man/teal_data_module.Rd
@@ -10,7 +10,17 @@
 \alias{within.teal_data_module}
 \title{Data module for \code{teal} applications}
 \usage{
-teal_data_module(ui, server, label = "data module", once = TRUE)
+teal_data_module(
+  ui = function(id) NULL,
+  server = function(id) {
+     moduleServer(id, function(input, output, session) {
+      
+      reactive(teal.data::teal_data())
+     })
+ },
+  label = "data module",
+  once = TRUE
+)
 
 \S4method{eval_code}{teal_data_module}(object, code)
 
--- a/tests/testthat/test-module_teal.R
+++ b/tests/testthat/test-module_teal.R
@@ -152,6 +152,23 @@ testthat::describe("srv_teal arguments", {
       "Must be a reactive"
     )
   })
+
+
+  testthat::it("app works with default teal_data_module", {
+    testthat::expect_no_error(
+      shiny::testServer(
+        app = srv_teal,
+        args = list(
+          id = "test",
+          data = teal_data_module(),
+          modules = modules(example_module())
+        ),
+        expr = {
+          session$flushReact()
+        }
+      )
+    )
+  })
 })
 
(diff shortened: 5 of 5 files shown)


R/module_nested_tabs.R
```diff
@@ -129,6 +129,30 @@ srv_teal_module <- function(id,
       )
     })
 
+    if (isTRUE(getOption("teal.enable_deep_linking", FALSE))) {
```
llrs-roche, line 132: This new option is not documented on `teal-options.Rmd` vignette.

R/module_nested_tabs.R
```diff
@@ -129,6 +129,30 @@ srv_teal_module <- function(id,
...
+      observeEvent(input$active_module_id, {
+        current_query <- shiny::parseQueryString(session$clientData$url_search)
+        if (!identical(current_query[["active_module"]], input$active_module_id)) {
+          shiny::updateQueryString(
+            paste0("?active_module=", input$active_module_id),
```
llrs-roche, line 137: Should we see `?active_module=` on the browser url? Because I don't see it. 
Even if I don't see it adding it to `http://127.0.0.1:6977/?active_module=mod2` didn't redirect me to mod2:

Image

R/module_nested_tabs.R
```diff
@@ -129,6 +129,30 @@ srv_teal_module <- function(id,
...
+        }
+      })
+      observeEvent(session$clientData$url_search,
+        {
+          query <- shiny::parseQueryString(session$clientData$url_search)
+          if (
+            !is.null(query[["active_module"]]) &&
+              !identical(query[["active_module"]], isolate(input$active_module_id))
+          ) {
+            shiny::updateTabsetPanel(session, "active_module_id", selected = query[["active_module"]])
```
llrs-roche, line 152: I am not sure the modules are `tabsetPanels`, tests pass and the functionality seems to work but I can't find a mention to `tabsetPanels()` on teal.

donyunardi:
The original post did not mention `options(teal.enable_deep_linking = TRUE)` which is needed to enable this.

I don't know what "options(`teal.navigate_modules` = TRUE)"  is. It doesn't seem to be needed.

This is what I see and it seems to be working. 

Kapture 2026-02-27 at 18 22 44

R/module_nested_tabs.R
gogonzo: Good point. I will add a comment explaining this 👍

R/module_nested_tabs.R
gogonzo: did you set option before running an app?

R/module_nested_tabs.R
gogonzo: will add, thanks 👍

gogonzo:
> I don't know what "options(teal.navigate_modules = TRUE)" is

@dony My mistake, I've posted PR first and then reflected about option name precision and changed to `enable_deep_linking` (deep linking is a common term for such behaviour)

R/module_nested_tabs.R
gogonzo: Added 👍

R/module_nested_tabs.R
gogonzo: Added 👍

R/module_nested_tabs.R
llrs-roche: I did, but probably the wrong option. As now I tried again (after pulling) and I could see it and it works nicely.

R/module_nested_tabs.R
llrs-roche: I'm marking as resolved with the new comment, but I'mwondering if we should document that too on the UI part where the class is added. So that in the future we know that updating/changing those classes affects beyond the UI and to the functionality.

gogonzo:
> I tested if the "Snapshot Manager" saved this parameter from the url and it did not. I think it would be great if the saved snapshots could also point to the specific module they want users to focus on.

Snapshot manager doesn't handle restoring anything else than filters. Do you mean bookmarking maybe? Bookmarking should restore the state

> The .modules_breadcrumb path is used instead of the label/id of the module, which results on the whitespaces around the / being encoded on the url. Maybe this could be documented on the option.

I don't have a good answer for this. I don't think people should treat url-parameters as interactive input. URI must be encoded to ASCII-US and it is only to share. You can change `active_module=<??> ` but it is not what it meant to be. URL is to share so that one can start with given module open on init.

R/module_nested_tabs.R
gogonzo: There is a comment in the UI part. I've added `important!` before the comment so that nobody removes it in the ui 👍

gogonzo:
@llrs-roche 

I've just checked bookmarking doesn't collide with deep linking:

```r
options(
  `teal.enable_deep_linking` = TRUE,
  `shiny.bookmarkStore` = "server"
)

data_module <- teal_data_module(
  ui = function(id) tags$div(actionButton(NS(id, "go"), "Go")),
  server = function(id) {
      reactive({
        req(input$go)
        data <- within(
          teal_data(),
          {
            dataset1 <- iris
            dataset2 <- mtcars
          }
        )
        data
      })
    })
  }
)


app <- init(
  data = data_module,
  modules = modules(
    example_module(label = "mod1"),
    example_module(label = "mod2")
  )
)


shinyApp(app$ui, app$server, enableBookmarking = "server")
```

PR #1711, opened 2026-04-23
Add "Reset active module" button to snapshot manager (#1707)

### LINKED ISSUE #1707: [Feature Request]: Snapshot Manager Reset Initial State enhancement
### Feature description

The "Reset Initial State" button of the snapshot manager resets the filter panel for all modules.

If module specific (`module_specific = TRUE`) filtering is used then could we have an additional button to reset the filter panel of only the active module?

Thanks

### Code of Conduct

### Contribution Guidelines

### Security Policy

stefanlinner:
Fixes #1707
TrackB_RMedicine_Hackathon_Apr2026

## Summary

When `module_specific = TRUE`, the snapshot manager now has an additional button ("Reset active module") that resets the filter panel of only the currently active module back to its initial state. The existing "Reset initial state" button continues to reset all modules at once.

## What changed

The new button (a rotate-left icon) appears in the snapshot manager modal, next to the existing undo icon. It is only rendered when `module_specific = TRUE` filtering is enabled, since it has no meaning in global filtering mode.

Clicking the button restores both the mapping (which filters are active) and the filter values for the active module to their initial application state, while leaving all other modules untouched.

To make this work, `srv_teal_module` now exposes the `active_module_id` reactive in its return value, which gets threaded through to the snapshot manager.

## Files changed

| File | What |
|------|------|
| `R/module_nested_tabs.R` | `srv_teal_module` returns a list including `active_module_id` (+ updated `@return` docs) |
| `R/module_teal.R` | Extracts `active_module_id` from the new return value and passes it to `srv_snapshot_manager_panel` |
| `R/module_snapshot_manager.R` | Adds conditional "Reset active module" button UI, accepts `active_module_id` parameter, implements the per-module reset logic |
| `tests/testthat/test-module_teal.R` | Two new tests: (1) reset only affects the active module's filters, (2) button is not rendered when `module_specific = FALSE` |

## Notes

If a filter slice is shared across modules (e.g. via `global_filters` in the mapping), resetting the active module will also restore that slice's value for all other modules that use it. This is consistent with how the existing "Reset initial state" button works.

### DIFF
--- a/R/module_nested_tabs.R
+++ b/R/module_nested_tabs.R
@@ -55,9 +55,13 @@
 #'   `id` of the currently active module. This helps to determine which module can listen to reactive events.
 #'
 #' @return
-#' Output of currently active module.
-#' - `srv_teal_module.teal_module` returns `reactiveVal` containing output of the called module.
-#' - `srv_teal_module.teal_modules` returns output of modules in a list following the hierarchy of `modules`
+#' `srv_teal_module` returns a `list` with:
+#' - `modules_output`: output of the module tree (see below).
+#' - `active_module_id`: `reactive` returning the `character(1)` path of the currently active module.
+#'
+#' `.srv_teal_module` dispatches on `modules` class:
+#' - `.srv_teal_module.teal_module` returns `reactiveVal` containing output of the called module.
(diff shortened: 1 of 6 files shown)


stefanlinner:

PR #1713, opened 2026-05-08
Improve on `validate_input` and exports it

averissimo:
# Pull Request

### Changes description

- [x] Resets message under input when it's changed to prevent outdated errors
  - See screencast on first comment
- [x] ~Allows for multiple validation using `add` parameter _(instead of 1 by 1)_~
  - ~Similar implementation to `checkmate::assert_***(..., add = collection)`~
  - Needs alternate solution
- [x] Exports `validate_input` function
- [x] Fixes initialization problem

### Example app with some errors:

- Validate `picks` with non-empty
- Both inputs cannot choose the same letter

```r
#
pkgload::load_all("../teal.picks")
pkgload::load_all("../teal")

data <- within(teal_data(), letters <- data.frame(letter = head(LETTERS)))

picks_letters <- picks(
  datasets("letters", "letters"),
  variables("letter", "letter"),
  values(selected = NULL)
)

my_module <- module(
  label = "My Module",
  datanames = NULL,
  ui = function(id) {
    ns <- NS(id)
      encoding = tagList(
        tags$h3("Letters"),
        teal.picks::picks_ui(ns("letters"), picks_letters),
        tags$h3("Letters2"),
        checkboxGroupInput(ns("letters2"), "Select letters:", choices = head(LETTERS), inline = TRUE),
      ),
      output = tagList(
        tags$h3("Sample plot"),
        plotOutput(ns("plot"))
      )
    )
  },
  server = function(id, data) {
      selectors <- list(letters = picks_letters)
      anl_inputs <- picks_srv(picks = selectors, data = data)

      validated_q <- reactive({
        validate_input(
          "letters-values-selected",
          condition = length(anl_inputs$letters()$values$selected) > 0,
          message = "Select at least one letter."
        )
        validate_input(
          c("letters-values-selected", "letters2"),
          condition = all(!anl_inputs$letters()$values$selected %in% input$letters2),
          message = "Letters in the first group should not be in the second group."
        )
        data()
      })

      output$plot <- renderPlot({
        letters <- validated_q()$letters$letter
        tab <- rbind(
          Group1 = as.integer(letters %in% anl_inputs$letters()$values$selected),
          Group2 = as.integer(letters %in% input$letters2)
        )
        colnames(tab) <- letters
        barplot(
          tab,
          beside = TRUE, legend.text = TRUE, main = "Selected letters per group",
          col = c("steelblue", "tomato")
        )
      })
    })
  }
)

init(data = data, modules = my_module) |> shiny::runApp()
```


averissimo:
Problem 1: (stale error message)

https://github.com/user-attachments/assets/bf6a7cb8-0a0f-4eda-84c3-873862aafe0d

Problem 2: (initialization with error on `teal.picks`)

image

averissimo:
@gogonzo ready for review!

Notable edge cases with serial validation.

```r
#... 
# Serial
validate(need_input(...))
validate(need_input(...))
validate(need_input(...))
#...
# Parallel
validate(
  need_input(...),
  need_input(...),
  need_input(...)
)
```

Stale messages
    1. change input so all are valid
    2. change input so last one is invalid
    3. change input so first one is invalid
    4. it should show messages below first and last (on output only first message is shown)
    5. change input on last one
    6. Only first message should be shown

`textInput` has many inputchanged events in shiny, 1) when value changes 2) when input loses focus

---

Serial validation seems like the biggest bottle neck as the R code is not being run.

So the strategy is to identify the affected via a fingerprint and apply changes only to errors that came from the same `need_input()`. The drawback here is that there cannot be duplicate messages, which I think is acceptable.

Fingerprint is calculated from the code in the arguments, not the arguments' value themselves.

Issue #1728, opened 2026-08-14
Add `CLAUDE.md`


donyunardi:
## Description

Add a `CLAUDE.md` file to this repository to provide repository-specific context and guidance when working with Claude Code.

The structure and expected content for `CLAUDE.md` are being defined in insightsengineering/nestdevs-tasks#121. The implementation should follow the agreed structure from that issue and adapt the content to this repository.

The `CLAUDE.md` should capture the relevant repository-specific information while following the common conventions agreed for the teal repositories.

## Acceptance criteria

* `CLAUDE.md` is added to the repository.
* The file follows the structure agreed in insightsengineering/nestdevs-tasks#121.
* Repository-specific guidance is included where relevant.

PR #1730, opened 2026-08-19
Update `AGENT.md` to understand teal.picks

averissimo:
# Pull Request

- Part of #1728

This PR follows best practices that we found when converting module packages to `teal.picks` to create a simple example module that uses this functionality.

### Changes description

- Add `teal.picks` to packages
- Update example module to a working module that:
    - Uses encoding selectors
        - `teal.picks` to select variables
    - Generates outputs inside qenv
    - Uses both `ggplot2` and `gtsummary`


AGENTS.md
averissimo, line 38: @osenan suggested while discussing that there is no teal-specific structure here.

While true, we need to at least say that package structure should include tests, `.lintr` and `_pkgdown.yml`

AGENTS.md
```diff
@@ -129,64 +130,120 @@ Teal modules follow a specific pattern with UI and server components:
...
+    ui_args = args[names(args) %in% names(formals(ui_example_module))],
+    server_args = args[names(args) %in% names(formals(srv_example_module))]
+  )
+}
 ```
 
 ### Code Style for Modules
 
-- **Use tidyverse style**: Write clear, readable code using dplyr, ggplot2 patterns
+- **Use tidyverse style**: Write clear, readable code using dplyr, ggplot2 patterns and maggritr pipes
```
osenan: I would not suggest to use maggritr pipes, while we keep the magrittr pipes in packages for compatibility,  for new apps I think it is fine to use native R pipes

AGENTS.md
```diff
@@ -129,64 +130,120 @@ Teal modules follow a specific pattern with UI and server components:
...
 - **Prefer ggplot2**: For all visualizations over base R plotting
-- **Use tern/rtables**: For statistical tables and summaries
-- **Error handling**: Implement proper validation using `checkmate` and `teal.widgets::validate_inputs()`
+- **Use gt and gtsummary**: For statistical tables and summaries
+- **Error handling**: Implement proper validation using `checkmate` and `shiny::validate(teal::need_input(...))`
```
osenan, line 283: why not suggesting `teal::validate_input`?

AGENTS.md
osenan: Thinking again, it can be useful that we inform the agent about necessary files in a teal package. Let's keep it.

AGENTS.md
```diff
@@ -129,64 +130,120 @@ Teal modules follow a specific pattern with UI and server components:
...
+          message = "Y-Axis Variable must be selected"
+        )
       )
+      shiny::validate(
+        teal::need_input(
+          inputId = c("var_x-variables-selected", "var_y-variables-selected"),
+          condition = !any(selectors$var_x()$variables$selected %in% selectors$var_y()$variables$selected),
+          message = "X-axis variable and Y-axis variable must be different"
+        )
+      )
```
osenan, line 213: This could be also calls to teal::validate_input
I do not see a big issue using teal::need_input tough

osenan (approved):
Very good improvement.
I have minor comments on the usage of teal::need_input or validate_input, but it is not much different.
The new example is much better

AGENTS.md
averissimo: With `need_input` you can pack multiple validation inside a single validate input, which is useful to give out independent errors in inputs.

Let's say, there are 3 inputs that need values being selected. With `validate_input()` we get the error messages one at a time.

Do you think it should be extended?

AGENTS.md
averissimo: This is an example where the first block actually makes sense, as we don't want to wait for xaxis to be selected to get the error in yaxis.

The second block can be replaced with `teal::validate_input()`, however I would argue that this is preferable as it uses `shiny::validate()`, which is a better known function and the developer immediately understands what this blocks does to the reactive.

AGENTS.md
```diff
@@ -129,64 +130,120 @@ Teal modules follow a specific pattern with UI and server components:
...
+    qenv <- reactive({
+      within(validated_q(), {
+        plot <- ggplot2::ggplot(anl) +
+          ggplot2::geom_point(ggplot2::aes(x = env_var_x, y = env_var_y))
+        plot
+
+        table <- gtsummary::tbl_summary(anl, by = env_var_x, missing = "no")
+        table
+      }, env_var_x = as.name(merged$variables()$var_x), env_var_y = as.name(merged$variables()$var_y))
     })
```
llrs-roche: Shouldn't this be in two different qenvs? so that we can decorate them separately?

AGENTS.md
```diff
@@ -283,13 +340,13 @@ testthat::test_that("my_module UI renders correctly", {
...
-  driver$navigate_to("My Module")
-  
+  driver$navigate_teal_tab("My Module")
+
   # Test UI elements are present
-  testthat::expect_true(driver$is_visible("#plot"))
-  
+  driver$expect_visible("#plot")
+
   driver$stop()
```
llrs-roche: Shouldn't it be stopped earlier and with `withr::defer` ?

AGENTS.md
```diff
@@ -313,7 +370,7 @@ testthat::test_that("my_module UI renders correctly", {
 
 ### Website Generation
 
-Use `_pkgdown.yml` for documentation websites:
+Use `_pkgdown.yml` for documentation websites (`nesttemplate` is optional):
```
llrs-roche: Is nesttemplate really optional? Or there is something that should change there?

AGENTS.md
```diff
@@ -7,13 +7,14 @@ The teal framework uses shiny to create reproducible environments for analysts.
 ### Core Packages
 
 - **teal** - The main framework package providing the application structure
-- **teal.data** - Data management and validation
-- **teal.slice** - Data filtering capabilities
-- **teal.code** - Code generation and evaluation
+- **teal.code** - Bare code generation and evaluation ensuring reproducibility
+- **teal.data** - Data management and relationships between datasets
```
llrs-roche, line 11: I would also mention it has the data for examples

AGENTS.md
```diff
@@ -7,13 +7,14 @@ The teal framework uses shiny to create reproducible environments for analysts.
...
+- **teal.reporter** - Report generation functionality
+- **teal.slice** - Data filtering capabilities for application
 - **teal.widgets** - Reusable UI components
 - **teal.logger** - Standardized logging across the framework
-- **teal.reporter** - Report generation functionality
-- **teal.transform** - Data transformation utilities
+- **teal.picks** - Data selection and merging utilities using `teal.data` objects
+- **teal.transform** - Data transformation utilities (deprecated)
```
llrs-roche: Might be worth mentioning that teal.picks should replace teal.transform:

```suggestion
- **teal.transform** - Data transformation utilities (deprecated in favor of teal.picks)
```

AGENTS.md
llrs-roche: The problem with using the native R pipeline is that code generated with `with(qenv, {a |> b() })` might end up unreadable as this is directly read as `b(a)`. For this simple example it doesn't matter but for long dplyr pipelines it affects a lot the readability. 

Maybe the comment should be more specific and mention the maggritr pipe for code meant to be used inside modules and reproducible.

llrs-roche (approved):
This is my first time reading the document. I left some comments but most of them are not focused on teal.picks, where it does a great job, but rather on how the file will steer the AI. 

General comments:
 - there is no mention of decorators and transformers on the examples provided
 - There is no mention on `shinytests2` if `testServer()` can't cover up to 80%

I am not sure all these are specific to the repository or they are more general for the whole teal framework. However, in general I agree with your answers to the previous comments and I think it can be merged.

AGENTS.md
averissimo: ?

```suggestion
- **teal.data** - Data management and relationships between datasets
(contains sample data for ADaM datasets and default keys to merge ADaM datasets)
```

AGENTS.md
averissimo: Reverting

```suggestion
Use `_pkgdown.yml` for documentation websites:
```

AGENTS.md
averissimo: Done

AGENTS.md
averissimo: Exacly @llrs-roche , I though I replied to this comment, but you make the point!

Maybe this should be its own bullet point, one of below? :thinking: 

```suggestion
- **Use tidyverse style**: Write clear, readable code using dplyr, ggplot2 patterns
- **Use maggritr pipes in reproducible execution**: For code executed for `teal_data`/`qenv` data objects with `eval_code()` and `within()`
```

```suggestion
- **Use tidyverse style**: Write clear, readable code using dplyr, ggplot2 patterns
- **Use maggritr pipes in reproducible code**: For code executed with `eval_code()` and `within()`
```

AGENTS.md
averissimo: Not necessarily, as in some situiations we just execute them in series.

In other we use shared `qenv` and then merge with `c()`

In practice it's the same thing, but that may be a good practice, I'm updating it

averissimo:
@llrs-roche added `transformators` and `decorators`. The example didn't grow too much and it becomes much more complete.

AGENTS.md
llrs-roche: I like the suggested change. Go for it @averissimo

PR #1735, opened 2026-09-08
Removes import of `%||%` in older versions of R

averissimo:
# Pull Request

### Changes description

- Remove `%||%` import as function is not being used


llrs-roche (commented):
While I like it (less imports, les trouble) I am not sure what bug is actually fixing.
We still support R > 4.1 where is not present, I didn't find any usage of it at R/*.R files but I do see it on the tests.

averissimo:
You're right! My find was limiting to `/R`. I'll revert it to the new solution

llrs-roche (approved):
Better solution than just in code imho.

PR #1737, opened 2026-09-15
Fixes reporter bug where buttons would appear in other tabs when reporter is `NULL` (alternative)

### LINKED ISSUE #1727: [Bug]: `init(reporter = NULL)` only removed the Add to Report button in the first module
### What happened?

When setting `init(reporter = NULL)`, the expectation is for Add to Report button to be removed from all modules in teal app. 

Current behavior only removes the button on the first module. 

Image

Code to reproduce:
```r
library(teal)
app <- init(
  data = within(teal_data(), iris <- iris),
  modules = modules(
    example_module(label = "example teal module"),
    example_module(label = "example teal module")
  ),
  reporter = NULL
)
if (interactive()) {
  shinyApp(app$ui, app$server)
}
```

### sessionInfo()

```R

```

### Relevant log output

```R

```

### Code of Conduct

### Contribution Guidelines

### Security Policy

averissimo:
# Pull Request

Fixes #1727

Alternative approach: https://github.com/insightsengineering/teal/pull/1737

### Changes description

- Add `reporter` argument to `ui_teal*` functions and propagate it down
- Menu item and "Add to Report" button is not rendered instead of being removed by JS

### Alternative

Part of the reason why we didn't do this before was to allow `srv_teal` and `ui_teal` to work with reporter by only supplying the server function and allow it to manage the UI

The drawback is that there is a visual glitch where the buttons are showed momentarily, before `removeUi` is executed (however that was something already established)


R/module_teal.R
```diff
@@ -67,32 +68,34 @@ ui_teal <- function(id, modules) {
...
-          },
-          if ("load" %in% getOption("teal.reporter.nav_buttons")) {
-            teal.reporter::report_load_ui(ns("load_report"), label = "Load Report")
-          },
-          tags$hr(style = "margin: 0.5rem;"),
-          if ("reset" %in% getOption("teal.reporter.nav_buttons")) {
-            teal.reporter::reset_report_button_ui(ns("reset_reports"), label = "Reset Report")
-          }
+    if (!is.null(reporter)) {
+      withr::with_options(reporter_opts, { # for backwards compatibility of the report_previewer_module$server_args
```
osenan, line 74: I am surprised `withr` package is in suggest, it shuold be on imports as it is necessary for the reporter funcionality.

R/module_teal.R
osenan, line 1: Maybe we can make small update in vignette `teal-as-shiny-module` to show the usage of reporter argument for function `ui_teal`

osenan (commented):
This PR is also simple, not as minimal as its alternative https://github.com/insightsengineering/teal/pull/1738 but stays consistent. I think we need to update the status of `withr` in the DESCRIPTION and slightly modify the vignette of the teal as a shiny module.
Corresponds more to the product decision to see which solution is accepted. The complexity of this PR is small, IMO opinion the improvement is not very different, but we can accept both solutions.

R/module_teal.R
averissimo: Good point, we can use `rlang::with_options` which should result in the same outcome

